PR #36, opened 2018-06-28
Add test-sanity script

davidz627:


davidz627:
ping @msau42 this is ready

msau42:
Why are there two pull-gcp-compute-persistent-disk-csi-driver-sanity jobs?

davidz627:
@msau42 I only see 1

msau42:
Bah I can't read
/lgtm

Issue #43, opened 2018-07-03
ControllerPublishVolume should Verify Disk Attached


davidz627:
After the attach call, and waiting for the attach op to complete. Sometimes the attach will not error but the disk will not be attached.

We should add an additional verification to make sure that the disk has been attached. We can see this on the disk object:
```
// Users: [Output Only] Links to the users of the disk (attached
// instances) in form: project/zones/zone/instances/instance
Users []string `json:"users,omitempty"`
```
OR on the instance there is AttachedDisks.

PR #44, opened 2018-07-03
Added a waitforattach verification at the end of ControllerPublishVolume

davidz627:
Fixes: #43 

There is an issue where Attach and the Attach Op can return success/DONE but the volume may not actually be attached to the instance. This PR adds an extra check where we poll the disk object for the instances it is attached to before returning success.

/assign @msau42


pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -255,6 +255,7 @@ func TestCreateVolumeArguments(t *testing.T) {
 }
 
 // Test volume already exists
+
```
msau42: hmmmmm

PR #50, opened 2018-07-03
Added unit tests for identity server, fixed small spec inconsistency

davidz627:
Added unit tests. In the process found out the GetPluginInfo was missing vendor version so it was fixed to return the vendor version now.

vendor version is: The version of the driver as declared by the vendor (driver author)
so it makes sense to be a const and not a flag

/assign @msau42 

Fixes: #47

### DIFF
--- a/pkg/gce-pd-csi-driver/gce-pd-driver.go
+++ b/pkg/gce-pd-csi-driver/gce-pd-driver.go
@@ -39,21 +39,11 @@ type GCEDriver struct {
 	nscap []*csi.NodeServiceCapability
 }
 
-const (
-	vendorVersion = "v0.3.0-alpha"
-)
-
 func GetGCEDriver() *GCEDriver {
-	return &GCEDriver{
-		vendorVersion: vendorVersion,
-	}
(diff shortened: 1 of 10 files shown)


pkg/gce-pd-csi-driver/gce-pd-driver.go
```diff
@@ -52,6 +57,7 @@ func (gceDriver *GCEDriver) SetupGCEDriver(cloudProvider gce.GCECompute, mounter
 
 	gceDriver.name = name
 	gceDriver.nodeID = nodeID
+	gceDriver.vendorVersion = vendorVersion
```
msau42: There's a place down in the cloud provider file where you're putting the driver name/version in the string.  You may want to plumb this down to there too

davidz627:
/hold

Going to implement better compile time versioning injection

davidz627:
/hold cancel

Added version compile time injection

davidz627:
@msau42 ready for review

davidz627:
@msau42 for re-lgtm, sorry

Issue #49, opened 2018-07-03
Run unit tests in CI


davidz627:
Depends on #46 #47 #48

davidz627:
CI add:
https://github.com/kubernetes/test-infra/pull/8641

PR #51, opened 2018-07-03
Add unit test runner script. Fixed some errors caught by unit tests

davidz627:
Part 1 of: #49 
Part 2 will be adding run-unit.sh to the test-infra config

/assign @msau42

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -52,23 +52,38 @@ const (
 	attachableDiskTypePersistent = "PERSISTENT"
 )
 
-func getRequestCapacity(capRange *csi.CapacityRange) (capBytes int64) {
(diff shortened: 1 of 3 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -52,23 +52,38 @@ const (
 	attachableDiskTypePersistent = "PERSISTENT"
 )
 
-func getRequestCapacity(capRange *csi.CapacityRange) (capBytes int64) {
+func getRequestCapacity(capRange *csi.CapacityRange) (int64, error) {
```
msau42: You may want to check out what I did for this (in fact, it should probably be a utility function that can be shared by multiple drivers), and my test cases.

davidz627: I wrote some tests for this function in a followup PR. Lets work on a canonical `getRequestCapacity`.

Issue #42, opened 2018-07-03
Flakey pull-gcp-compute-persistent-disk-csi-driver-e2e


davidz627:
The E2E test on CI is really flaky. Needs investigation.

PR #56, opened 2018-07-10
Lessen severity of waiting loop errors to warnings

davidz627:
This might help #42 

It seems like the e2e test is throwing too many errors when trying to bring up the instance.


davidz627:
/retest pull-gcp-compute-persistent-disk-csi-driver-e2e

to make sure we've fixed this?

PR #57, opened 2018-07-10
Fix for E2E test Flake

davidz627:
Fixes: #42 

Trying to fix the flake. Got a couple theories, gonna try them out


davidz627:
/assign @msau42 
This one actually fixes the flake I am 100% sure. I have determined this through "proprietary science methods"

PR #62, opened 2018-07-11
Fix deployment readme

saad-ali:
Follow-up PR will add example YAML

### DIFF
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -1,10 +1,17 @@
-1. One-time per project: Create GCP service account for CSI driver
-    1. Export environment variables for location for service account private key file and name of the service account
+1. [One-time per project] Create GCP service account credential for CSI Driver and download JSON:
+    1. Go to https://pantheon.corp.google.com/apis/credentials
+    2. Select `Create credentials` -> ~Service account key`
+    3. For `Service Account` select `Compute Engine Default service account`
+    4. Select `JSON`
+    5. Click `Create` to download JSON for the private key to your local machine.
+    6. Rename the downloaded JSON file to `cloud-sa.json`
+    6. Export environment variables for 1) location of service account private key JSON file, 2) name you want of service account to be created by `setup_project.sh`, and 3) GCP project:
     ```
-    $ export SA_FILE=~/.../cloud-sa.json
-    $ export GCEPD_SA_NAME=sample-service-account
+    $ export SA_FILE=~/Downloads/cloud-sa.json
+    $ export GCEPD_SA_NAME=gcepd-csi-service-account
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/README.md
```diff
@@ -1,18 +1,37 @@
-1. One-time per project: Create GCP service account for CSI driver
-    1. Export environment variables for location for service account private key file and name of the service account
+1. [One-time per project] Create GCP service account credential for CSI Driver and download JSON:
+    1. Go to https://pantheon.corp.google.com/apis/credentials
+    2. Select `Create credentials` -> ~Service account key`
+    3. For `Service Account` select `Compute Engine Default service account`
+    4. Select `JSON`
+    5. Click `Create` to download JSON for the private key to your local machine.
```
msau42: You shouldn't have to download keys from pantheon. The script should already create the service account and key.

deploy/kubernetes/README.md
```diff
@@ -1,18 +1,37 @@
...
+    6. Rename the downloaded JSON file to `cloud-sa.json`
+    6. Export environment variables for 1) location of service account private key JSON file, 2) name you want of service account to be created by `setup_project.sh`, and 3) GCP project:
     ```
-    $ export SA_FILE=~/.../cloud-sa.json
-    $ export GCEPD_SA_NAME=sample-service-account
+    $ export SA_FILE=~/Downloads/cloud-sa.json
```
davidz627: Could be any directory. Need to emphasise that it MUST be called "cloud-sa.json" though

deploy/kubernetes/README.md
```diff
@@ -1,18 +1,37 @@
...
 ```
-3. Create example PVC and Pod
-```
-$ kubectl apply -f ./examples/demo-pod.yaml
-```
\ No newline at end of file
+3. Create example workload to verify
+    1. Deploy yaml for PVC and Pod
+    ```
+    $ kubectl apply -f ./examples/demo.yaml
```
davidz627: `./examples/kubernetes/demo.yaml` ?

Also we can take the demopod from `./deploy/kubernetes/demo-pod.yaml`

deploy/kubernetes/README.md
saad-ali: @davidz627 could you clarify? Are these steps to create and download the key required? If not, is the target JSON path something that will be created?

davidz627: there is a script that does substeps 1-6, it is called `setup_project.sh`. As long as you set all the right env variables the script will handle creating service account, binding roles, and downloading the json key to the provided directory (`SA_FILE`)

davidz627:
Incorporated into #86 as part of a larger documentation fix

PR #63, opened 2018-07-11
Add simple example for verification of driver

saad-ali:
Follow up to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/62

### DIFF
--- a/examples/demo.yaml
+++ b/examples/demo.yaml
@@ -1,0 +1,41 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mypvc
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: csi-gce-pd
+
+---
+
+apiVersion: v1
+kind: ReplicationController
+metadata:
+  name: sleepypod
+spec:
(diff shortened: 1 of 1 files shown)


examples/demo.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: csi-gce-pd
+
+---
+
+apiVersion: v1
+kind: ReplicationController
```
msau42: We recommend deployments or Statefulsets now?

examples/demo.yaml
```diff
@@ -0,0 +1,41 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: mypvc
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: csi-gce-pd
```
msau42: Should the storageclass also be an example?

davidz627:
There's already an example pod/pvc in deployment/kubernetes. Maybe it would make more sense to move it into an examples/ folder somewhere.

davidz627:
Closing due to existing example

PR #73, opened 2018-07-18
Remove old policy bindings so that setup-project is reentrant

davidz627:
Have the `setup-project.sh` script remove old policy bindings so that they can be recreated successfully.

Part of: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/73

/assign @msau42

### DIFF
--- a/deploy/setup-project.sh
+++ b/deploy/setup-project.sh
@@ -3,17 +3,40 @@
 set -o nounset
 set -o errexit
 
+readonly PKGDIR="${GOPATH}/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver"
+readonly KUBEDEPLOY="${PKGDIR}/deploy/kubernetes"
+
+BIND_ROLES="roles/compute.storageAdmin roles/iam.serviceAccountUser projects/${PROJECT}/roles/gcp_compute_persistent_disk_csi_driver_custom_role"
 IAM_NAME="${GCEPD_SA_NAME}@${PROJECT}.iam.gserviceaccount.com"
 
-# Cleanup old Service Account and Key
+# Create or Update Custom Role
+if gcloud iam roles describe gcp_compute_persistent_disk_csi_driver_custom_role --project "${PROJECT}";
+then
+  yes | gcloud iam roles update gcp_compute_persistent_disk_csi_driver_custom_role \
+          --project "${PROJECT}"                                                   \
+          --file "${PKGDIR}/deploy/gcp-compute-persistent-disk-csi-driver-custom-role.yaml"
+else
+  gcloud iam roles create gcp_compute_persistent_disk_csi_driver_custom_role \
+    --project "${PROJECT}"                                                   \
+    --file "${PKGDIR}/deploy/gcp-compute-persistent-disk-csi-driver-custom-role.yaml"
+fi
+
+# Delete Service Account Key
 if [ -f $SA_FILE ]; then
   rm "$SA_FILE"
 fi
+# Delete Bindings
+for role in ${BIND_ROLES}
+do
+  gcloud projects remove-iam-policy-binding "${PROJECT}" --member serviceAccount:"${IAM_NAME}" --role $role --quiet || true
+done
+# Delete Service Account
 gcloud iam service-accounts delete "$IAM_NAME" --quiet || true
-# TODO: Delete ALL policy bindings
 
 # Create new Service Account and Keys
 gcloud iam service-accounts create "${GCEPD_SA_NAME}"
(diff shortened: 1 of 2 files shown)


davidz627:
@msau42 updated so that roles are fairly minimal

deploy/setup-project.sh
```diff
@@ -3,17 +3,40 @@
...
+readonly KUBEDEPLOY="${PKGDIR}/deploy/kubernetes"
+
+BIND_ROLES="roles/compute.storageAdmin roles/iam.serviceAccountUser projects/${PROJECT}/roles/gcp_compute_persistent_disk_csi_driver_custom_role"
 IAM_NAME="${GCEPD_SA_NAME}@${PROJECT}.iam.gserviceaccount.com"
 
-# Cleanup old Service Account and Key
+# Create or Update Custom Role
+if gcloud iam roles describe gcp_compute_persistent_disk_csi_driver_custom_role --project "${PROJECT}";
+then
+  yes | gcloud iam roles update gcp_compute_persistent_disk_csi_driver_custom_role \
```
msau42: does the -q option not work for suppressing interactive mode?

deploy/setup-project.sh
```diff
@@ -3,17 +3,40 @@
...
+  gcloud projects remove-iam-policy-binding "${PROJECT}" --member serviceAccount:"${IAM_NAME}" --role $role --quiet || true
+done
+# Delete Service Account
 gcloud iam service-accounts delete "$IAM_NAME" --quiet || true
-# TODO: Delete ALL policy bindings
 
 # Create new Service Account and Keys
 gcloud iam service-accounts create "${GCEPD_SA_NAME}"
-gcloud iam service-accounts keys create "${SA_FILE}" --iam-account "${IAM_NAME}"
-gcloud projects add-iam-policy-binding "${PROJECT}" --member serviceAccount:"${IAM_NAME}" --role roles/compute.admin
```
msau42: your role-binding teardown logic will not be backwards compatible anymore. maybe it needs to somehow iterate through all the rolebindings the SA has, and delete them

deploy/setup-project.sh
davidz627: brainfarted

deploy/setup-project.sh
davidz627: resolved

deploy/setup-project.sh
```diff
@@ -12,24 +12,24 @@ IAM_NAME="${GCEPD_SA_NAME}@${PROJECT}.iam.gserviceaccount.com"
...
   rm "$SA_FILE"
 fi
-# Delete Bindings
-for role in ${BIND_ROLES}
-do
-  gcloud projects remove-iam-policy-binding "${PROJECT}" --member serviceAccount:"${IAM_NAME}" --role $role --quiet || true
-done
+# Delete ALL EXISTING Bindings
+gcloud projects get-iam-policy "${PROJECT}" --format json > "${PKGDIR}/deploy/iam.json"
+sed -i "/${IAM_NAME}/d" "${PKGDIR}/deploy/iam.json"
```
msau42: Could this accidentally match a substring of the IAM_NAME?

Say you had something called "foobar", and your IAM_NAME was "bar"

davidz627: good catch, amended with prefer `serviceAccount:`

deploy/setup-project.sh
```diff
@@ -12,24 +12,24 @@ IAM_NAME="${GCEPD_SA_NAME}@${PROJECT}.iam.gserviceaccount.com"
...
 fi
-# Delete Bindings
-for role in ${BIND_ROLES}
-do
-  gcloud projects remove-iam-policy-binding "${PROJECT}" --member serviceAccount:"${IAM_NAME}" --role $role --quiet || true
-done
+# Delete ALL EXISTING Bindings
+gcloud projects get-iam-policy "${PROJECT}" --format json > "${PKGDIR}/deploy/iam.json"
+sed -i "/serviceAccount:${IAM_NAME}/d" "${PKGDIR}/deploy/iam.json"
+gcloud projects set-iam-policy "${PROJECT}" "${PKGDIR}/deploy/iam.json"
```
msau42: hmm although in general I find this kind of scary because it sets all the rolebindings in your entire project to the .json, not just this particular service account.  If someone is also modifying the iam policies at the same time, this could override their changes.

davidz627: theres an etag that is used for consistency. `gcloud projects set-iam-policy` will fail if the etag is out of date. I haven't written in any failure handling logic. It shouldn't be too difficult to just run the script again.

PR #74, opened 2018-07-20
Add detach disk permissions to custom role

davidz627:
Lol I forgot this one. Its sort of important

/assign @msau42

### DIFF
--- a/deploy/gcp-compute-persistent-disk-csi-driver-custom-role.yaml
+++ b/deploy/gcp-compute-persistent-disk-csi-driver-custom-role.yaml
@@ -4,3 +4,4 @@ stage: ALPHA
 includedPermissions:
 - compute.instances.get
 - compute.instances.attachDisk
+- compute.instances.detachDisk


deploy/gcp-compute-persistent-disk-csi-driver-custom-role.yaml
```diff
@@ -4,3 +4,4 @@ stage: ALPHA
 includedPermissions:
 - compute.instances.get
 - compute.instances.attachDisk
+- compute.instances.detachDisk
```
msau42: does the e2e not test this?

davidz627: e2e currently doesn't test service account  because thats kind of "out-of-band." I guess I could have it invoke the setup scripts and pull the environment variables for Service Account just like everywhere else.... Now that I write that it seems like a good idea.

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/75

PR #77, opened 2018-07-23
Topology support

davidz627:
The driver will return the node toplogy on `GetNodeInfo` and supports simple `Requisite` topology specification on the create volume call.

Also the testing infrastructure was changed to support topology.

Fixes: #13

/assign @msau42 
/assign @verult

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -1,0 +1,26 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+const (
+	// Keys for Storage Class Parameters
+	ParameterKeyZone = "zone"
+	ParameterKeyType = "type"
+
+	// Keys for Topology. This key will be shared amonst drivers from GCP
+	TopologyKeyZone = "com.google.topology/zone"
+)
--- a/pkg/gce-cloud-provider/metadata/metadata.go
+++ b/pkg/gce-cloud-provider/metadata/metadata.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	"fmt"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+// MetadataService is a fakeable interface exposing necessary data
+// from the GCE Metadata service
+type MetadataService interface {
+	GetZone() string
+	GetProject() string
+}
+
+type metadataServiceManager struct {
+	// Current zone the driver is running in
+	zone    string
+	project string
+}
+
+var _ MetadataService = &metadataServiceManager{}
+
+func NewMetadataService() (MetadataService, error) {
+	zone, err := metadata.Zone()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get current zone: %v", err)
+	}
+	projectID, err := metadata.ProjectID()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get project: %v", err)
+	}
+
+	return &metadataServiceManager{
+		project: projectID,
+		zone:    zone,
+	}, nil
+}
+
+func (manager *metadataServiceManager) GetZone() string {
+	return manager.zone
+}
+
+func (manager *metadataServiceManager) GetProject() string {
+	return manager.project
+}
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -16,6 +16,7 @@ package gceGCEDriver
 
 import (
 	"fmt"
+	"math/rand"
 	"strings"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi/v0"
@@ -24,8 +25,8 @@ import (
 	compute "google.golang.org/api/compute/v1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider"
-	utils "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/utils"
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
+	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
 )
 
 // TODO: Add noisy glog.V(5).Infof() EVERYWHERE
@@ -53,7 +54,6 @@ const (
 )
 
 func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
-	// TODO: Check create zone against Driver zone. They must MATCH
 	glog.Infof("CreateVolume called with request %v", *req)
 
 	// Validate arguments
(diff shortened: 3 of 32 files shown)


pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -136,6 +136,65 @@ func TestCreateVolumeArguments(t *testing.T) {
...
+				Parameters:         map[string]string{"type": "test-type"},
+				AccessibilityRequirements: &csi.TopologyRequirement{
+					Requisite: []*csi.Topology{
+						{
+							Segments: map[string]string{"zone": "topology-zone"},
+						},
+					},
+				},
+			},
+			expVol: &csi.Volume{
```
msau42: The returned volume should also have topology set right?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -136,6 +136,65 @@ func TestCreateVolumeArguments(t *testing.T) {
...
+				CapacityBytes: utils.GbToBytes(20),
+				Id:            "topology-zone/test-vol",
+				Attributes:    nil,
+			},
+		},
+		{
+			name: "success with topology overwrite zone in params",
```
msau42: I think we actually wanted to error in this case because we don't just want to drop/ignore user config

davidz627:
@msau42 comments addressed

davidz627:
@msau42 sorry this PR really balooned but everything was kind of intertwined :(

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -341,12 +368,23 @@ func (gceCS *GCEControllerServer) ValidateVolumeCapabilities(ctx context.Context
 	// TODO: Factor out the volume capability functionality and use as validation in all other functions as well
 	glog.V(5).Infof("Using default ValidateVolumeCapabilities")
 	// Validate Arguments
+	if req.GetVolumeCapabilities() == nil || len(req.GetVolumeCapabilities()) == 0 {
```
verult: What about `AccessibleTopology`?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -458,3 +496,52 @@ func diskIsAttachedAndCompatible(volume *compute.Disk, instance *compute.Instanc
...
 }
+
+func pickTopology(top *csi.TopologyRequirement) (string, error) {
+	reqTop := top.GetRequisite()
+	prefTop := top.GetPreferred()
+
+	// Pick the preferred toplogy in order
+	if len(prefTop) != 0 {
+		if prefTop[0].GetSegments() == nil {
+			return "", fmt.Errorf("preferred toplogies specified specified but no segments")
```
verult: "specified" duplicated, and a couple of "toplogy" and "toplogies" in this file

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -458,3 +496,52 @@ func diskIsAttachedAndCompatible(volume *compute.Disk, instance *compute.Instanc
...
+		}
+
+		// GCE PD cloud provider Create has no restrictions so just create in top preferred zone
+		zone, err := getZoneFromSegment(prefTop[0].GetSegments())
+		if err != nil {
+			return "", fmt.Errorf("could not get zone from preferred topology: %v", err)
+		}
+		return zone, nil
+	} else if len(reqTop) != 0 {
+		r := rand.Intn(len(reqTop))
```
verult: The in-tree plugin has a round-robin mechanism to choose zones ([here](https://github.com/kubernetes/kubernetes/blob/845a55dbbda9a35cfd3659f1fe880d35a03dc44b/pkg/volume/util/util.go#L439)). Could we do something similar?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -78,29 +79,52 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
+		}
+		configuredZone, err = pickTopology(req.GetAccessibilityRequirements())
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume failed to pick topology: %v", err))
+		}
+	}
+
+	if len(configuredZone) == 0 {
+		// Default to zone that the driver is in
+		configuredZone = gceCS.CloudProvider.GetZone()
```
verult: The zone-picking strategy should be consistent, regardless of whether zones are specified through parameters or AccessibilityRequirement

pkg/gce-cloud-provider/metadata/metadata.go
```diff
@@ -0,0 +1,60 @@
...
+
+package metadata
+
+import (
+	"fmt"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+type MetadataService interface {
```
verult: Comments, especially around the intended purpose of the interface in order to limit its unwanted expansion. I'm worried about future developers adding arbitrary methods here.

nit: "Service" gives me an impression of a long running process, and it's already an overloaded term in the container world

pkg/gce-pd-csi-driver/controller.go
davidz627: what do you mean by this? consistent with what?

pkg/gce-pd-csi-driver/controller.go
davidz627: we are not guaranteed to have a monotonically increasing id/name for the CSI Driver. Using a hash strategy could have unwanted behavior. We would have to actually keep track of which we've done or not to get round-robin going. I think a random pick is actually our best option her e

pkg/gce-pd-csi-driver/controller.go
davidz627: done

pkg/gce-pd-csi-driver/controller.go
davidz627: done

davidz627:
addressed comments.

pkg/common/constants.go
```diff
@@ -0,0 +1,26 @@
...
+
+package common
+
+const (
+	// Keys for Storage Class Parameters
+	ParameterKeyZone = "zone"
+	ParameterKeyType = "type"
+
+	// Keys for Topology
+	TopologyKeyZone = "com.google.gcp-compute-persistent-disk-csi-driver.topology/zone"
```
msau42: I think we should not put the driver name in the key.  Filestore driver should be able to share this key too.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -78,29 +79,52 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
 			diskType = v
-		case "zone":
+		case common.ParameterKeyZone:
 			configuredZone = v
 		default:
 			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("invalid option %q", k))
 		}
 	}
 
+	// Topology overwrites Storage Class Params if specified
```
msau42: I think we should not allow both zone/zones and AllowedTopologies to be specified at the same time.  But I'm not sure if we can actually detect that because csi provisioner will always fill in the requisite topologies, even when AllowedTopologies is not specified. cc @verult 

So I think the next best option would be to use the zone/zones parameters because that's what the user specified, so we shouldn't just ignore it.

pkg/gce-pd-csi-driver/controller.go
msau42: I think what @verult means is that once you enable the topology capability, then you should always get an accessibility requirement.  So if you run into a case where there is no zone specified, then you should error.

pkg/gce-pd-csi-driver/controller.go
msau42: From a backwards compatibility standpoint for in-tree migration, do we need to support this though?  cc @saad-ali

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -458,3 +519,52 @@ func diskIsAttachedAndCompatible(volume *compute.Disk, instance *compute.Instanc
...
 	return false, nil
 }
+
+func pickTopology(top *csi.TopologyRequirement) (string, error) {
+	reqTop := top.GetRequisite()
+	prefTop := top.GetPreferred()
+
+	// Pick the preferred topology in order
+	if len(prefTop) != 0 {
+		if prefTop[0].GetSegments() == nil {
```
msau42: You could have a common function to pick extract a zone from a topology, to simplify the preference vs required logic.

pkg/common/constants.go
davidz627: I thought we didn't want collisions between topology keys? I may be misunderstanding what this is used for exactly

msau42: We want to be able to share topology keys amongst drivers from the same vendor.
1. To avoid proliferation of labels on Node objects
2. To allow users to use a common set of topology labels in their Pods.

pkg/gce-pd-csi-driver/controller.go
davidz627: I believe accessibility requirements can be unset:
```
// If this field is not specified and the SP has the
// ACCESSIBILITY_CONSTRAINTS plugin capability, the SP MAY choose
// where the provisioned volume is accessible from.
```

The current logic is:
1) Only one of `Parameters` or `AccessibilityRequirements` may be set. Error if both are set.
2) If neither are set (valid according to CSI Spec) default to zone that driver is in

This logic makes sense to me according to the Spec. Please let me know if you think it should be changed and we can discuss

pkg/gce-pd-csi-driver/controller.go
davidz627: picking a zone from topology is different between preference vs required so it doesn't make sense to me to abstract this out, I may be misunderstanding your suggestion

pkg/gce-pd-csi-driver/controller.go
davidz627: This does not affect migration. There are two potential cases for the migration path:
1) The whole PD lifecycle is migrated to CSI, then CreateVolume is called and we pick a zone in whatever method we want (random here), all future references to this volume just continue to use the zone picked
2) PD is precreated in some zone, this logic is never used, we just use whatever zone was used in the creation (the round robin method)

I don't think theres any point after creation where we make assumptions on "how the zone was picked". so therefore it can't create any issues.

msau42: If a user created a statefulset:
1. Before the migration, the PDs would be evenly spread across zones
2. After the migration, the PDs would be randomly spread across zones

The question is, do we want to retain the old spreading behavior for backwards compatibility?  Or should we just declare multi-zone in-tree provisioning broken and require users to used delayed provisioning going forward?

davidz627:
@msau42 hoping to get this in by today if it looks good to you, I think all open questions have been resolved. Let me squash commits before your lgtm though, there are currently too many

davidz627:
@msau42 rebased, thanks!

Issue #6, opened 2018-06-13
Use well-tested k8s/k8s Mount utilities for Mount


davidz627:
k8s/k8s has some pretty well tested and thorough mount utilities. It would be nice to consume them directly instead of copy-pasting the mount utilities into this driver. There is an active issue tracking factoring these utilities out of k8s for easier consumption: https://github.com/kubernetes/kubernetes/issues/64953

Issue #48, opened 2018-07-03
Unit tests for Node


davidz627:

Mockery-Li:
Hi, I have added unit test for GetNodeInfo in issue[#19](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/19), and hope to complete other unit tests for Node.
Is there any sample for testing the mounter?

davidz627:
@Mockery-Li you may be able to find some examples in the existing k8s code: https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/gcepd/gce_pd_test.go#L129
^ I'm not really sure if this is what you're looking for but it might be a good starting point

msau42:
For mounter, we usually use the FakeMounter for testing.

Mockery-Li:
Thanks, that's what I need

davidz627:
Closed by #248

PR #78, opened 2018-07-26
Use well tested Kubernetes Mount package utilities instead of copy pasting

davidz627:
Fixes: #6 

Important precursor to #48 

Using well tested k8s/pkg/util/mount. Need to extend the interface a little bit to account for device name logic so the MountManager still needs to exist, but in a lesser capacity.

/assign @msau42


pkg/mount-manager/mount-manager.go
```diff
@@ -19,88 +17,106 @@ package mountmanager
...
-		// The device is formatted
-		return lines[0], nil
-	}
+}
 
-	if len(lines) == 1 {
-		// The device is unformatted and has no dependent devices
-		return "", nil
-	}
+var _ DeviceUtils = &deviceUtils{}
```
msau42: Can you put the DeviceUtils into a separate file?

davidz627: done

Issue #34, opened 2018-06-28
Add Flag to Driver for Controller or Node Mode


davidz627:
Add a flag `--node` to turn on the node server, `--controller` to turn on the controller server. They should both default to false.

Currently every driver runs all the node, controller, and identity servers by default.

msau42:
Could potentially also look into building them into separate binaries too

Jiawei0227:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/449 fixed it.
/close

mattcary:
/open

Shouldn't we actually use this flag as well to not run the controller on the node pods?

Jiawei0227:
Yep. I thought this issue is about adding the flag. We can use the issue to track changing the deployment for sure.

mattcary:
SG, I started #702.

PR #79, opened 2018-07-26
fix node deployment to include GOOGLE_APPLICATION_CREDENTIALS envvar

ashleyschuett:


msau42:
Are you seeing an issue without this fix?

The node controller only handles mount and unmount so should not need to make cloud provider calls

ashleyschuett:
When I tried to deploy the manifests without this fix it kept going into a crash loop back off. I'll test again today without this change and get the exact logs.

ashleyschuett:
Without the env var specified the `gce-pd-driver` container contains the following logs

```
# kubectl logs csi-gce-pd-node-5s5g5 gce-pd-driver
I0726 18:38:38.501711       1 main.go:49] Driver vendor version latest
W0726 18:38:38.512482       1 gce.go:97] GOOGLE_APPLICATION_CREDENTIALS env var not set
I0726 18:38:38.512520       1 gce.go:99] Using DefaultTokenSource &oauth2.reuseTokenSource{new:google.computeSource{account:""}, mu:sync.Mutex{state:0, sema:0x0}, t:(*oauth2.Token)(nil)}
E0726 18:38:38.513326       1 gce.go:106] error fetching initial token: metadata: GCE metadata "instance/service-accounts/default/token" not defined
```

msau42:
Ah ok yes I think this is needed for GetNodeInfo in order to get the zone 

/ok-to-test

davidz627:
@ashleyschuett thanks for the catch and the contribution!

/lgtm

davidz627:
Sorry I approved prematurely, I am unable to reproduce this error.

The Node should not need any GCP credentials to function. The topology support PR (which includes `NodeGetInfo` that Michelle mentioned) (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/77) introduces a dependency on the GCE Metadata server so I thought this might be related to that, but that hasn't made it into master yet.

Depending on whether we need the credential for actual node operations (metadata server calls) or not, I'm thinking the actual solution may be this: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/34 instead of giving credentials to the node container that it doesn't strictly need.

Some additional questions to try figure out why we're seeing this error:
Are you deploying using the provided scripts in an unmodified state? Which version of Kubernetes are you running and is it on GCE or GKE?

ashleyschuett:
I am using the script on a GCE cluster that I created.

msau42:
How are you deploying the GCE cluster?

I think at least in a GKE and GCE kube-up deployment, there is some extra logic done to give the nodes a default service account.

davidz627:
Hi @ashleyschuett wondering if you have been able to deploy the driver without this issue since then?

davidz627:
Closing due to inactivity. Feel free to reopen if you are still seeing this issue.

ridv:
Now that #77  change has landed in Stable, I was unable to run this on our 1.18 cluster without adding these changes to the driver. (I was hitting the same errors as the author of the PR). 

I see #34 still has not landed. 

Any guidance here? Running the scripts from this repo untouched.

Issue #28, opened 2018-06-25
Service account file name is  hard coded, needs to be configurable


krunaljain:
Currently while deploying the driver, the Service Account environment variable needs to be in the format */cloud-sa.json. Ideally should be configurable based on the variable set by the user.

PR #81, opened 2018-08-01
Improve deployment scripts, make service account JSON configurable by directory, not name

davidz627:
Fixes: #28

Added some comments to the deployment scripts. JSON is finally configurable by directory, not name.

/assign @msau42


deploy/setup-project.sh
```diff
@@ -1,13 +1,26 @@
 #!/bin/bash
 
+# This script will setup the given project with a Service Account that has the correct
+# restricted permissions to run the gcp_compute_persistent_disk_csi_driver and download
+# the keys to a specified directory
+
+# WARNING: This script will delete and recreate all service accounts, bindings, and keys.
```
msau42: ALL service accounts???

deploy/setup-project.sh
```diff
@@ -1,13 +1,26 @@
...
+# Great care must be taken to not run the script with a service account that is currently
+# in use.
+
+# Args:
+# PROJECT: GCP project
+# GCE_PD_SA_NAME: Name of the service account to create
+# GCE_PD_SA_DIR: Directory to save the service account key
```
msau42: Are there defaults set?

deploy/setup-project.sh
davidz627: ohoh thats alarming. No just the one set.

deploy/setup-project.sh
davidz627: Should there be defaults? I think these are all important enough that there shouldn't be any defaults. I have `set -o nounset` so it throws an error if the vars are unset when the script runs

deploy/kubernetes/deploy-driver.sh
```diff
@@ -1,5 +1,15 @@
 #!/bin/bash
 
+# This script will deploy the GCP Compute Persistent Disk CSI Driver to the
+# currently available Kubernetes cluster
+
+# Note: setup-cluster.yaml depends on the existence of cluster-roles
+# system:csi-external-attacher and system:csi-external-provisioner
+# which are in Kubernetes version 1.10.5+
```
msau42: I would put this note in the readme right up top

davidz627:
Comments addressed

Issue #53, opened 2018-07-09
WaitForOp should also error on Status Error


davidz627:
For some operations the operation poll doesn't actually return an error but there is an error on the field on the op itself.

"You can see that a 200 OK was returned and the operation has status=DONE, but there's an "error" field on the operation you should be looking at.

So your code should probably look at pollOp.error in addition to err.

There are additional fields "httpErrorStatusCode" and "httpErrorMessage". Perhaps it should be 403 instead of 400, but that's a separate issue."

Kubernetes recently solved by:
```
if op.Error != nil && len(op.Error.Errors) > 0 && op.Error.Errors[0] != nil {
		e := op.Error.Errors[0]
		o.err = &GCEOperationError{HTTPStatusCode: op.HTTPStatusCode, Code: e.Code, Message: e.Message}
}
```

Issue #7, opened 2018-06-13
Support RePD


davidz627:
Support RePD for GCE PD.

Will likely depend on topology work as well

PR #84, opened 2018-08-08
Regional PD Implementation [Removing parameter "zone"] 

davidz627:
Fixes: #7 

This is the full regional PD implementation. 

BREAKING CHANGES:
* It includes REMOVAL of the parameter "zone"
* It ADDS the parameter "replication-type"
* NodeID format has been changed to: `{zone}/{nodeNameFromGCE}`
* VolumeID format has been changed to fully qualified format: 
`projects/{project}/zones/{zone}/disks/{name}`
or
`projects/{project}/regions/{region}/disks/{name}`

Also snuck in,
Fixes: #53

/assign @msau42 
/assign @verult


pkg/common/utils.go
```diff
@@ -31,14 +37,40 @@ func GbToBytes(Gb int64) int64 {
...
+	} else if splitId[2] == "regions" {
+		return meta.RegionalKey(splitId[5], splitId[3]), nil
+	} else {
+		return nil, fmt.Errorf("could not get id components, expected either zones or regions, got: %v", splitId[2])
+	}
+}
+
+func NodeIDToZoneAndName(id string) (string, string, error) {
 	splitId := strings.Split(id, "/")
 	if len(splitId) != 2 {
```
msau42: consts

pkg/common/utils.go
```diff
@@ -31,14 +37,40 @@ func GbToBytes(Gb int64) int64 {
...
 func CombineVolumeId(zone, name string) string {
 	return fmt.Sprintf("%s/%s", zone, name)
 }
+
+func GetRegionFromZones(zones []string) (string, error) {
+	if len(zones) < 1 {
+		return "", fmt.Errorf("no zones specified")
+	}
+	zone := zones[0]
+	splitZone := strings.Split(zone, "-")
```
msau42: Add a comment with the expected syntax

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -517,37 +502,57 @@ func diskIsAttachedAndCompatible(volume *compute.Disk, instance *compute.Instanc
...
+		// Add all preferred zones into zones
+		zones.Insert(prefZones...)
+		remainingNumZones := numZones - len(prefZones)
+		// Take all of the remaining zones from requisite zones
+		reqSet := sets.NewString(reqZones...)
+		prefSet := sets.NewString(prefZones...)
+		remainingZones := reqSet.Difference(prefSet)
+
+		if remainingZones.Len() < remainingNumZones {
+			return nil, fmt.Errorf("need %v zones in preferred and requisite topologies to satisfy, only have %v unique zones", numZones, reqSet.Union(prefSet).Len())
```
msau42: message could be simplified a little bit to say "need %v zones from topology, only got %v unique zones"

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -565,3 +570,125 @@ func getZoneFromSegment(seg map[string]string) (string, error) {
...
+		fullyQualifiedReplicaZones = append(
+			fullyQualifiedReplicaZones, cloudProvider.GetReplicaZoneURI(replicaZone))
+	}
+
+	err = cloudProvider.InsertDisk(ctx, meta.RegionalKey(name, region), diskType, capBytes, capacityRange, fullyQualifiedReplicaZones)
+	if err != nil {
+		return nil, fmt.Errorf("failed to insert regional disk: %v", err)
+	}
+
+	glog.Infof("Completed creation of disk %v", name)
```
msau42: Add log levels to all the Info statements

msau42:
Just did a quick scan review. Will do more in-depth later.

Nit: we're not actually deprecating 'zone', we're outright removing it

davidz627:
@msau42 addressed comments :)

pkg/common/utils.go
```diff
@@ -31,14 +46,41 @@ func GbToBytes(Gb int64) int64 {
...
+	splitId := strings.Split(id, "/")
+	if len(splitId) != volIDTotalElements {
+		return nil, fmt.Errorf("failed to get id components. Expected projects/{project}/zones/{zone}/disks/{name}. Got: %s", id)
+	}
+	if splitId[volIDToplogyKey] == "zones" {
+		return meta.ZonalKey(splitId[volIDDiskNameValue], splitId[volIDToplogyValue]), nil
+	} else if splitId[volIDToplogyKey] == "regions" {
+		return meta.RegionalKey(splitId[volIDDiskNameValue], splitId[volIDToplogyValue]), nil
+	} else {
+		return nil, fmt.Errorf("could not get id components, expected either zones or regions, got: %v", splitId[2])
```
msau42: 2 => volIDTopologyKey

pkg/common/utils.go
```diff
@@ -19,6 +19,21 @@ package common
 import (
 	"fmt"
 	"strings"
+
+	"k8s.io/kubernetes/pkg/cloudprovider/providers/gce/cloud/meta"
+)
+
+const (
+	volIDFmt = "projects/%s/zones/%s/disks/{}"
```
msau42: is this used anywhere?

pkg/common/utils.go
```diff
@@ -19,6 +19,21 @@ package common
...
+
+	volIDToplogyKey    = 2
```
msau42: Add a comment above here with the syntax

pkg/common/utils.go
```diff
@@ -31,14 +46,41 @@ func GbToBytes(Gb int64) int64 {
 	return Gb * 1024 * 1024 * 1024
 }
 
-func SplitZoneNameId(id string) (string, string, error) {
+func VolumeIDToKey(id string) (*meta.Key, error) {
```
msau42: unit tests?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -517,37 +505,57 @@ func diskIsAttachedAndCompatible(volume *compute.Disk, instance *compute.Instanc
...
+
+		if remainingZones.Len() < remainingNumZones {
+			return nil, fmt.Errorf("need %v zones from topology, only got %v unique zones", numZones, reqSet.Union(prefSet).Len())
 		}
-
-		// GCE PD cloud provider Create has no restrictions so just create in top preferred zone
-		zone, err := getZoneFromSegment(prefTop[0].GetSegments())
+		// Add the remaining number of zones into the set
+		// TODO: Maybe we should pick a random set to get
+		nSlice, err := pickRandNFromSlice(remainingZones.List(), remainingNumZones)
```
msau42: I think we actually want the first zone to be random.  But the second zone should always be +1 from the first (sorted)

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -19,128 +19,365 @@ import (
...
+	}
+	return zones, nil
+
 }
 
-func (cloud *CloudProvider) GetZone() string {
-	return cloud.zone
+func (cloud *CloudProvider) GetDisk(ctx context.Context, key *meta.Key) (*CloudDisk, error) {
+	switch key.Type() {
+	case meta.Zonal:
```
msau42: Instead of having case statements for the two types of disks, I'm wondering if it would be simpler to define a generic Disk interface, and have a zonal implementation, and a regional implementation.

davidz627: I think there are merits to both ways, although the current method I still like better for 3 reasons:
1) This isn't just an interface for disks, there are some other "types" of calls as well that wouldn't cleanly fit in a disk interface and I would rather just maintain 1 interface for "cloud provider"
2) Some calls are not different between regional/zonal, I believe in go since there is no inheritance we would just have to copy/paste that code if there were two types of disks
3) It's already like this and would take some effort to change. Without a strong reason otherwise it would be better to spend that effort elsewhere.

davidz627:
/hold
Will squash commits before merging

davidz627:
@msau42 addressed comments

davidz627:
/hold cancel
@msau42 for lgtm

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -565,3 +574,127 @@ func getZoneFromSegment(seg map[string]string) (string, error) {
...
+		return nil, fmt.Errorf("got wrong number of zones for zonal create volume: %v", len(zones))
+	}
+	diskZone := zones[0]
+	err := cloudProvider.InsertDisk(ctx, meta.ZonalKey(name, diskZone), diskType, capBytes, capacityRange, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to insert zonal disk: %v", err)
+	}
+
+	glog.V(4).Infof("Completed creation of disk %v", name)
+	disk, err := cloudProvider.GetDisk(ctx, meta.ZonalKey(name, diskZone))
```
jingxu97: now for each creation, there is an extra cloud provider GET API calls. Why not just use insertDisk which can return the disk object directly? In previous version, there is no GET call after the CREATE call, right?

davidz627: insertDisk doesn't return a disk object AFAIK. It returns an operation which we can poll for completion of the disk create operation, but getting the disk itself will be a seperate operation still.

We required this change not strictly because of RePD but because we wanted to have `volume ID` be the disk `selfLink` of the form `projects/{project}/zones|regions/{zone|region}/disks/{diskname}` and to get the self-link we need to get the disk

Issue #30, opened 2018-06-27
Improve Documentation


davidz627:
* Add Known Issues
* Add Project Status
* Add Plugin Features (supported parameters)
* Add Future Features
* Add User Guide
  * Improve setup scripts to work every time, with reasonable error messages when variables not defined
* Add Kubernetes Development
* Add Dependency Management Information

msau42:
I think it would also be better to restructure the front page README to focus on the user guide, more than development (there should will be more users than devs visiting the page)

PR #86, opened 2018-08-15
Revised documentation. Made versioning more clear. Made distinction between installation and examples more clear.

davidz627:
See Title.

Fixes: #30 

/assign @msau42


deploy/kubernetes/v0.1.0/controller.yaml
```diff
@@ -0,0 +1,70 @@
+kind: StatefulSet
```
msau42: I think keeping deployment specs per version is going to get difficult to maintain.  The approach I take is to tag the repo with the version, and if you want an older version, you need to get the spec from that tag.

deploy/kubernetes/latest/controller.yaml
```diff
@@ -42,7 +42,7 @@ spec:
               mountPath: /csi
         - name: gce-pd-driver
           imagePullPolicy: Always
-          image: gcr.io/google-containers/volume-csi/gcp-compute-persistent-disk-csi-driver:v0.1.0.alpha
+          image: gcr.io/dyzz-csi-staging/csi/gce-pd-driver:latest
```
msau42: Consider using kustomization for this purpose

msau42: Instead of calling this folder "latest", I would maybe consider "dev"

davidz627: I think I'll make another branch called dev and leave the master branch with the v0.1.0.alpha version since I have a feeling a lot of people might pull master to get the driver up and running

msau42: What I was thinking was having maybe two directories: stable and dev.  Stable on master will contain whatever is the latest stable version.

davidz627: that makes sense to me. saves branch switching as well as doesn't have an explosion of version numbers

davidz627:
comments addressed, hope this makes using the driver much easier!

deploy/kubernetes/dev/controller.yaml
```diff
@@ -42,7 +42,7 @@ spec:
               mountPath: /csi
         - name: gce-pd-driver
           imagePullPolicy: Always
-          image: gcr.io/google-containers/volume-csi/gcp-compute-persistent-disk-csi-driver:v0.1.0.alpha
+          image: gcr.io/dyzz-csi-staging/csi/gce-pd-driver:latest
```
msau42: You should look into kustomize at a later time.  It lets you patch specs with only the difference.  So anytime you want to change your specs, you don't have to change it both places.

Issue #83, opened 2018-08-06
Support partition parameter


saad-ali:
GCE PD in-tree driver supports partitioning

```YAML
    gcePersistentDisk:
      pdName: partitioned-disk
      fsType: ext4
      partition: 2
```

ehmm:
Hi @saad-ali, @davidz627 I've created a quick patch that works that allows specifying the partition number as another parameters after the zone/volumeName. (https://github.com/ehmm/gcp-compute-persistent-disk-csi-driver/tree/feature/partitionInVolumeId)

As you can see I'm new here, so I don't know if you want to merge it, or maybe it's missing some testing (I see the unit-tests there) -- or just leave it for now, but leaving a link here in the mean time and if you want me to create a PR let me know.

msau42:
Can you uniquely identify a volume if the volume id doesn't contain partition number though?  ie multiple volumes on the same disk but on different partitions

davidz627:
Ah you're right, they will collide as-is. Looks like it will have to go in the Volume ID.

davidz627:
You're right, the volumes are uniquely identified via volume ID so we would have conflicts here. I will change the PR to include partitions in Volume ID.

davidz627:
Additionally just found this in the spec:
```
Attributes MAY NOT uniquely identify
a volume. A volume uniquely identified by `id` SHALL always report
the same attributes.
```
just as extra reason to put it in the volume id

msau42:
@ehmm For your usecase, do you need hard capacity isolation (ie partition boundaries) or could using subdirectories + Kubernetes subpath feature work for you?

I think representing each partition as a separate PV in Kubernetes has some challenges with Pod scheduling, attaching and detaching, that I pointed out in #91

ehmm:
@msau42 for my use case the subPath solution would be great, I also wouldn't mind just having the root with all the subdirectories exposed as I have no isolation requirements. I personally don't imagine a lot of scenarios where isolation requirements would be needed, and I think that auto-mounting all partitions (with any supported fsType #93) is a good idea for general use cases.

davidz627:
Currently partition cannot be specified in attributes, it runs into the same issue as described with the in tree GCE PD here: https://github.com/kubernetes/kubernetes/issues/57915.

[...]

PR #91, opened 2018-08-16
[WIP] Add support for Partitions

davidz627:
Fixes: #83 

/assign @msau42 

/cc @ehmm
PTAL to see if this fulfills your needs.

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -28,10 +28,17 @@ const (
 	// Volume ID Expected Format
 	// "projects/{projectName}/zones/{zoneName}/disks/{diskName}"
 	// "projects/{projectName}/regions/{regionName}/disks/{diskName}"
-	volIDToplogyKey    = 2
-	volIDToplogyValue  = 3
-	volIDDiskNameValue = 5
-	volIDTotalElements = 6
+	// OR
+	// "projects/{projectName}/zones/{zoneName}/disks/{diskName}/partitions/{partition}"
+	// "projects/{projectName}/regions/{regionName}/disks/{diskName}/partitions/{partition}"
+
+	volIDToplogyKey                 = 2
+	volIDToplogyValue               = 3
+	volIDDiskNameValue              = 5
+	volIDPartitionKey               = 6
+	volIDPartitionValue             = 7
+	volIDTotalElements              = 6
+	volIDTotalElementsWithPartition = 8
 
 	// Node ID Expected Format
 	// "{zoneName}/{instanceName}"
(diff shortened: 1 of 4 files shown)


pkg/common/utils.go
```diff
@@ -28,10 +28,17 @@ const (
...
+	// "projects/{projectName}/zones/{zoneName}/disks/{diskName}/partitions/{partition}"
+	// "projects/{projectName}/regions/{regionName}/disks/{diskName}/partitions/{partition}"
+
+	volIDToplogyKey                 = 2
+	volIDToplogyValue               = 3
+	volIDDiskNameValue              = 5
+	volIDPartitionKey               = 6
+	volIDPartitionValue             = 7
+	volIDTotalElements              = 6
+	volIDTotalElementsWithPartition = 8
```
msau42: Can we just require all the elements while we're still alpha?  I think this conditional number of elements is going to get messy if we ever have to add new elements

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -185,7 +185,7 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 		return nil, status.Error(codes.InvalidArgument, "DeleteVolume Volume ID must be provided")
 	}
 
-	volKey, err := common.VolumeIDToKey(volumeID)
+	volKey, _, err := common.VolumeIDToKey(volumeID)
```
msau42: I think because we are not supporting creating volumes with partitions, we should not support delete volume with partitions.  You don't want to delete a disk that has multiple partitions on it because DeleteVolume got called for one of the partitions

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -218,7 +218,7 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 		return nil, status.Error(codes.InvalidArgument, "ControllerPublishVolume Volume capability must be provided")
 	}
 
-	volKey, err := common.VolumeIDToKey(volumeID)
+	volKey, _, err := common.VolumeIDToKey(volumeID)
```
msau42: Thinking about the case where you have pv1 and pv2 on the same disk, but different partitions.  All the consumers of those pvs must be scheduled on the same node, either by using Pod affinity on hostname, or only use a single Pod.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -295,7 +295,7 @@ func (gceCS *GCEControllerServer) ControllerUnpublishVolume(ctx context.Context,
 		return nil, status.Error(codes.InvalidArgument, "ControllerUnpublishVolume Node ID must be provided")
 	}
 
-	volKey, err := common.VolumeIDToKey(volumeID)
+	volKey, _, err := common.VolumeIDToKey(volumeID)
```
msau42: Similar story for detach.  How do we ensure that all PVs of the same disk (but different partition) are all unmounted before we Detach?

davidz627:
Closing due to: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/83#issuecomment-415839215

Issue #93, opened 2018-08-17
Support `fs type` parameter


davidz627:
Support changing filesystem type instead of just always defaulting to ext4

PR #102, opened 2018-08-24
Don't FS Type Parameter

davidz627:
No need to support FS Type parameter, it is already supported as a first class field in CSI through volume capabilities.

This PR just removes the TODO
Fixes: #93 

/assign @msau42


davidz627:
No need to support FS Type parameter, it is already supported as a first class field in CSI through volume capabilities.

This PR just removes the TODO

msau42:
don't forget to add it to the list of parameters we need translation logic for

/lgtm

Issue #96, opened 2018-08-17
NodeUnpublishVolume should check if mount path still exists before attempting to unmount


davidz627:
see title

PR #103, opened 2018-08-24
Made unmounting more robust

davidz627:
Fixes: #96 

/assign @msau42


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -58,3 +63,97 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
 	}
 	return resp, err
 }
+
+// UnmountMountPoint is a common unmount routine that unmounts the given path and
```
msau42: This looks like it's straight from kubernetes volume util.  Is there anyway we can use that instead?

davidz627: yea, I was wondering whether I should copy-paste or just import. Since that file has a bunch of random very Kubernetes specific code too and I'm not sure what the status is of that code.

I think I overheard at some point we're only "supposed to" import kubernetes code if its in "staging" (im not 100% sure what that means)

msau42: We're in the process of refactoring the mount library to make it easier to import.  If you import it now, it means that you will probably pull in a lot of unnecessary kubernetes dependencies.

davidz627:
@msau42 changed this to an import instead

pkg/gce-pd-csi-driver/node.go
```diff
@@ -66,7 +67,7 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Volume Capability must be provided")
 	}
 
-	notMnt, err := ns.Mounter.Interface.IsLikelyNotMountPoint(targetPath)
+	notMnt, err := ns.Mounter.Interface.IsNotMountPoint(targetPath)
```
msau42: in tree uses the "likely" variant.  Is there a reason why we need to switch to the unlikely variant?

davidz627: If there is a mount point here its likely that its from a `NodePublish` bind-mount. the "likely" variant doesn't detect bind-mounts and therefore can't find the preexisting mount. I think it should be changed in-tree as well.

msau42: The comment is misleading.

It doesn't detect if the root volume itself is a bind mount, ie you bind mounted a hostpath volume and that's the volume.

What the IsLikelyNotMountPoint check does is compare the filesystem of the volume with its parent.  And if they're different filesystems then it says it's a mount point.

davidz627: reverted

pkg/gce-pd-csi-driver/node.go
```diff
@@ -133,9 +134,7 @@ func (ns *GCENodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeU
 		return nil, status.Error(codes.InvalidArgument, "NodeUnpublishVolume Target Path must be provided")
 	}
 
-	// TODO(#96): Check volume still exists
-
-	err := ns.Mounter.Interface.Unmount(targetPath)
+	err := volumeutils.UnmountMountPoint(targetPath, ns.Mounter.Interface, true /* bind mount */)
```
msau42: I think the last arg can also be false

Issue #101, opened 2018-08-24
Refactor E2E Tests so that they do not redeploy a new driver for each test, instead just once for each suite should be enough


davidz627:
/assign

Issue #97, opened 2018-08-17
Pick a better boot disk image for E2E Test instances


davidz627:
Currently is the ml-images tensorflow image. We could probably pick a more suitable image with all the requirements

PR #104, opened 2018-08-27
Test contexts are now initialized once on startup in parallel instead of per-test, Chose a better boot disk image. 

davidz627:
Fixes: #101
Fixes: #97 

/assign @msau42

### DIFF
--- a/test/e2e/tests/setup_e2e_test.go
+++ b/test/e2e/tests/setup_e2e_test.go
@@ -35,7 +35,7 @@ var (
 	runInProw       = flag.Bool("run-in-prow", false, "If true, use a Boskos loaned project and special CI service accounts and ssh keys")
 	deleteInstances = flag.Bool("delete-instances", false, "Delete the instances after tests run")
 
-	testInstances      = []*remote.InstanceInfo{}
+	testContexts       = []*remote.TestContext{}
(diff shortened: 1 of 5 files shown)


davidz627:
@msau42 sorry this PR kind of became multifaceted, I wanted to make the tests faster and I just couldn't stop

test/e2e/tests/setup_e2e_test.go
```diff
@@ -68,27 +71,41 @@ var _ = BeforeSuite(func() {
...
-			i.DeleteInstance()
+			tc.Instance.DeleteInstance()
 		}
 	}
 })
+
+func getRandomTestContext() *remote.TestContext {
+	Expect(testContexts).ToNot(BeEmpty())
+	rn := rand.Intn(len(testContexts))
+	return testContexts[rn]
```
msau42: Is there any issue if multiple tests run in parallel and choose the same context?

davidz627: I don't believe so, since the driver can handle multiple incoming requests and all volume's should be unique due to UUID's. If there is a parallelism issue here it would be with the driver, not the tests and that is something we'd want to catch if it existed.

Issue #69, opened 2018-07-17
Verify udevadm works as intended in driver


davidz627:
There exists some manual udevadm triggering in the mount manager code due to a preexisting issue of devices not showing up properly. We need to verify that this udevadm trigger behavior is still called and works properly in this driver

msau42:
This also needs to be ported: https://github.com/kubernetes/kubernetes/pull/66832

davidz627:
Porting the fix is in this issue: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/106

PR #107, opened 2018-08-27
Fix so that udevadm can be run from inside kubernetes container on the host

davidz627:
Fixes: #69

/assign @msau42

### DIFF
--- a/deploy/kubernetes/dev/node.yaml
+++ b/deploy/kubernetes/dev/node.yaml
@@ -51,7 +51,16 @@ spec:
             - name: plugin-dir
               mountPath: /csi
             - name: device-dir
-              mountPath: /host/dev
+              mountPath: /dev
+            # The following mounts are required to trigger host udevadm from container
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/dev/node.yaml
```diff
@@ -49,8 +49,17 @@ spec:
               mountPropagation: "Bidirectional"
             - name: plugin-dir
               mountPath: /csi
+            # The following mounts are required to trigger host udevadm from container
             - name: device-dir
-              mountPath: /host/dev
+              mountPath: /dev
```
msau42: Can you add a comment on what each of these are needed for?

msau42: Can all of these be readonly?

msau42:
let me know when tests have been done (current e2es don't test the driver in a container or kubernetes)

davidz627:
@msau42 now running Kubernetes integration test and it is passing :+1:

PR #110, opened 2018-09-06
Add VolumeSnapshot Support

jingxu97:
This PR adds the support for volume snapshots in csi driver


pkg/common/utils.go
```diff
@@ -64,6 +65,18 @@ func VolumeIDToKey(id string) (*meta.Key, error) {
...
 
+func SnapshotIDToKey(id string) (string, error) {
+	splitId := strings.Split(id, "/")
+	if len(splitId) != snapshotTotalElements {
+		return "", fmt.Errorf("failed to get id components. Expected projects/{project}/global/snapshot/{name}. Got: %s", id)
+	}
+	if splitId[volIDToplogyKey] == "global" {
+		return splitId[snapshotTotalElements-1],  nil
+	} else {
+		return "", fmt.Errorf("could not get id components, expected either zones or regions, got: %v", splitId[volIDToplogyKey])
```
davidz627: nit: expected global, not `volIDTopologyKey`

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -456,3 +474,116 @@ func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cloud *CloudProvider) CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*compute.Snapshot, error) {
+	switch volKey.Type() {
+	case meta.Zonal:
+		snapshot, err := cloud.createSnapshotOfZonalDisk(ctx, volKey, snapshotName)
```
davidz627: nit: `createZonalDiskSnapshot`? and you can just return on the same line here: `return cloud....`

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -456,3 +474,116 @@ func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone
...
+	}
+
+	return cloud.waitForSnapshotCreation(ctx, snapshotName)
+
+}
+
+func (cloud *CloudProvider) waitForSnapshotCreation(ctx context.Context, snapshotName string) (*compute.Snapshot, error) {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+	timer := time.NewTimer(waitForSnapshotCreationTimeOut)
```
davidz627: curious as to why we choose 2 minutes here and why we use a separate timeout instead of the parent context

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -456,3 +474,116 @@ func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone
...
+			if err != nil {
+				glog.Warningf("Error in getting snapshot %s, %v", snapshotName, err)
+			} else if snapshot != nil {
+				if snapshot.Status != "CREATING" {
+					glog.V(5).Infof("Snapshot %s status is %s", snapshotName, snapshot.Status)
+					return snapshot, nil
+				} else {
+					glog.V(5).Infof("Snapshot %s is still creating ...", snapshotName)
+				}
+			}
```
davidz627: `else` case where `err == nil` and `snapshot == nil`?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -456,3 +474,116 @@ func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone
...
+
+	for {
+		select {
+		case <-ticker.C:
+			glog.V(5).Infof("Checking GCE Snapshot %s.", snapshotName)
+			snapshot, err := cloud.GetSnapshot(ctx, snapshotName)
+			if err != nil {
+				glog.Warningf("Error in getting snapshot %s, %v", snapshotName, err)
+			} else if snapshot != nil {
+				if snapshot.Status != "CREATING" {
```
davidz627: instead of negative check for `CREATING` is there something specific we're looking for like `READY`?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -431,14 +432,118 @@ func (gceCS *GCEControllerServer) ControllerGetCapabilities(ctx context.Context,
...
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Unknown create snapshot error: %v", err))
+	}
+	t, err := time.Parse(time.RFC3339, snapshot.CreationTimestamp)
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to covert creation timestamp: %v", err))
+	}
+	createResp := &csi.CreateSnapshotResponse{
+		Snapshot: &csi.Snapshot{
+			SizeBytes:      	common.GbToBytes(snapshot.DiskSizeGb),
+			Id:                 cleanSelfLink(snapshot.SelfLink),
```
davidz627: just making sure `cleanSelfLink` works properly here, since maybe snapshot selfLinks have different prefixes

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -431,14 +432,118 @@ func (gceCS *GCEControllerServer) ControllerGetCapabilities(ctx context.Context,
...
+
+	return &csi.DeleteSnapshotResponse{}, nil
 }
 
 func (gceCS *GCEControllerServer) ListSnapshots(ctx context.Context, req *csi.ListSnapshotsRequest) (*csi.ListSnapshotsResponse, error) {
+	glog.V(4).Infof("ListSnapshots called with request %v", *req)
+
+	// In the first version, only support ListSnapshots if snapshot id is specified.
+	// If no snapshot id is specified, return Unimplemented error code
+	if snapshotId := req.GetSnapshotId(); snapshotId != "" {
```
davidz627: nit: can you reverse the block here. `if snapshotId == "" { then error }` and have the main functionality be outside of the conditional

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -431,14 +432,118 @@ func (gceCS *GCEControllerServer) ControllerGetCapabilities(ctx context.Context,
...
+		key, err := common.SnapshotIDToKey(snapshotId)
+		if err != nil {
+			// Cannot get snapshot ID from the passing request
+			return nil, status.Error(codes.InvalidArgument, "snapshot id is not right")
```
davidz627: nit: add debug string such as the snapshotId provided

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -456,3 +474,116 @@ func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone
...
+		return err
+	}
+	err = cloud.waitForGlobalOp(ctx, op)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cloud *CloudProvider) CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*compute.Snapshot, error) {
```
davidz627: Could you also add an E2E test or two to test this new functionality? Feel free to let me know if you want a walkthrough of how the framework works!

pkg/common/utils.go
jingxu97: I used volIDTopologyKey here because it would be same for snapshot. I can add a new one for snapshotIDTopologyKey too.

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: sure, will add it in a separate PR

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: done

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: Because normally it only takes seconds to create snapshot, two minutes should be enough. which parent context you mean?

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: The status could be CREATING, DELETING, FAILED, READY, or UPLOADING. Except CREATING, all other states mean that snapshot is already been taken or failed to create.

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: else case should be err == nil, snapshot != nil, snapshot.Status == "CREATING", just use it for logging.

pkg/gce-pd-csi-driver/controller.go
jingxu97: for now, snapshot has the same link format as volume

pkg/gce-pd-csi-driver/controller.go
jingxu97: modify the code here in the new commit

pkg/gce-pd-csi-driver/controller.go
jingxu97: done

pkg/gce-cloud-provider/compute/gce-compute.go
davidz627: could you open an issue and link to this pr

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -68,8 +68,248 @@ var (
...
+			continue
+		}
+		if tc.expErrCode != codes.OK {
+			t.Fatalf("Expected error: %v, got no error", tc.expErrCode)
+		}
+
+	}
+}
+
+func TestListSnapshotsArguments(t *testing.T) {
```
davidz627: what about tests where we dont specify the ID and need to list all snapshots

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go
```diff
@@ -1622,8 +1622,11 @@ var _ = DescribeSanity("CreateSnapshot [Controller Server]", func(sc *SanityCont
 		Expect(snap1).NotTo(BeNil())
 		verifySnapshotInfo(snap1.GetSnapshot())
 
+		volume2, err := c.CreateVolume(context.Background(), MakeCreateVolumeReq(sc, "CreateSnapshot-volume-3"))
+		Expect(err).ToNot(HaveOccurred())
+
 		By("creating a snapshot with the same name but different volume source id")
-		req2 := MakeCreateSnapshotReq(sc, "CreateSnapshot-snapshot-2", "test001", nil)
+		req2 := MakeCreateSnapshotReq(sc, "CreateSnapshot-snapshot-2", volume2.GetVolume().GetId(), nil)
```
davidz627: was this a `dep` update? If this is an error in the sanity package I'd rather we do an update in the package itself then do a dep update here

pkg/gce-pd-csi-driver/controller_test.go
jingxu97: The the last test "no id" is to return all the snapshots.

vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go
jingxu97: PR to update the sanity test https://github.com/kubernetes-csi/csi-test/pull/116

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: actually I just added the e2e tests in a second commit. PTAL.

Issue #109, opened 2018-09-04
Resolve Disk Name Collisions


davidz627:
Something like this PR: https://github.com/kubernetes/kubernetes/pull/65198

The issue is that if the disk name is the same for a regional and zonal PD they will have a naming collision when attaching to the node. We need to disambiguate the two types of disks so that the names don't collide on the node.

PR #111, opened 2018-09-14
Resolve disk name collisions for attached regional disks

davidz627:
Fixes: #109 

/assign @msau42 
/assign @verult

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -177,7 +177,12 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
 	// TODO(#83): Get real partitions
 	partition := ""
 
-	devicePaths := ns.DeviceUtils.GetDiskByIdPaths(volumeKey.Name, partition)
+	deviceName, err := common.GetDeviceName(volumeKey)
+	if err != nil {
+		status.Error(codes.Internal, fmt.Sprintf("error getting device name: %v", err))
+	}
+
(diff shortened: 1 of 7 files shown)


davidz627:
ping @msau42 @verult PTAL when you have some time

pkg/gce-pd-csi-driver/node.go
```diff
@@ -177,7 +177,12 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
 	partition := ""
 
-	devicePaths := ns.DeviceUtils.GetDiskByIdPaths(volumeKey.Name, partition)
+	deviceName, err := common.GetDeviceName(volumeKey)
+	if err != nil {
+		status.Error(codes.Internal, fmt.Sprintf("error getting device name: %v", err))
+	}
+
+	devicePaths := ns.DeviceUtils.GetDiskByIdPaths(deviceName, partition)
 	devicePath, err := ns.DeviceUtils.VerifyDevicePath(devicePaths)
```
msau42: did you merge that udev fix that used scsi_id and validated the device name?  I wonder if that will have issues with this change.

davidz627: hmm I'm not really sure what you're talking about. I have a vague memory of something like that but I can't find a PR about it...

davidz627: do you mean the change you made to k8s/k8s where you try to rerun `udevadm trigger` a bunch of times after doing some file parsing for validation? If so, then no I haven't merged that

msau42: ok we may have to revisit this in that PR then

PR #112, opened 2018-09-17
[WIP] RBAC roles and rolebindings for external provisioner leader election

verult:
This past quarter, leader election was implemented for external-provisioner, which now requires full ReadWrite permissions for Endpoints objects. This is a bit scary since a compromised provisioner can then reroute network traffic. PR https://github.com/kubernetes-incubator/external-storage/pull/957 introduced the ability to make external-provisioner leader election. This PR is a draft of required RBAC changes if leader election were to be kept enabled.

Do we want to disable leader election? Does this RBAC role pose a risk for GCE? Or maybe the provisioner should use a different leader election technique altogether, and we could push for that in future releases and have a temporary measure for the time being.

@davidz627 @msau42 @saad-ali @jingxu97

### DIFF
--- a/deploy/kubernetes/dev/setup-cluster.yaml
+++ b/deploy/kubernetes/dev/setup-cluster.yaml
@@ -68,3 +68,28 @@ roleRef:
   kind: ClusterRole
   name: system:csi-external-provisioner
   apiGroup: rbac.authorization.k8s.io
+---
+
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: leader-locking-csi-external-provisioner
+rules:
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    verbs: ["get", "list", "watch", "create", "update", "patch"]
+
+---
+
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: leader-locking-csi-external-provisioner
+subjects:
+  - kind: ServiceAccount
+    name: csi-controller-sa
+    namespace: default
+roleRef:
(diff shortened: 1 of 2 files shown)


msau42:
What is the leader election for? Multiple external provisioners?  If so, I would say let's disable it as it's not relevant for the CSI driver deployment.

verult:
I'm not too familiar with exactly what leader election is doing. @saad-ali ?

msau42:
I had a discussion on slack this morning with @jsafrane.  It looks like it's used for serializing Provision() operations in a single controller.  I feel like using leader election to serialize within a single instance is overkill and not the correct solution.  CreateVolume() calls are supposed to be idempotent so if it is called while another operation is currently in progress, the driver should appropriately handle that.  But we need the external-provisioner to be able to pass in the same volume_id every time.  How does the volume_id get generated?

davidz627:
@verult any update on whether this is necessary?

msau42:
To clarify, the rbac rules should only give permissions within the driver namespace, not to the entire cluster (that's what ClusterRole does).  My understanding was that the leader election was added to be able to scale the provisioner to multiple instances. 

I'm not sure how much we need it right now.  We've been fine so far with single provisioner.  Maybe let's consider it once we have a better benchmarking/monitoring story to see the difference.

verult:
The leader election logic is going to change in the next quarter, so the RBAC rules here are going to change too. Let's disable leader election for now. That being said it looks like there's a bug in the disabling logic (my understanding is right now the external provisioner has to be configured internally to disable leader election, rather than through a user-facing interface like a CLI flag). I'll follow up on that.

PR #117, opened 2018-09-27
[WIP] E2E Instance Up Fix

davidz627:
something is wrong with bringing the instance up in the e2e test


davidz627:
nvm looks like its fine

Issue #121, opened 2018-10-02
Minimum volume size should be less than 5Gi


davidz627:
Driver fails if request for volume less than 5Gi, however, it is possible to create GCE PD Volumes smaller than that, we should not fail in that case

/assign

davidz627:
Fixed by: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/114

PR #114, opened 2018-09-26
Adding Kubernetes Integration Test Framework

davidz627:
Fixes: #22

This PR adds a full integration test framework. It is ready for review.

Couple WIP items:

- [x] Determine and set staging image location for running in CI
- [ ] Attempt to parse manifests for more robust image substitutions
- [x] Create CI entry to run the script and test PROW integration (https://github.com/kubernetes/test-infra/pull/9596)

/kind feature
/assign @msau42

PR #122, opened 2018-10-02
Change minimum volume size to 1Gi to be in line with GCE PD

davidz627:
Fixes: #121 

/assign @msau42


davidz627:
Merged in: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/114

PR #123, opened 2018-10-02
Support partitions in VolumeAttributes

davidz627:
Fixes: #83 

Tested with Kubernetes and it works.
The deployment yamls used:
https://gist.github.com/davidz627/ae1d3071913eda1f6dacfe59779073cf

/assign @msau42


msau42:
I'm surprised your yaml works. The pods are likely to be scheduled to 2 different nodes

davidz627:
@msau42 It was just for a quick test so I got lucky. But yea in a real situation you'd want to nodeselector

PR #129, opened 2018-10-09
Support UNSPECIFIED zone and project in volume ID. Static driver name. To prep for migration

davidz627:
When using Kubernetes does migration it may not have access to zone or region of the disk when translating to CSI. This is because k8s GCE PD object uniquely (sort of) identifies disks by PD Name only. We need to support this functionality so in case of an unspecified zone the driver will try to find out where it is by querying all zones.

In the case that the same PD Name is defined in multiple zones, the behavior is undefined. This is consistent with behavior inside Kubernetes and so is sufficient for Migration.

The driver name must also be a well-known name.

/assign @msau42 
/cc @saad-ali

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -27,7 +27,9 @@ import (
 const (
 	// Volume ID Expected Format
 	// "projects/{projectName}/zones/{zoneName}/disks/{diskName}"
+	volIDZonalFmt = "projects/%s/zones/%s/disks/%s"
 	// "projects/{projectName}/regions/{regionName}/disks/{diskName}"
+	volIDRegionalFmt   = "projects/%s/regions/%s/disks/%s"
 	volIDToplogyKey    = 2
 	volIDToplogyValue  = 3
 	volIDDiskNameValue = 5
@@ -69,6 +71,13 @@ func VolumeIDToKey(id string) (*meta.Key, error) {
 	} else {
 		return nil, fmt.Errorf("could not get id components, expected either zones or regions, got: %v", splitId[volIDToplogyKey])
 	}
+}
+
+func GenerateUnderspecifiedVolumeID(diskName string, isZonal bool) string {
+	if isZonal {
+		return fmt.Sprintf(volIDZonalFmt, UnspecifiedValue, UnspecifiedValue, diskName)
+	}
+	return fmt.Sprintf(volIDRegionalFmt, UnspecifiedValue, UnspecifiedValue, diskName)
 }
 
 func SnapshotIDToKey(id string) (string, error) {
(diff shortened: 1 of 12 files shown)


pkg/common/utils.go
```diff
@@ -71,6 +73,13 @@ func VolumeIDToKey(id string) (*meta.Key, error) {
 	}
 }
 
+func GenerateUnderspecifiedVolumeID(diskName string, isZonal bool) string {
+	if isZonal {
```
msau42: How can you know that a disk is zonal or regional and not know the zone/region?

We do not support regional disks for inline volumes

msau42: Ok I see that this is only being used by the e2e test.  Can we move this function there instead?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -191,6 +191,11 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 		return &csi.DeleteVolumeResponse{}, nil
 	}
 
+	volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, volKey)
```
msau42: unit tests for success and error cases?

pkg/common/utils.go
davidz627: I actually want to use it in the translation library as well. Apparently we can tell whether a disk is zonal or regional based on the PV (which we have access to in the translation library). If it is inline, its always zonal

msau42: If you're translating from a PV, you can already extract the zones from the label.

davidz627: although its true that I guess in that case we will know the zone/region... is it true that we can always get the zone/zones on the PV for GCE PD? I think it is based on the code I've seen. In that case we may only hit the UNSPECIFIED case for in-line volumes. I still want this util function to distinguish between zonal/regional regardless

msau42: If the labels were missing, then the PV would be likely non-functional in regional clusters.  Reasons why it may be missing:
* User modified the PV afterwards to remove/edit the labels
* PV was created in an old version that didn't have the admission controller running.

davidz627: Ok cool, either way this PR as a whole makes the driver resilient to unspecified values. In the case where they are specified there is no difference in functionality

pkg/gce-pd-csi-driver/controller.go
davidz627: unit tests for `DeleteVolume` or `RepairUnderspecifiedVolumeKey`?

msau42: DeleteVolume, when Repair returns success and error.

Ideally, unit tests for Repair would be good too, but that would also require mocking the cloud sdk layer (which I think is possible in the in-tree library)

davidz627:
@msau42 comments have been addressed

Issue #130, opened 2018-10-17
Make Topology Label prefix configurable


davidz627:
The topology label might need to be different in different deployments. It should be configurable on the driver, the kubernetes deployment can then use:
```
LabelZoneFailureDomain = "failure-domain.beta.kubernetes.io/zone"
```
which is a well known kubernetes topology label that has existing integrations

davidz627:
we've come up with a "gcp wide" topology zone label `topology.gke.io/zone`. Since this is a GCP only driver this should be sufficient

PR #132, opened 2018-10-17
Made topology configurable and set default to Kubernetes well-known zone label

davidz627:
Fixes: #130 

/assign @msau42

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/golang/glog"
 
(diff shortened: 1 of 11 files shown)


davidz627:
@msau42 not sure whether we should default to Kubernetes Label or if it should be empty and be a required flag. Leaning towards the latter but curious what you think

msau42:
/hold
We still have some discussions going on about csi topology labeling

davidz627:
we've come up with a "gcp wide" topology zone label. Since this is a GCP only driver this should be sufficient

PR #141, opened 2018-10-27
Add gke prow config

msau42:
Also add option to not build the driver from source

### DIFF
--- a/deploy/kubernetes/overlays/prow-gke-staging/controller_images.yaml
+++ b/deploy/kubernetes/overlays/prow-gke-staging/controller_images.yaml
@@ -1,0 +1,14 @@
+kind: StatefulSet
+apiVersion: apps/v1
+metadata:
+  name: csi-gce-pd-controller
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-provisioner
+          image: gcr.io/gke-release-staging/csi-provisioner:v0.4.1-gke.0
+        - name: csi-attacher
+          image: gcr.io/gke-release-staging/csi-attacher:v0.4.1-gke.0
+        - name: gce-pd-driver
+          image: gcr.io/gke-release-staging/gcp-compute-persistent-disk-csi-driver:v0.2.0-gke.0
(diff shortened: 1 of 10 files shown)


test/k8s-integration/main.go
```diff
@@ -100,14 +101,16 @@ func handle() error {
 			}
 		}()
 
-		*stagingImage = fmt.Sprintf("gcr.io/%s/gcp-persistent-disk-csi-driver", project)
+		if !*skipDriverBuild {
```
davidz627: lot of double negatives here. Looks like every time we use `skipDriverBuild` we do `!skipDriverBuild`, lets turn all these double negatives into a positive and just use `doDriverBuild` or something.

test/run-k8s-integration.sh
```diff
@@ -1,9 +1,16 @@
 #!/bin/bash
 
+# Optional environment variables
+# GCE_PD_OVERLAY_NAME: which Kustomize overlay to deploy with
+# GCE_PD_SKIP_DRIVER_BUILD: if set, don't build the driver from source and just
```
davidz627: same note here about the flag

deploy/kubernetes/overlays/prow-gke-staging/controller_images.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-provisioner
+          image: gcr.io/gke-release-staging/csi-provisioner:v0.4.1-gke.0
+        - name: csi-attacher
+          image: gcr.io/gke-release-staging/csi-attacher:v0.4.1-gke.0
+        - name: gce-pd-driver
+          image: gcr.io/gke-release-staging/gcp-compute-persistent-disk-csi-driver:v0.2.0-gke.0
```
davidz627: is the gke patch necessary/automatic?

davidz627: also, are we going to add a test for the latest prod version as well?

msau42: right now it's manual. every time we want to qualify a combination of sidecars and containers, we need to update this config with the versions of each container we want to use.

eventually i hope to make it automatic similar to how it works in kubernetes, where there's a file that stores the latest tagged versions.

I will update the base (stable) config with the prod images once we qualify them and make them available.

test/k8s-integration/main.go
msau42: done

test/run-k8s-integration.sh
msau42: done

davidz627:
nice! Please squash but LGTM

msau42:
I fixed the comment in the bash script

PR #143, opened 2018-10-30
Make boskos resource name configurable

msau42:


msau42:
Wait for #141 to merge first
/hold

davidz627:
@msau42 when do you need a different boskos resource type? I don't see anywhere where its being used other than "gce-project". Also please rebase

msau42:
gke testing doesn't have a "gce-project" defined, so gke prow config will pass in something else.

PR #153, opened 2018-11-02
Add retry for acquiring a boskos resource

msau42:
If you have multiple periodic jobs that have similar timing, and not enough projects available, then the slightly later job will always fail instead of waiting for a project to get freed up.

### DIFF
--- a/test/e2e/utils/utils.go
+++ b/test/e2e/utils/utils.go
@@ -26,6 +26,7 @@ import (
 	"golang.org/x/oauth2/google"
 	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
 	boskosclient "k8s.io/test-infra/boskos/client"
+	"k8s.io/test-infra/boskos/common"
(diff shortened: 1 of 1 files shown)


test/e2e/utils/utils.go
```diff
@@ -64,20 +65,34 @@ func GCEClientAndDriverSetup(instance *remote.InstanceInfo) (*remote.TestContext
...
+		select {
+		case <-timeout:
+			glog.Fatalf("timed out trying to acquire boskos project")
+		case <-tick:
+			p, err := boskos.Acquire(resourceType, "free", "busy")
+			if err != nil {
+				glog.V(4).Infof("boskos failed to acquire project: %v", err)
+			}
+			if p == nil {
+				glog.V(4).Infof("boskos does not have a free %s at the moment", resourceType)
```
davidz627: should this be a warning instead

msau42: done

msau42:
debugging e2e test failure
/hold

Issue #142, opened 2018-10-27
Simplify arguments in yamls


msau42:
Like in https://github.com/kubernetes/kubernetes/pull/69868/commits/dcbcef13b4955e10eed3f76b65f5f0789d6d1101, there's not much gain by defining env variables for hard coded values that are passed in as arguments.

PR #154, opened 2018-11-02
Remove hardcoded environment variables from spec

msau42:
Fixes: #142

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -12,17 +12,15 @@ spec:
       labels:
         app: gcp-compute-persistent-disk-csi-driver
     spec:
-      serviceAccount: csi-node-sa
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/base/node.yaml
```diff
@@ -18,11 +18,9 @@ spec:
           image: quay.io/k8scsi/driver-registrar:v0.4.1
           args:
             - "--v=5"
-            - "--csi-address=$(ADDRESS)"
+            - "--csi-address=/var/lib/kubelet/plugins/com.google.csi.gcepd/csi.sock"
```
davidz627: I know this is unrelated but since you're touching these files can you also change the `serviceAccount` field in this yaml to the non-deprecated `serviceAccountName` :) :) :)

msau42: done

PR #155, opened 2018-11-08
fix retry logic

msau42:
it didn't actually retry

### DIFF
--- a/test/e2e/utils/utils.go
+++ b/test/e2e/utils/utils.go
@@ -67,21 +67,24 @@ func GCEClientAndDriverSetup(instance *remote.InstanceInfo) (*remote.TestContext
 
 // getBoskosProject retries acquiring a boskos project until success or timeout
 func getBoskosProject(resourceType string) *common.Resource {
-	timeout := time.After(30 * time.Minute)
-	tick := time.After(1 * time.Minute)
+	timer := time.NewTimer(30 * time.Minute)
+	defer timer.Stop()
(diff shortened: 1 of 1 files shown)


test/e2e/utils/utils.go
```diff
@@ -80,8 +80,9 @@ func getBoskosProject(resourceType string) *common.Resource {
 			}
 			if p == nil {
```
davidz627: think this should have been an `if, else if, else`

test/e2e/utils/utils.go
```diff
@@ -80,8 +80,9 @@ func getBoskosProject(resourceType string) *common.Resource {
...
 				glog.Warningf("boskos does not have a free %s at the moment", resourceType)
+			} else {
+				return p
 			}
-			return p
```
davidz627: not convinced this fixes it as `if err != nil || p== nil` we will still return `p`. Maybe there should be a `continue` in each of the error blocks so it will go to the next `tick/timeout` loop?

davidz627:
also I think you need to use `tickers` like here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/e28864bf4c995eb6ebfb3d657bb5c931da28f2ac/pkg/gce-cloud-provider/compute/gce-compute.go#L606

time.After just pings once 😂

msau42:
Fixed

davidz627:
hopefully this works.
/lgtm

PR #156, opened 2018-11-09
Add the cluster role for snapshotter

jingxu97:
This PR adds back the cluster role for external snapshotter


davidz627:
Cool! It's working in my local test.
Thanks for working on this
/lgtm
/approve

PR #159, opened 2018-11-14
Update CSI Spec dependency to v1.0.0-rc2

davidz627:
/assign @msau42 @saad-ali 

xref https://github.com/kubernetes/kubernetes/pull/71020


pkg/common/utils.go
```diff
@@ -133,3 +136,11 @@ func GetDeviceName(volKey *meta.Key) (string, error) {
 func CreateNodeID(project, zone, name string) string {
 	return fmt.Sprintf(nodeIDFmt, project, zone, name)
 }
+
+func TimeToTimestamp(t time.Time) *timestamp.Timestamp {
```
saad-ali: Use the `github.com/golang/protobuf/ptypes` library. It has a method to do this correctly: `func TimestampProto(t time.Time) (*tspb.Timestamp, error)`

davidz627:
/hold

Needs to be updated again after csi-test is cut

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -152,7 +152,7 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 	content := req.GetVolumeContentSource()
 	if content != nil {
 		if content.GetSnapshot() != nil {
-			snapshotId = content.GetSnapshot().GetId()
+			snapshotId = content.GetSnapshot().GetSnapshotId()
```
saad-ali: Add a TODO or open a bug to add support for "Volume Source" (cloning) which is new in 1.0.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -152,7 +152,7 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
 		}
 	}
```
saad-ali: Modify `ControllerServiceCapability` to return new `PUBLISH_READONLY` capability.

See https://github.com/container-storage-interface/spec/pull/310/files for details

davidz627:
@saad-ali resolved your comments

davidz627:
I don't think the snapshot implementation is broken, I've done a small refactor to ensure that it works with the new spec. @jingxu97 @xing-yang can verify at some later point

saad-ali:
@davidz627 Got it, thanks for the clarification.

davidz627:
Final TODOs before unholding:

- [ ] Update all sidecar containers to v1.0.0 (in dev or base?)
- [ ] Verify everything works end to end on k8s

PR #163, opened 2018-11-15
Fix project setup script

saad-ali:
Project setup script kept failing for me without these changes with:

```
./setup-project.sh: line 22: /usr/local/.../go/src/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/deploy/common.sh: No such file or directory
```


msau42:
I think the alias is needed for CI.  I just created the same directory structure on my local workspace

saad-ali:
Ah ok, that's annoying.

PR #168, opened 2018-11-15
Strict Kustomize Patching

davidz627:
Make it so that `base != stable`, this makes versioning crystal clear

/assign @msau42

### DIFF
--- a/deploy/kubernetes/overlays/dev/controller_images.yaml
+++ b/deploy/kubernetes/overlays/dev/controller_images.yaml
@@ -6,6 +6,10 @@ spec:
   template:
     spec:
       containers:
+        - name: csi-provisioner
+          image: gcr.io/gke-release/csi-provisioner:v0.4.1-gke.0
+        - name: csi-attacher
+          image: gcr.io/gke-release/csi-attacher:v0.4.1-gke.0
         - name: csi-snapshotter
           imagePullPolicy: Always
           image: quay.io/k8scsi/csi-snapshotter:v0.4.0
@@ -21,3 +25,4 @@ spec:
         - name: gce-pd-driver
           imagePullPolicy: Always
           image: gcr.io/dyzz-csi-staging/csi/gce-pd-driver:latest
+
--- a/deploy/kubernetes/overlays/prow-head-template/node_images.yaml
+++ b/deploy/kubernetes/overlays/prow-head-template/node_images.yaml
@@ -8,3 +8,5 @@ spec:
       containers:
         - name: gce-pd-driver
           image: PROW_GCEPD_IMAGE:latest
+        - name: csi-driver-registrar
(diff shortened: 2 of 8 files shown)


deploy/kubernetes/overlays/dev/controller_images.yaml
```diff
@@ -6,6 +6,12 @@ spec:
   template:
     spec:
       containers:
+        - name: csi-provisioner
+          image: gcr.io/gke-release-staging/csi-provisioner:v0.4.1-gke.0
```
msau42: gke-release-staging is not publicly accessible, so we need to use gke-release or quay here.  Same with stable and prow-head-template (which is used in oss pull jobs)

davidz627: ah my bad, botched copy-paste job

davidz627:
@msau42 should be fixed

deploy/kubernetes/overlays/dev/controller_images.yaml
```diff
@@ -6,6 +6,12 @@ spec:
   template:
     spec:
       containers:
+        - name: csi-provisioner
```
msau42: Separately we should also consider if we want a job that tests latest sidecars too

deploy/kubernetes/overlays/dev/controller_images.yaml
```diff
@@ -6,6 +6,12 @@ spec:
...
+          image: gcr.io/gke-release/csi-provisioner:v0.4.1-gke.0
+        - name: csi-attacher
+          image: gcr.io/gke-release/csi-attacher:v0.4.1-gke.0
+        - name: gce-pd-driver
```
msau42: there's already a gce-pd-driver entry below

deploy/kubernetes/overlays/prow-head-template/node_images.yaml
```diff
@@ -8,3 +8,5 @@ spec:
       containers:
         - name: gce-pd-driver
           image: PROW_GCEPD_IMAGE:latest
+        - name: gce-pd-driver
```
msau42: remove?

davidz627: should be driver registrar

deploy/kubernetes/overlays/dev/controller_images.yaml
davidz627: oops

deploy/kubernetes/overlays/dev/controller_images.yaml
davidz627: prow-gke-head does that

davidz627:
@msau42 fixed

PR #172, opened 2018-11-15
Update staging PROW config to newest staging containers

davidz627:
/hold
until external container v1.0.0 is tagged
and GCE PD CSI Driver v0.3.0 is tagged

/assign @msau42

PR #173, opened 2018-11-15
v0.3.0 Release Docs and YAML updates

davidz627:
/hold
until https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/172 goes in and soaks for a while
and release images are cut for all external components and GCE PD Driver

/assign @msau42


davidz627:
@msau42 PTAL

PR #181, opened 2018-11-19
Fix kubelet socket path

saad-ali:
Kubernetes/kubernetes PR #70494 changed the kubelet plugin watcher
directory from #70494 `{kubelet_root_dir}/plugins/` to
`{kubelet_root_dir}/plugins_registry/`. This PR fixes the deployment
manifests for the GCE PD CSI Driver.


saad-ali:
Should also be cherry picked to 0.3.0 release branch (the one for k8s v1.13)

PR #184, opened 2018-11-21
Pass ARTIFACTS to k8s e2e

msau42:
Prow will pick up the test logs and parse them and make them available.

Also enable block tests now that the feature is beta.


davidz627:
very cool! Are the block tests not going to fail since they're not supported by GCE PD?

/lgtm

msau42:
Correct the block tests check if the plugin supports block or not.  And if it doesn't support block, it runs a different test that makes sure block fails.

PR #197, opened 2018-12-04
Update staging test versions

msau42:


deploy/kubernetes/overlays/prow-gke-staging/controller_images.yaml
```diff
@@ -7,8 +7,8 @@ spec:
     spec:
       containers:
         - name: csi-provisioner
-          image: gcr.io/gke-release-staging/csi-provisioner:v1.0.0-gke.0
+          image: gcr.io/gke-release-staging/csi-provisioner:v1.0.1-rc1
```
davidz627: not `v1.0.1-gke.0-rc1`?

msau42: there's an issue with our ci config and it's not building the -gke.0 variant for "-rcX" tags

davidz627: is there a tracking issue?

msau42: I don't think we need -gke.x variants for rc's

PR #198, opened 2018-12-10
Fail CreateVolume() when Block volume capability is requested

### LINKED ISSUE #94: Validate Volume Capabilities in Various Calls
Refactor out volume capability validation code, call it in various controller functions to validate that we support the volume capabilities before doing any controller calls

### LINKED ISSUE #196: CreateVolume doesn't check if Block volume capability is requested
So the volume will be erroneously provisioned.  It should return error for block until we implement it.

verult:
Fixes #196, partially fixes #94


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -81,7 +81,13 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume Request Capacity is invalid: %v", err))
 	}
 
-	// TODO(#94): Validate volume capabilities
+	// TODO(#94): Validate AccessModes in VolumeCapabilities
+	for _, capability := range volumeCapabilities {
+		if blk := capability.GetBlock(); blk != nil {
+			// TODO(#64): Block volume support
+			return nil, status.Error(codes.Unimplemented, fmt.Sprintf("Block volume support is not yet implemented"))
```
msau42: lgtm, can you also add a unit test

msau42:
@verult can you cherry pick this back to the release-0.3 branch?

Issue #180, opened 2018-11-18
[IMPORTANT] Take advantage of `csi_secret` in CSI 1.0


saad-ali:
CSI 1.0 decorates sensitive fields with `csi_secret`. Let's take advantage of this feature to programmatically ensure no sensitive fields are ever logged by this driver.

saad-ali:
CSI 1.0 decorates senstive fields with `csi_secret`. Let's take advantage of this feature to programmatically ensure no sensitive fields are ever logged by this driver.

davidz627:
Marking this as high priority so that we dont have security issues. GCE PD doesn't actually accept any secrets but if a user mistakenly sends them in, we still don't want to log them

PR #225, opened 2019-03-08
Strip secret fields from GRPC loggin

davidz627:
GCE PD Driver does not use secret fields, but this is cheap and will strip secret fields from logs in case they are accidentally passed to the driver.

Fixes: #180

/assign @msau42 @saad-ali


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -49,12 +50,12 @@ func NewNodeServiceCapability(cap csi.NodeServiceCapability_RPC_Type) *csi.NodeS
 
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
```
msau42: i think there's actually a logGRPC function under csi-lib-utils/rpc if you want

davidz627: it is the client side version. since pd driver is server side we have to use a different logGRPC

Issue #236, opened 2019-03-15
Maybe a manual "how to run the driver for non Go guru"?


de1m:
Hi,
I don't knowing the golang. I've tryid to install the driver, but I get the error below. Can someone write me the steps, how to install the driver or exactly say, how to config the go environment to compile and run the driver?
Maybe it is possible to start it's throuht a docker image?
```
de1m@comp0:~/workspace/gcp/gcp-compute-persistent-disk-csi-driver$ echo $GOPATH
/home/de1m/go
de1m@comp0:~/workspace/gcp/gcp-compute-persistent-disk-csi-driver$ ./deploy/kubernetes/deploy-driver.sh 
+ readonly NAMESPACE=default
+ NAMESPACE=default
+ readonly DEPLOY_VERSION=stable
+ DEPLOY_VERSION=stable
+ readonly PKGDIR=/home/de1m/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
+ PKGDIR=/home/de1m/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
+ source /home/de1m/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/deploy/common.sh
./deploy/kubernetes/deploy-driver.sh: Line 22: /home/de1m/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/deploy/common.sh: File or Foder not found
```

davidz627:
Hi @de1m,

You will need to clone the repository at
`${HOME}/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver`
instead of your current location of:
`${HOME}/workspace/gcp/gcp-compute-persistent-disk-csi-driver`.

go paths as well as relative pathing in bash are a little finicky. So the location of your repository is important. I will make sure the documentation captures this

de1m:
Thanks it has working

PR #237, opened 2019-03-16
Add more detailed instructions for getting driver code into user environment

davidz627:
Fixes: #236

/assign @msau42


davidz627:
not actually super useful and README has been refactored

Issue #19, opened 2018-06-21
Set MaxVolumesPerNode on NodeGetInfo call based on Node Type


davidz627:
Currently `NodeGetInfoResponse` returns the default of `0` for `MaxVolumesPerNode` so the CO will decide how many volumes can be published on a node.

For GCE we need to return a different number based on node type as the Max Attachable Volumes depends on the number of vCPUs the instance has.

For the actual limits see:
See: https://cloud.google.com/compute/docs/disks/
"persistent disk limits" section

You should be able to GET the instance from the cloud and pull the number of vCPUs from that.
Bonus: We seem to need information from the node object a lot, caching the relevant information somewhere would be nice. Maybe in the `GCENodeServer` object

Mockery-Li:
Hi, I want to take this as my first issue, and I hope to make sure I did not misunderstand the problem.
Is it about modifying `GetMaxVolumesPerNode()` in csi.pb.go (both V0 and V1), to pull the number of vCPUs as the default return value when it knows it's under GCE environment?

davidz627:
Hi @Mockery-Li! Thank you for your interest :)

To implement `MaxVolumesPerNode` you will not have to modify `csi.pb.go`. That is the CSI Specification Protobuf and is generated from the [CSI Specification](https://github.com/container-storage-interface/spec/blob/master/spec.md)

What this issue entails is modifying `NodeGetInfo` in `pkg/gce-pd-csi-driver/node.go` to return a value `MaxVolumesPerNode`.

The max number of nodes can be determined via machine type: see [Persistent Disk Limits section](https://cloud.google.com/compute/docs/disks). You should be able to get the Machine type through the Metadata Service in `pkg/gce-cloud-provider/metadata/metadata.go` (I am not 100% sure about this part).

This should just be a matter of piping the metadata service through to the `Node` service and making the machine type checking call in the `NodeGetInfo` function and casing on the Machine Type.

Hope that is enough information to get you started. Feel free to reach out to me on the Kubernetes Slack (I am `davidz627` there as well) or post questions on this issue if you are wanting guidance. Looking forward to your Pull Request!

msau42:
You can probably also copy a lot of the functionality from [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/gcepd/gce_pd.go#L126)

davidz627:
That's true but I think there are a few differences in the current state (and the driver) from that code.

[...]

PR #240, opened 2019-03-26
Set MaxVolumesPerNode in NodeGetInfo

Mockery-Li:
Fixed: [#19](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/19)
Get volume limits by sending HTTP request to Google metadata server
Tested for several different machine types. But don't know how to integrate it into the current e2e test, and don't know if it is necessary.
@davidz627 @msau42

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -17,6 +17,7 @@ package gceGCEDriver
 import (
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
@@ -41,6 +42,14 @@ type GCENodeServer struct {
 }
 
 var _ csi.NodeServer = &GCENodeServer{}
+
+// The constants are used to map from the machine type to the limit of
+// persistent disks that can be attached to an instance. Please refer to gcloud doc
+// https://cloud.google.com/compute/docs/disks/#pdnumberlimits
+const (
+	volumeLimit16  int64 = 16
+	volumeLimit128 int64 = 128
+)
 
 func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
 	ns.mux.Lock()
@@ -284,15 +293,17 @@ func (ns *GCENodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRe
 
 	nodeID := common.CreateNodeID(ns.MetadataService.GetProject(), ns.MetadataService.GetZone(), ns.MetadataService.GetName())
 
+	volumeLimits, err := ns.GetVolumeLimits()
+
 	resp := &csi.NodeGetInfoResponse{
 		NodeId: nodeID,
 		// TODO(#19): Set MaxVolumesPerNode based on Node Type
 		// Default of 0 means that CO Decides how many nodes can be published
 		// Can get from metadata server "machine-type"
-		MaxVolumesPerNode:  0,
+		MaxVolumesPerNode:  volumeLimits,
 		AccessibleTopology: top,
 	}
-	return resp, nil
+	return resp, err
 }
 
 func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
@@ -301,4 +312,17 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
 
 func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
 	return nil, status.Error(codes.Unimplemented, fmt.Sprintf("NodeExpandVolume is not yet implemented"))
+}
+
+func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
+	var volumeLimits int64
+
+	// Machine-type format: n1-type-CPUS or custom-CPUS-RAM or f1/g1-type
+	machineType := ns.MetadataService.GetMachineType()
+	if strings.HasPrefix(machineType, "n1-") || strings.HasPrefix(machineType, "custom-") {
+		volumeLimits = volumeLimit128
+	} else {
+		volumeLimits = volumeLimit16
+	}
+	return volumeLimits, nil
 }
--- a/pkg/gce-pd-csi-driver/node_test.go
+++ b/pkg/gce-pd-csi-driver/node_test.go
@@ -13,3 +13,76 @@ limitations under the License.
 */
 
 package gceGCEDriver
+
+import (
+	"testing"
+
+	csi "github.com/container-storage-interface/spec/lib/go/csi"
+	"golang.org/x/net/context"
+	metadataservice "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/metadata"
+	mountmanager "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/mount-manager"
+)
+
+func getTestGCEDriver(t *testing.T) *GCEDriver {
+	gceDriver := GetGCEDriver()
+	err := gceDriver.SetupGCEDriver(nil, mountmanager.NewFakeSafeMounter(), mountmanager.NewFakeDeviceUtils(), metadataservice.NewFakeService(), driver, "test-vendor")
+	if err != nil {
+		t.Fatalf("Failed to setup GCE Driver: %v", err)
+	}
+	return gceDriver
+}
+
+func TestNodeGetVolumeLimits(t *testing.T) {
+
+	gceDriver := getTestGCEDriver(t)
+	ns := gceDriver.ns
+	req := &csi.NodeGetInfoRequest{}
+
+	testCases := []struct {
+		name           string
+		machineType    string
+		expVolumeLimit int64
+	}{
+		{
+			name:           "Predifined standard machine",
+			machineType:    "n1-standard-1",
+			expVolumeLimit: volumeLimit128,
+		},
+		{
+			name:           "Predifined micro machine",
+			machineType:    "f1-micro",
+			expVolumeLimit: volumeLimit16,
+		},
+		{
+			name:           "Predifined small machine",
+			machineType:    "g1-small",
+			expVolumeLimit: volumeLimit16,
+		},
+		{
+			name:           "Custom machine with 1GiB Mem",
+			machineType:    "custom-1-1024",
+			expVolumeLimit: volumeLimit128,
+		},
+		{
+			name:           "Custom machine with 4GiB Mem",
+			machineType:    "custom-2-4096",
+			expVolumeLimit: volumeLimit128,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Logf("Test case: %s", tc.name)
+		metadataservice.SetMachineType(tc.machineType)
+		res, err := ns.NodeGetInfo(context.Background(), req)
+		if err != nil {
+			t.Fatalf("Failed to get node info: %v", err)
+		} else {
+			volumeLimit := res.GetMaxVolumesPerNode()
+			if volumeLimit != tc.expVolumeLimit {
+				t.Fatalf("Expected volume limit: %v, got %v, for machine-type: %v",
+					tc.expVolumeLimit, volumeLimit, tc.machineType)
+			}
+			t.Logf("Get node info: %v", res)
+		}
+	}
+}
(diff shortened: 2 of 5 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -302,3 +320,55 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
 func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandVolumeRequest) (*csi.NodeExpandVolumeResponse, error) {
 	return nil, status.Error(codes.Unimplemented, fmt.Sprintf("NodeExpandVolume is not yet implemented"))
 }
+
+func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
+	client := &http.Client{}
+	// Default value 16 for shared-CPU VM
+	volumeLimits := VolumeLimit16
+
+	// Get machine-type from metadata server
```
msau42: Instead of directly calling the http here, I think you can extend https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/metadata/metadata.go.  It looks like the compute metadata service doesn't have a method for machine type, but you could use the general Get() call.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -347,11 +345,13 @@ func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
...
 		return volumeLimits, err
 	}
+	var str = string(body[:])
+	*/
 
 	// Response format: "projects/[NUMERIC_PROJECT_ID]/machineTypes/[MACHINE_TYPE]"
-	var str = string(body[:])
-	splits := strings.Split(str, "/")
-	var machineType = splits[len(splits)-1]
+	fullMachineType := ns.MetadataService.GetMachineType()
```
msau42: Can you also do this parsing inside metadata.go?

Mockery-Li: Sure, I'm working on it.

Mockery-Li: I have updated the code, please help to review it

pkg/gce-pd-csi-driver/node.go
```diff
@@ -42,6 +44,18 @@ type GCENodeServer struct {
 
 var _ csi.NodeServer = &GCENodeServer{}
 
+// The constants are used to map from the machine type (number of CPUs) to the limit of
+// persistent disks that can be attached to an instance. Please refer to gcloud doc
+// https://cloud.google.com/compute/docs/disks/#increased_persistent_disk_limits
```
msau42: Update the page: https://cloud.google.com/compute/docs/disks/#pdnumberlimits

pkg/gce-pd-csi-driver/node.go
```diff
@@ -42,6 +44,18 @@ type GCENodeServer struct {
...
+const (
+	OneCPU               = 1
+	EightCPUs            = 8
+	VolumeLimit16 int64  = 16
+	VolumeLimit32 int64  = 32
+	VolumeLimit64 int64	 = 64
+	VolumeLimit128 int64 = 128
```
msau42: Also from the updated page, it looks like the limits are now:

128 for 1 CPU and above
16 for shared-core

pkg/gce-pd-csi-driver/node.go
```diff
@@ -302,3 +318,28 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
 	return nil, status.Error(codes.Unimplemented, fmt.Sprintf("NodeExpandVolume is not yet implemented"))
 }
+
+func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
+	// Default value 16 for shared-CPU VM
+	volumeLimits := VolumeLimit16
+
+	// Machine-type format: n1-type-NCPU
+	machineType := ns.MetadataService.GetMachineType()
+	if strings.HasPrefix(machineType, "n1-") {
```
msau42: custom machine types are also supported now and have the [syntax](https://cloud.google.com/sdk/gcloud/reference/container/clusters/create):

custom-cpus-memory

where memory is in MiB.

pkg/gce-pd-csi-driver/node.go
msau42: Actually, the limits are determined more by memory now instead of CPU.

It seems like 3.75 GB is the cutoff for whether or not it's 16 or 128.

pkg/gce-pd-csi-driver/node.go
Mockery-Li: Updated the rules

Mockery-Li: I should have checked the original website, instead of the translated one..... It's official but obsolete

pkg/gce-pd-csi-driver/node.go
```diff
@@ -44,15 +44,12 @@ type GCENodeServer struct {
...
 
-// The constants are used to map from the machine type (number of CPUs) to the limit of
+// The constants are used to map from the machine type (Memory) to the limit of
 // persistent disks that can be attached to an instance. Please refer to gcloud doc
-// https://cloud.google.com/compute/docs/disks/#increased_persistent_disk_limits
+// https://cloud.google.com/compute/docs/disks/#pdnumberlimits
 const (
-	OneCPU               = 1
-	EightCPUs            = 8
+	MemCutoff            = 3840		// 3.75GB
```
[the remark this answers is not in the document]
msau42: In GCP, GB == GiB, so this should be 3750

pkg/gce-pd-csi-driver/node.go
```diff
@@ -320,25 +317,25 @@ func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpa
 }
 
 func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
```
msau42: This may be better as a new file, maybe "gce-util.go" under https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/tree/master/pkg/gce-cloud-provider/compute.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -320,25 +317,25 @@ func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpa
...
-	// Default value 16 for shared-CPU VM
+	// Default value 16 for memory lower than cutoff
 	volumeLimits := VolumeLimit16
 
-	// Machine-type format: n1-type-NCPU
+	// Machine-type format: n1-type-CPUS or custom-CPUS-RAM
 	machineType := ns.MetadataService.GetMachineType()
 	if strings.HasPrefix(machineType, "n1-") {
+		volumeLimits = VolumeLimit128
+	} else if machineType == "f1-micro" || machineType == "g1-small" {
```
msau42: Can you add unit tests for all of these?

pkg/gce-pd-csi-driver/node.go
Mockery-Li: Move the function `GetVolumeLimits()` to a new file under the package of gcecloudprovider? Then do we need to pass the `MetadataService` as a parameter?

msau42: Oh I see.  Ok it is fine where it is

pkg/gce-pd-csi-driver/node.go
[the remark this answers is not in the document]
Mockery-Li: Yes, it should be GiB and MiB. I tried to create a custom machine, it was *1024.

pkg/gce-pd-csi-driver/node.go
Mockery-Li: OK, I will have a try

Mockery-Li: For the unit test, I add a function to set machine type for fake metaserver. I'm not sure if it's an appropriate way to do so.

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -13,3 +13,76 @@ limitations under the License.
...
+			machineType:    "g1-small",
+			expVolumeLimit: volumeLimit16,
+		},
+		{
+			name:           "Custom machine with 1GiB Mem",
+			machineType:    "custom-1-1024",
+			expVolumeLimit: volumeLimit16,
+		},
+		{
+			name:           "Custom machine with 4GiB Mem",
```
msau42: Can you also add tests right at the 3.75GiB cutoff?

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -13,3 +13,76 @@ limitations under the License.
...
+func getTestGCEDriver(t *testing.T) *GCEDriver {
+	gceDriver := GetGCEDriver()
+	err := gceDriver.SetupGCEDriver(nil, mountmanager.NewFakeSafeMounter(), mountmanager.NewFakeDeviceUtils(), metadataservice.NewFakeService(), driver, "test-vendor")
+	if err != nil {
+		t.Fatalf("Failed to setup GCE Driver: %v", err)
+	}
+	return gceDriver
+}
+
+func TestNodeGetInfo(t *testing.T) {
```
msau42: Maybe call it "TestNodeGetVolumeLimits" to be more specific

pkg/gce-pd-csi-driver/node.go
```diff
@@ -302,3 +317,35 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
+		// Custom machine types
+		splits := strings.Split(machineType, "-")
+		if len(splits) != customTypeParts {
+			return volumeLimits, fmt.Errorf("Unexpected machine type: %s", machineType)
+		}
+		memStr := splits[customTypeMem]
+		mem, err := strconv.Atoi(memStr)
+
+		if (err == nil) && (mem >= memCutOff) {
+			volumeLimits = volumeLimit128
```
msau42: Ok I got some extra clarification here.  Actually all custom machine types are 128.  There's no memory cutoff.

Mockery-Li: ok, rules and test updated

pkg/gce-pd-csi-driver/node.go
```diff
@@ -322,28 +318,17 @@ func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
...
-		mem, err := strconv.Atoi(memStr)
-
-		if (err == nil) && (mem >= memCutOff) {
-			volumeLimits = volumeLimit128
-		} else {
-			return volumeLimits, err
-		}
+		volumeLimits = volumeLimit128
 	} else {
 		return volumeLimits, fmt.Errorf("Unexpected machine type: %s", machineType)
```
msau42: Maybe this can be simplified to a case statement.  Where if it's custom or n1, then it's 128.  Everything else is 16.  I think we don't need to error for unexpected machine type. That will hopefully make this more resilient to any new machine types added in the future.

Mockery-Li: Yes, it makes sense

msau42:
This is looking great!  Thanks for all your work!  Can you squash your commits, and also add a test case here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/e2e/tests/single_zone_e2e_test.go.  This test suite runs as part of pull-gcp-compute-persistent-disk-csi-driver-e2e

Mockery-Li:
Test case is added to e2e test. I used the forced-push after rebasing, is it the usual way to do so?

msau42:
I don't know why, but for some reason go test is not picking up the new test case :-/  I'm trying out the change locally and am having no luck

msau42:
I tested out the driver manually on a cluster, and I see that the NodeGetInfo is returning the correct response:

```
I0402 22:36:59.949024       1 utils.go:52] GRPC call: /csi.v1.Node/NodeGetInfo
I0402 22:36:59.949048       1 utils.go:53] GRPC request: {}
I0402 22:36:59.949724       1 node.go:288] NodeGetInfo called with req: &csi.NodeGetInfoRequest{XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
I0402 22:36:59.949774       1 utils.go:58] GRPC response: {"accessible_topology":{"segments":{"topology.gke.io/zone":"us-central1-b"}},"max_volumes_per_node":128,"node_id":"projects/msau-k8s-dev/zones/us-central1-b/instances/e2e-test-msau-minion-group-m9td"}
```

And the Kubernetes Node object shows the correct entry:
```
    allocatable:
      attachable-volumes-csi-pd.csi.storage.gke.io: "128"
      attachable-volumes-gce-pd: "64"
      cpu: "2"
      ephemeral-storage: "91117161526"
      hugepages-2Mi: "0"
      memory: 7408816Ki
      pods: "110"
```

msau42:
/lgtm
/approve
I don't know why the new e2e test is not getting run, but we can debug that later.

PR #242, opened 2019-04-04
Upgrade kustomize and use the set image command to patch the image name and version

msau42:


deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -1,5 +1,22 @@
...
-- controller_images.yaml
-- node_images.yaml
+- controller_add_snapshotter.yaml
+- controller_always_pull.yaml
+- node_always_pull.yaml
+images:
+- name: REPLACEME/gcp-compute-persistent-disk-csi-driver
+  # Replace this with your private image name
+  newName: gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver
+  newTag: "v0.4.0-gke.0"
```
verult: Should this be "latest"?

deploy/kubernetes/base/controller.yaml
```diff
@@ -51,3 +51,4 @@ spec:
         - name: cloud-sa-volume
           secret:
             secretName: cloud-sa
+  volumeClaimTemplates: []
```
verult: Why is this needed?

deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -1,5 +1,22 @@
...
+- name: REPLACEME/csi-provisioner
+  newName: quay.io/k8scsi/csi-provisioner
```
verult: Why do we use quay here but GCR elsewhere?

deploy/kubernetes/overlays/dev/kustomization.yaml
msau42: We don't release any official "latest" tags, so I just put our latest supported image.  What I expect developers to do, is to replace both newName and newTag with their own repo and tags.

deploy/kubernetes/base/controller.yaml
msau42: It's a bug in kustomize, will add a comment: https://github.com/kubernetes-sigs/kustomize/issues/504

deploy/kubernetes/overlays/dev/kustomization.yaml
msau42: We have a bit of a chicken and egg problem.  Currently in order to test a new deployment, we need to modify the staging deployment here, check it in, and then look at the staging test results.  

Ideally, I want to be able to validate a new deployment in a pull job first before needing to check it in.  So I want to use the dev overlay for that, but the pull jobs cannot access gke-staging images, so we need to use public images.

msau42: Maybe I'll rename this template to pull instead of dev to make it more obvious what it's purpose is for

deploy/kubernetes/overlays/prow-pull/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+- ../../base
+images:
+- name: REPLACEME/gcp-compute-persistent-disk-csi-driver
+  newName: gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver
+  newTag: "v0.4.0-gke.0"
+- name: REPLACEME/csi-provisioner
+  newName: quay.io/k8scsi/csi-provisioner
```
davidz627: the pull job should be using the gke release versions of sidecars right? and the `REPLACEME/gcp-compute-persistent-disk-csi-driver` should actually be replaced by a pull request specific built container?

deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -1,5 +1,22 @@
...
-- controller_images.yaml
-- node_images.yaml
+- controller_add_snapshotter.yaml
+- controller_always_pull.yaml
+- node_always_pull.yaml
+images:
+- name: REPLACEME/gcp-compute-persistent-disk-csi-driver
+  # Replace this with your private image names and tags
+  newName: gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver
+  newTag: "v0.4.0-gke.0"
```
davidz627: should this instead be not replaced instead. Therefore when someone tries to use a dev overlay without changing anything they get an image pull failure (saying `REPLACEME` and reminding the user to replace to image) instead of failing silently by just giving them a working image.

deploy/kubernetes/overlays/prow-pull/kustomization.yaml
msau42: Ideally the pull job should use the gke versions of sidecars, however, we have a chicken and egg problem when we want to test a new version of a sidecar but it's not available in gke-release yet.  For example, [enabling topology](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/243).

The other option, is to not test new sidecar features in the pull job (only test stable features), and only switch over pull jobs when we change stable.  Actually, the pull job could just use the stable overlay.

deploy/kubernetes/overlays/dev/kustomization.yaml
msau42: We currently use the dev overlay as our alpha features overlay too (for snapshot).  I can create a separate overlay for alpha if we want to separate that out

deploy/kubernetes/overlays/alpha/rbac_add_snapshots_to_provisioner.yaml
```diff
@@ -0,0 +1,16 @@
...
+# using jsonpatch
+# https://github.com/kubernetes-sigs/kustomize/blob/master/examples/jsonpatch.md
+- op: add
+  path: /rules/-
+  value:
+    apiGroups: ["snapshot.storage.k8s.io"]
+    resources: ["volumesnapshots"]
+    verbs: ["get", "list"]
+
+- op: add
```
davidz627: these have to be two seperate rules? We can't consolidate them somehow?

deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -1,5 +1,22 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 bases:
 - ../../base
 patches:
```
davidz627: shouldn't the dev overlay also have the `rbac_add_snapshotter.yaml`

deploy/kubernetes/overlays/prow-gke-head/kustomization.yaml
```diff
@@ -1,5 +1,17 @@
...
+kind: Kustomization
 bases:
 - ../../base
-patches:
-- controller_images.yaml
-- node_images.yaml
+images:
+- name: REPLACEME/gcp-compute-persistent-disk-csi-driver
+  newName: gcr.io/gke-release-staging/gcp-compute-persistent-disk-csi-driver
+  newTag: "latest"
```
davidz627: would it make more sense to rename this overlay `prow-gke-release-staging` or something like that?

davidz627: wait nevermind, thats another overlay. `prow-gke-release-staging-latest`?

deploy/kubernetes/overlays/alpha/rbac_add_snapshots_to_provisioner.yaml
msau42: Based on the syntax, I don't think it's possible to combine them.  The `value` field is not an array

deploy/kubernetes/overlays/dev/kustomization.yaml
msau42: Rebased dev overlay on top of alpha overlay

deploy/kubernetes/overlays/prow-gke-head/kustomization.yaml
msau42: renamed all prow-gke overlays

PR #243, opened 2019-04-04
Enable topology

msau42:
Depends on https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/242

/hold


davidz627:
@msau42 is this ready for review?

msau42:
Yes this is ready.

I have a hold here so we can decide if we want to cut a new driver before adding topology since it will require min K8s 1.14

msau42:
We can always backport things to older releases if we want something for < 1.14

msau42:
We should cut this as 0.5.x since it requires K8s version bump

deploy/kubernetes/overlays/stable/kustomization.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 bases:
 - ../../base
+patches:
+- controller_disable_topology.yaml
```
davidz627: I would prefer that all patches are additive to minimize confusion.
The way I think dependencies should work:
```
         -> stable
base -> alpha (add topology here) -> dev
                                                       -> staging
```
Let me know what you think

msau42: sure that works

msau42: pushed a new commit

PR #247, opened 2019-04-19
Fix stable image name

msau42:
I have no idea how pull jobs are passing


davidz627:
@msau42 pull jobs probably working because they use stable overlay but overwrite the driver image with the built image from the PR commit. (so a wacky driver image gets overwritten I think)

/lgtm

PR #248, opened 2019-04-19
Add unit test for Node

Mockery-Li:
Fixes issue #48 

The test cases cover valid and invalid arguments, but don't know how to test internal errors.

### DIFF
--- a/pkg/gce-pd-csi-driver/node_test.go
+++ b/pkg/gce-pd-csi-driver/node_test.go
@@ -25,6 +25,10 @@ import (
 	mountmanager "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/mount-manager"
 )
 
+const defaultVolumeID = "project/test001/zones/c1/disks/testDisk"
+const defaultTargetPath = "/mnt/test"
+const defaultStagingPath = "/staging"
+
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -86,3 +88,270 @@ func TestNodeGetVolumeLimits(t *testing.T) {
...
+	ns := gceDriver.ns
+	testCases := []struct {
+		name       string
+		req        *csi.NodePublishVolumeRequest
+		expErrCode codes.Code
+	}{
+		{
+			name: "Valid request",
+			req: &csi.NodePublishVolumeRequest{
+				VolumeId:          "1",
```
davidz627: can we use a valid volume ID here of the form `projects/{project}/zones/{zone}/disks/{name}`
technically I think we want to validate the ID format and return an invalidargument error if it is not in that format (although right now we don't do that)

davidz627: ditto for the other tests as well

Mockery-Li: OK. I created several const variables for default testing volume ID, staging path and target path.

PR #254, opened 2019-04-26
Statically validate volume capabilities for CreateVolume, ControllerPublish, NodeStage, NodePublish

davidz627:
Fixes: #94

/assign @msau42 @saad-ali 

```release-note
BREAKING CHANGE: The driver now enforces AccessMode validation for all calls, supported access modes are SINGLE_NODE_WRITER, SINGLE_NODE_READER_ONLY, MULTI_NODE_READER_ONLY
```

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -17,6 +17,9 @@ limitations under the License.
 package gceGCEDriver
 
 import (
+	"errors"
+	"fmt"
+
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/glog"
 	"github.com/kubernetes-csi/csi-lib-utils/protosanitizer"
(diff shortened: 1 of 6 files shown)


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -59,3 +62,47 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
...
+		}
+	}
+	return nil
+}
+
+func validateVolumeCapability(vc *csi.VolumeCapability) error {
+	if err := validateAccessModes(vc.GetAccessMode()); err != nil {
+		return err
+	}
+	if blk := vc.GetBlock(); blk != nil {
```
msau42: can both block and mount be set at the same time?

davidz627: I don't believe so: https://github.com/container-storage-interface/spec/blame/master/spec.md#L855

PR #258, opened 2019-04-30
Add test config for testing CSI Migration with PD Driver

davidz627:
Depends on: https://github.com/kubernetes/kubernetes/pull/77101

Adds a test config and local test config for CSI Migration testing. Brings up a cluster with migration flags, gce provider and zone, then runs the test with the `storage.migratedPlugins` flag. This will run all in-tree gcepd driver tests with the Migrated driver.

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -31,20 +31,29 @@ import (
 )
 
 var (
-	teardownCluster    = flag.Bool("teardown-cluster", true, "teardown the cluster after the e2e test")
-	teardownDriver     = flag.Bool("teardown-driver", true, "teardown the driver after the e2e test")
-	bringupCluster     = flag.Bool("bringup-cluster", true, "build kubernetes and bringup a cluster")
-	stagingImage       = flag.String("staging-image", "", "name of image to stage to")
-	kubeVersion        = flag.String("kube-version", "master", "version of Kubernetes to download and use")
-	inProw             = flag.Bool("run-in-prow", false, "is the test running in PROW")
-	saFile             = flag.String("service-account-file", "", "path of service account file")
-	deployOverlayName  = flag.String("deploy-overlay-name", "", "which kustomize overlay to deploy the driver with")
-	localK8sDir        = flag.String("local-k8s-dir", "", "local kubernetes/kubernetes directory to run e2e tests from")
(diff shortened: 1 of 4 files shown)


test/k8s-integration/main.go
```diff
@@ -231,13 +248,24 @@ func setEnvProject(project string) error {
 	return nil
 }
 
-func runTests(pkgDir, k8sDir, storageClassFile, testFocus string) error {
-	testDriverConfigFile, err := generateDriverConfigFile(pkgDir, storageClassFile)
-	if err != nil {
-		return err
+func runTests(pkgDir, k8sDir, storageClassFile, testFocus, gceZone string, migrationTest bool) error {
```
msau42: since these two tests are mutually exclusive, should we have two functions? One for setting the appropriate arguments for each?  And they can funnel down to the common method.

davidz627:
@msau42 resolved your comment, I wasn't sure if this is exactly what you meant. Let me know if it isn't!

test/k8s-integration/main.go
```diff
@@ -231,13 +248,30 @@ func setEnvProject(project string) error {
 	return nil
 }
 
-func runTests(pkgDir, k8sDir, storageClassFile, testFocus string) error {
-	testDriverConfigFile, err := generateDriverConfigFile(pkgDir, storageClassFile)
-	if err != nil {
-		return err
+func runTests(pkgDir, k8sDir, testFocus, storageClassFile, gceZone string, migrationTest bool) error {
```
msau42: I was thinking more like having methods like "runMigrationTest" and "runCSITests" so it's very clear that we have two completely different test suites going on.

davidz627: done

davidz627:
/hold
integration tests aren't actually running any tests anymore

davidz627:
/hold cancel

@msau42 ready for review again

test/k8s-integration/config/test-config.yaml
```diff
@@ -0,0 +1,21 @@
+ShortName: pdtest
```
msau42: this file should not be checked in.  maybe we should add to .gitignore?

davidz627: @msau42 resolved :+1:

PR #260, opened 2019-04-30
[Cherry-pick] Make integration test kube version configurable

davidz627:
So that we can run a k8s 1.13 test with 0.4 version of the driver

/assign @msau42


davidz627:
@msau42 there was a typo where I added an `x` in a random place so the tests were failing. Please re-LGTM

PR #269, opened 2019-05-10
Tolerate legacy 'fstype' parameter

davidz627:
Migration scenario storage classes may have "fstype" parameter set that gets passed directly through external provisioner to the driver (while also translating it). We need to tolerate this on the driver side.

/assign @msau42

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -106,6 +106,14 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		case common.ParameterKeyDiskEncryptionKmsKey:
 			// Resource names (e.g. "keyRings", "cryptoKeys", etc.) are case sensitive, so do not change case
 			diskEncryptionKmsKey = v
+		case common.ParameterKeyFsType:
+			// We have to support seeing the "fstype" parameter to support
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -106,6 +106,14 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		case common.ParameterKeyDiskEncryptionKmsKey:
 			// Resource names (e.g. "keyRings", "cryptoKeys", etc.) are case sensitive, so do not change case
 			diskEncryptionKmsKey = v
+		case common.ParameterKeyFsType:
+			// We have to support seeing the "fstype" parameter to support
+			// migrated volumes with legacy storage classes that contain
+			// "fstype". "fstype" through volume parameters will not be
```
msau42: Should we be translating the storageclass "fstype" parameter to our prefixed one?

msau42: (just like we should be translating our zone/zones storageclass parameters?)

davidz627: oh you're right. The translation happens before this

davidz627:
Replaced by: https://github.com/kubernetes/kubernetes/pull/77837

PR #271, opened 2019-05-13
Add kubernetes failure domain as a valid topology key for zone

davidz627:
/assign @msau42 

In-tree storage classes use Kubernetes failure domain as zone topology keys.

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -23,7 +23,8 @@ const (
 	ParameterKeyDiskEncryptionKmsKey = "disk-encryption-kms-key"
 
 	// Keys for Topology. This key will be shared amongst drivers from GCP
-	TopologyKeyZone = "topology.gke.io/zone"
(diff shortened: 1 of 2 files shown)


pkg/common/constants.go
```diff
@@ -23,7 +23,8 @@ const (
 	ParameterKeyDiskEncryptionKmsKey = "disk-encryption-kms-key"
 
 	// Keys for Topology. This key will be shared amongst drivers from GCP
-	TopologyKeyZone = "topology.gke.io/zone"
+	TopologyKeyZone           = "topology.gke.io/zone"
+	KubernetesTopologyKeyZone = "failure-domain.beta.kubernetes.io/zone"
```
msau42: you don't want to translate this one?

davidz627: You're right, should be possible after this change: https://github.com/kubernetes/kubernetes/pull/77837

PR #273, opened 2019-05-16
Bump Provisioner in staging to v1.2.0-rc1

davidz627:
/assign @msau42


davidz627:
We should invest in a more automated release process when things settle down. This is annoying

PR #274, opened 2019-05-16
Integration with GCE cloud config

verult:
This PR introduces parsing logic for GCE cloud config file. In particular, it extracts information for:
* An AltTokenSource for authenticating with GCP
* ID of the project in which the driver runs.

In adding token source integration, I used the following as reference:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/gce/gce.go
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/gce/token_source.go

/assign @davidz627 @mikedanese

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -15,7 +15,10 @@ limitations under the License.
 package gcecloudprovider
 
 import (
+	"context"
 	"fmt"
+	"golang.org/x/oauth2/google"
+	"gopkg.in/gcfg.v1"
 	"net/http"
 	"os"
 	"runtime"
@@ -24,9 +27,8 @@ import (
 	"cloud.google.com/go/compute/metadata"
 	"github.com/golang/glog"
 	"golang.org/x/oauth2"
-	"golang.org/x/oauth2/google"
 	beta "google.golang.org/api/compute/v0.beta"
-	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
@@ -57,20 +59,44 @@ type CloudProvider struct {
 
 var _ GCECompute = &CloudProvider{}
 
-func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
-	svc, err := createCloudService(vendorVersion)
+type ConfigFile struct {
+	Global ConfigGlobal `gcfg:"global"`
+}
+
+type ConfigGlobal struct {
+	TokenURL  string `gcfg:"token-url"`
+	TokenBody string `gcfg:"token-body"`
+	ProjectId string `gcfg:"project-id"`
+}
+
+func CreateCloudProvider(vendorVersion string, configPath string) (*CloudProvider, error) {
+	configFile, err := readConfig(configPath)
+	if err != nil {
+		return nil, err
+	}
+	// At this point configFile could still be nil.
+	// Any following code that uses configFile should handle nil pointer gracefully.
+
+	glog.V(1).Infof("Using GCE provider config %+v", configFile)
+
+	tokenSource, err := generateTokenSource(configFile)
 	if err != nil {
 		return nil, err
 	}
 
-	betasvc, err := createBetaCloudService(vendorVersion)
+	svc, err := createCloudService(vendorVersion, tokenSource)
 	if err != nil {
 		return nil, err
 	}
 
-	project, zone, err := getProjectAndZoneFromMetadata()
+	betasvc, err := createBetaCloudService(vendorVersion, tokenSource)
 	if err != nil {
-		return nil, fmt.Errorf("Failed getting Project and Zone from Metadata server: %v", err)
+		return nil, err
+	}
+
+	project, zone, err := getProjectAndZone(configFile)
+	if err != nil {
+		return nil, fmt.Errorf("Failed getting Project and Zone: %v", err)
 	}
 
 	return &CloudProvider{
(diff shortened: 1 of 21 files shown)


pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -83,8 +111,33 @@ func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
...
 }
 
-func createBetaCloudService(vendorVersion string) (*beta.Service, error) {
-	client, err := newDefaultOauthClient()
+func generateTokenSource(configFile *ConfigFile) (tokenSource oauth2.TokenSource, err error) {
+	if configFile != nil {
+		if configFile.Global.TokenURL != "" {
+			// if tokenURL is nil, set tokenSource to nil. This will force the OAuth client to fall
+			// back to use DefaultTokenSource. This allows running gceCloud remotely.
+			if configFile.Global.TokenURL == "nil" {
```
mikedanese: yikes, do you actually see string "nil" sometimes?

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -83,8 +111,33 @@ func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
 
 }
 
-func createBetaCloudService(vendorVersion string) (*beta.Service, error) {
-	client, err := newDefaultOauthClient()
+func generateTokenSource(configFile *ConfigFile) (tokenSource oauth2.TokenSource, err error) {
```
mikedanese: err is unused now.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -83,8 +111,33 @@ func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
...
+	if configFile != nil {
```
mikedanese:
```
if configFile != nil && configFile.Global.TokenURL != "" && configFile.Global.TokenURL != "nil" {
  return NewAltTokenSource(configFile.Global.TokenURL, configFile.Global.TokenBody)
}
return google.DefaultTokenSource()
```

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -83,8 +111,33 @@ func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
...
+			} else {
+				tokenSource = NewAltTokenSource(configFile.Global.TokenURL, configFile.Global.TokenBody)
+			}
+		}
+	}
+
+	return
+}
+
+func readConfig(reader io.Reader) (*ConfigFile, error) {
```
mikedanese: maybe accept a file path and do the file handling in here?

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -114,20 +167,25 @@ func createCloudServiceWithDefaultServiceAccount(vendorVersion string) (*compute
...
-func newDefaultOauthClient() (*http.Client, error) {
-	// No compute token source, fallback on default
-	tokenSource, err := google.DefaultTokenSource(
-		oauth2.NoContext,
-		compute.CloudPlatformScope,
-		compute.ComputeScope)
-	if gac, ok := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); ok {
-		glog.V(4).Infof("GOOGLE_APPLICATION_CREDENTIALS env var set %v", gac)
+func newOauthClient(tokenSource oauth2.TokenSource) (*http.Client, error) {
+	if tokenSource == nil {
```
mikedanese: handle this in generateTokenSource

pkg/gce-cloud-provider/compute/token_source.go
```diff
@@ -0,0 +1,91 @@
+/*
```
mikedanese: This seems fine for now but it would be nice to get this into a place where the code can be shared. I realize the one in cloud provider carries too many dependencies. We did the same thing here:

https://github.com/kubernetes/cloud-provider-gcp/blob/master/cmd/gke-exec-auth-plugin/alt_token_source.go

mikedanese:
LGTM

pkg/gce-cloud-provider/compute/token_source.go
verult: Sounds good, if there's a tracking issue for a refactor somewhere I'm happy to link to it

pkg/gce-cloud-provider/compute/gce.go
verult: I haven't seen it myself. This was copied from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/gce/gce.go#L293

What would you recommend?

verult:
Addressed all comments except the one about string == "nil"

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -83,8 +106,49 @@ func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
...
+		tokenSource := NewAltTokenSource(configFile.Global.TokenURL, configFile.Global.TokenBody)
+		glog.V(4).Infof("Using AltTokenSource %#v", tokenSource)
+		return tokenSource, nil
+	}
+
+	tokenSource, err := google.DefaultTokenSource(
+		context.Background(),
+		compute.CloudPlatformScope,
+		compute.ComputeScope)
+	if gac, ok := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); ok {
```
davidz627: I know you didn't write this code but I got confused when reading it again, could you put a comment saying that `DefaultTokenSource` tries to use `GOOGLE_APPLICATION_CREDENTIALS` and thats why we spit out a warning if its not set.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -57,13 +59,34 @@ type CloudProvider struct {
...
+func CreateCloudProvider(vendorVersion string, configPath string) (*CloudProvider, error) {
+	configFile, err := readConfig(configPath)
 	if err != nil {
+		// Error is returned only when configPath is non-empty.
 		return nil, err
 	}
+	glog.Infof("Using GCE provider config %+v", configFile)
 
-	betasvc, err := createBetaCloudService(vendorVersion)
+	tokenSource, err := generateTokenSource(configFile)
```
davidz627: can we do instead `if configFile == nil` then `generateAltToknSource` 
`else` `generateDefaultTokenSource` (or something like thati) instead of doing completely orthogonal logic depending on whether the input is set or not.

pkg/gce-cloud-provider/compute/token_source.go
davidz627: If theres no existing issue could you create an issue and link to it, thanks!

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -66,6 +66,7 @@ type ConfigFile struct {
 type ConfigGlobal struct {
```
davidz627: are these structs defined elsehwere? They feel like structs that should be defined in the token library or something

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -192,15 +193,24 @@ func newOauthClient(tokenSource oauth2.TokenSource) (*http.Client, error) {
...
 	zone, err := metadata.Zone()
 	if err != nil {
 		return "", "", err
 	}
-	projectID, err := metadata.ProjectID()
-	if err != nil {
-		return "", "", err
+
+	var projectID string
+	if config == nil || config.Global.ProjectId == "" {
```
davidz627: should we spit out a warning when this happens? Is this expected behavior? 

If `config != nil` I would sort of expect the project ID being empty to be an error or something. If thats not a case could you write a comment explaining how we could get into a situation like that?

It just feels like a case where we could potentially want to use the `config`'s project ID and we'd have a silent failure here if that ever had any problems.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -83,8 +106,49 @@ func CreateCloudProvider(vendorVersion string) (*CloudProvider, error) {
 
 }
 
-func createBetaCloudService(vendorVersion string) (*beta.Service, error) {
-	client, err := newDefaultOauthClient()
+func generateTokenSource(configFile *ConfigFile) (oauth2.TokenSource, error) {
+	// Note: if tokenURL is nil, set tokenSource to DefaultTokenSource.
```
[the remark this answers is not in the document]
verult: Sorry, copy paste artifact. Removing these comments since they don't add much value

pkg/gce-cloud-provider/compute/gce.go
verult: Yeah good call, I had to dig in a little deeper to answer this question too

pkg/gce-cloud-provider/compute/gce.go
verult: I'm not sure why that'd be better - isn't this just splitting up the branches one function call level above what's already written?

pkg/gce-cloud-provider/compute/gce.go
verult: These structs follow the GCE configuration file format. It's defined in kubernetes/legacy-cloud-providers but I don't think it's worth adding a whole extra dependency for just a few fields we need in the conf file. This isn't limited to tokens; project ID comes from this conf too.

I do wonder why this format isn't in kubernetes/cloud-provider-gcp and whether this is gonna be a problem. Looking into it

pkg/gce-cloud-provider/compute/gce.go
verult: ACK on the warning. I'll also add a comment to explain things

pkg/gce-cloud-provider/compute/gce.go
davidz627: right. the logic ends up being the same but I think its more maintainable.

Instead of having one function do two completely different things based on an input parameter. It makes sense to split those "two different things" into two different functions.

verult: I have a slight preference for this version actually, since someone reading this function (`CreateCloudProvider()`) could just skip over the details of how the token source is instantiated if they don't care.

To also have the benefit you mentioned, I could add in a comment in each of the branches to make it explicit they are doing two very different things

davidz627:
LGTM please squash commits

PR #279, opened 2019-05-31
Release v0.5.0 version as stable

davidz627:
newest release version been cut, update deployment specs and docs.

/assign @msau42 @verult

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,6 @@
 STAGINGIMAGE=${GCE_PD_CSI_STAGING_IMAGE}
 STAGINGVERSION=${GCE_PD_CSI_STAGING_VERSION}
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -18,9 +18,6 @@
 STAGINGIMAGE=${GCE_PD_CSI_STAGING_IMAGE}
 STAGINGVERSION=${GCE_PD_CSI_STAGING_VERSION}
 
-PRODIMAGE=gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver
```
verult: Why were these lines removed?

davidz627: we aren't building a prod image from the makefile ever (we use the internal release flows). I wanted to keep them around for "documentation" purposes for a while, but now our actual documentation captures this same information so I think its reasonable to remove

davidz627:
cmon prow you can do this

PR #280, opened 2019-05-31
Updating image paths to gke.gcr.io

verult:
gcr.io/gke-releases is only available in the US, while gke.gcr.io is available in multiple regions.

/assign @davidz627 
/cc @msau42


davidz627:
Cool! Thanks for finding this

/lgtm
/approve

PR #282, opened 2019-06-04
Update scripts to be compatible with project ids that include the org…

hantaowang:
Some project ids are of the format `<org-name>:<project>`. For this case, the gCloud IAM ServiceAccount create function creates a service account email of the format `<service-account-name>@<project>.<org-name>.iam.serviceaccount.com`, which is incompatible with how our script expects the email. This PR adds a special cases that provides to correct formatting.


hantaowang:
I signed it

hantaowang:
I signed it

davidz627:
/lgtm
/approve
thanks @hantaowang!

Issue #281, opened 2019-06-03
External attacher RBAC out of date


verult:
It still references csinodeinfos, and it's missing some permissions. See: https://github.com/kubernetes-csi/external-attacher/blob/master/deploy/kubernetes/rbac.yaml#L19

verult:
/assign

* external-attacher: csinodeinfos should now be csinodes
* external-attacher: volumeattachments needs patch permissions
* external-provisioner: missing node permissions
* external-snapshotter rules shouldn't exist in stable version

PR #284, opened 2019-06-05
Update RBAC rules for external provisioner and attacher to the updated roles required based on upstream repositories

davidz627:
Resolves: #281 

Should fix the "latest" test failures.

/assign @msau42 @verult


verult:
/lgtm

In the future it would be nice to include snapshotter RBAC rules only for the alpha deployment, if possible

davidz627:
moved snapshot rules to alpha overlay

deploy/kubernetes/overlays/alpha/rbac_add_snapshotter.yaml
```diff
@@ -4,27 +4,33 @@ kind: ClusterRole
...
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["secrets"]
```
verult: Secrets were originally included in attacher roles and then later removed. Do we need the secrets permissions here for different reasons?

@jingxu97

msau42: I don't think so. Secrets are only required if the plugin needs them, and pd doesn't.

davidz627: resolved

PR #288, opened 2019-06-06
add ability to use gcloud to boot a gke cluster for e2e integration

hantaowang:
Adds the following flags to `k8s-integration-test`:
- `deployment-strategy`: If a cluster needs to be booted up, choose if it is done on GCE through the `kube-ip.sh` script or GKE through the `gcloud` utility. The GKE option is incompatible with the migration test and must use a cluster version supported by GKE. Feature gates cannot be turned on when GKE is used. The default value is "gce", and GKE can be turned on by setting it to "gke".
- `gke-cluster-version`: If GKE is used, this flag can be set to an available cluster version, which can be checked with `gcloud container get-server-config`. Also accepted values are "latest" and "stable". The deault value is "latest".

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -39,6 +39,8 @@ var (
 	kubeVersion      = flag.String("kube-version", "master", "version of Kubernetes to download and use")
 	kubeFeatureGates = flag.String("kube-feature-gates", "", "feature gates to set on new kubernetes cluster")
 	localK8sDir      = flag.String("local-k8s-dir", "", "local kubernetes/kubernetes directory to run e2e tests from")
+	deploymentStrat  = flag.String("deployment-strategy", "gce", "choose between deploying on gce or gke")
+	gkeClusterVer    = flag.String("gke-cluster-version", "latest", "version of Kubernetes master and node for gke")
 
 	// Test infrastructure flags
 	boskosResourceType = flag.String("boskos-resource-type", "gce-project", "name of the boskos resource type to reserve")
@@ -59,6 +61,7 @@ var (
 const (
 	pdImagePlaceholder = "gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver"
 	k8sBuildBinDir     = "_output/dockerized/bin/linux/amd64"
+	gkeTestClusterName = "gcp-pd-csi-driver-test-cluster"
 )
 
 func init() {
@@ -97,6 +100,10 @@ func main() {
 
 	if len(*gceZone) == 0 {
 		glog.Fatalf("gce-zone is a required flag")
+	}
+
+	if *deploymentStrat == "gke" && *migrationTest {
+		glog.Fatalf("Cannot set deployment strategy to 'gke' for migration tests.")
 	}
 
 	err := handle()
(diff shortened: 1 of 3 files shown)


test/k8s-integration/main.go
```diff
@@ -39,6 +39,8 @@ var (
 	kubeVersion      = flag.String("kube-version", "master", "version of Kubernetes to download and use")
 	kubeFeatureGates = flag.String("kube-feature-gates", "", "feature gates to set on new kubernetes cluster")
 	localK8sDir      = flag.String("local-k8s-dir", "", "local kubernetes/kubernetes directory to run e2e tests from")
+	deploymentStrat  = flag.String("deployment-strategy", "gce", "choose between deploying on gce or gke")
+	gkeClusterVer    = flag.String("gke-cluster-version", "latest", "version of Kubernetes master and node for gke")
```
msau42: Can we reuse kube-version?

test/k8s-integration/main.go
```diff
@@ -59,6 +61,7 @@ var (
 const (
 	pdImagePlaceholder = "gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver"
 	k8sBuildBinDir     = "_output/dockerized/bin/linux/amd64"
+	gkeTestClusterName = "gcp-pd-csi-driver-test-cluster"
```
msau42: Can we append some random suffix to it? That will let us run multiple test instances in parallel.  Or if some previous test cluster failed to cleanup properly, then subsequent test runs would not get blocked.

test/k8s-integration/main.go
```diff
@@ -196,15 +203,30 @@ func handle() error {
 			glog.V(4).Infof("Set Kubernetes feature gates: %v", *kubeFeatureGates)
 		}
 
-		err = clusterUp(k8sDir, *gceZone)
-		if err != nil {
-			return fmt.Errorf("failed to cluster up: %v", err)
+		if *deploymentStrat == "gce" {
```
msau42: this may look a little cleaner as a switch statement

test/run-k8s-integration.sh
```diff
@@ -19,4 +19,5 @@ make -C ${PKGDIR} test-k8s-integration
 ${PKGDIR}/bin/k8s-integration-test --kube-version=${GCE_PD_KUBE_VERSION:-master} \
 --run-in-prow=true --deploy-overlay-name=${overlay_name} --service-account-file=${E2E_GOOGLE_APPLICATION_CREDENTIALS} \
 --do-driver-build=${do_driver_build} --boskos-resource-type=${boskos_resource_type} \
---storageclass-file=sc-standard.yaml --test-focus="External.Storage" --gce-zone="us-central1-b"
+--storageclass-file=sc-standard.yaml --test-focus="External.Storage" --gce-zone="us-central1-b" \
+--deployment-strategy=${DEPLOYMENT_STRATEGY:-gce}
```
msau42: can you default variables at the top similar to the others?

test/k8s-integration/main.go
hantaowang: Its a little hard, because "master" is not an option for GKE. And also the set of possible options for GCE and GKE are different (i.e. a GKE option can be 1.13.6-gke.6). Because there is essentially no overlap in the version names and also the default option, I felt it would be less confusing to use a new flag, but thoughts?

test/k8s-integration/main.go
hantaowang: Unless I'm mistaken, the current kube-up script does not do this either. Are separate tests run through prow supposed to be run on separate projects, thus avoiding any concurrent and subsequent conflicts.

test/k8s-integration/main.go
hantaowang: Also, the kube-version is used to pull the repo to build tests, and repo versioning is not the same as gke versioning.

test/k8s-integration/main.go
msau42: Yes, technically with boskos project lending, there shouldn't be concurrent runs in the same project.  However, I'm thinking more about the scenario where a previous run failed to properly clean up the cluster (for whatever reason).

The kube-up script first tries to delete any existing cluster before trying to create a new one to get around this issue.  We may want to consider doing something similar for gke.

test/k8s-integration/main.go
msau42: Ah I see, I forgot we still need to download kubernetes to get the test suite even for GKE runs.  This makes sense then.

test/k8s-integration/main.go
```diff
@@ -196,15 +203,31 @@ func handle() error {
...
+			return fmt.Errorf("deployment-strategy must be set to 'gce' or 'gke', but is: %s", *deploymentStrat)
 		}
+
 	}
 
 	if *teardownCluster {
 		defer func() {
-			err := clusterDown(k8sDir)
+			var err error
+			if *deploymentStrat == "gke" {
```
davidz627: nit: this can also be better as a switch statement and should contain a default with an error.

It's possible to have `teardownCluster` set without `bringUpCluster` set. So it's not safe to assume `*deploymentStrat != "gke" --> deploymentStrat == gce` in this case.

test/run-k8s-integration-local.sh
```diff
@@ -14,7 +16,8 @@ make -C ${PKGDIR} test-k8s-integration
 # ${PKGDIR}/bin/k8s-integration-test --kube-version=master --run-in-prow=false \
 # --staging-image=${GCE_PD_CSI_STAGING_IMAGE} --service-account-file=${GCE_PD_SA_DIR}/cloud-sa.json \
 # --deploy-overlay-name=dev --storageclass-file=sc-standard.yaml \
-# --test-focus="External.Storage" --gce-zone="us-central1-b"
+# --test-focus="External.Storage" --gce-zone="us-central1-b" \
+# --deployment-strategy=${deployment_strategy}
```
davidz627: add new flag to the other version of the local script as well

hantaowang:
@msau42 @davidz627 the latest commit should address your remaining comments. if there is nothing else, I will squash tomorrow morning for the final approval.

PR #289, opened 2019-06-06
Remove cross validation of access modes, multiple access modes can be specified that represent all the capabilities of the volume

davidz627:
/assign @msau42 @saad-ali 

Needs to be cherry-picked to release-0.5.0 and new release version to be cut.

Also we should identify why none of the tests specify more than one access mode...

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -72,31 +72,6 @@ func validateVolumeCapabilities(vcs []*csi.VolumeCapability) error {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


davidz627:
/hold
for https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/291

msau42:
Should we add an e2e test with multiple access modes?

davidz627:
@msau42 there are now unit tests with multiple access modes. K8s has e2e tests with multiple access modes that are not in the external test suites (yet, I assume)

msau42:
Do you think it is a valuable test case to add to the k8s external test suite?

PR #290, opened 2019-06-06
Cherrypick to v0.5.0: Remove cross validation of access modes, multiple access modes can be specified that represent all the capabilities of the volume

davidz627:
Also just pick up @verult's change updating image paths to gke.gcr.io

/assign @msau42 @verult 
/hold
until I get in the integration test fix


davidz627:
/hold cancel
cherry-pick now also picks up integration test fix

verult:
Commits for updating image paths LGTM

PR #291, opened 2019-06-06
Change pd image placeholder in integration tests to gke.gcr.io vanity url

davidz627:
/assign @msau42 
Also added comments to make sure this doesn't happen again


davidz627:
/hold
for confirmation integration test is run with image built during the test

davidz627:
/hold cancel
confirmed the integration test now uses the image

PR #302, opened 2019-06-14
parallelize node operations

hantaowang:
Improve performance while maintaining idempotentcy by implementing volume granularity locking in node operations.

### DIFF
--- a/pkg/gce-pd-csi-driver/lock_manager_test.go
+++ b/pkg/gce-pd-csi-driver/lock_manager_test.go
@@ -1,0 +1,147 @@
+package gceGCEDriver
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+// Checks that the lock manager has the expected number of locks allocated.
+// this function is implementation dependant! It acquires the lock and directly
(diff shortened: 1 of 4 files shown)


pkg/gce-pd-csi-driver/lock_manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+// checks the map of the lock manager.
+func checkAllocation(lm *LockManager, expectedNumAllocated int, t *testing.T) {
+	lm.mux.Lock()
+	defer lm.mux.Unlock()
+	if len(lm.locks) != expectedNumAllocated {
+		t.Fatalf("expected %d locks allocated, but found %d", expectedNumAllocated, len(lm.locks))
+	}
+}
+
+// coinOperatedMutex is a mutex that only acquires if a "coin" is provided. Otherwise
```
davidz627: this is super cool. Definitely keep this code around and we can reference it for your reviews, though unfortunately sometimes the right thing to do is not the interesting/challenging thing :(

Issue #136, opened 2018-10-23
Allow parallel operations on different volumes


msau42:
We put global locks on mount operations (I didn't check about attach/detach).  This limits the driver from only executing one mount at a time.  We should allow mounts to happen in parallel if they're on different volumes.

hantaowang:
See PR #302 and #303. 

After scale testing on a single node, there is reason to believe that much of the delay comes from not actually NodePublishVolume/NodeStageVolume but actually from the Attaching side. Pod events show constant RPC context DeadlineExceeded errors, Attachment timeout errors, and Internal errors (stemming from an error from googleapi regarding too many pending operations on the same resource). All of this causes a backoff loop for retrying.

I've also observed some timeouts on the kubelet side due to [WaitForAttach](https://github.com/kubernetes/kubernetes/blob/c47ab33e670cd5589b07418457e39b25cea6512f/pkg/volume/util/operationexecutor/operation_generator.go#L631) taking too long, which also cause backoffs.

davidz627:
resolved by #303

PR #303, opened 2019-06-15
return an error if the volume ID already in use

hantaowang:
An alternate approach to #302, but just returns an error if it finds that another operation with that volume ID already exists.

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -36,8 +36,10 @@ type GCENodeServer struct {
 	Mounter         *mount.SafeFormatAndMount
 	DeviceUtils     mountmanager.DeviceUtils
 	MetadataService metadataservice.MetadataService
-	// TODO: Only lock mutually exclusive calls and make locking more fine grained
-	mux sync.Mutex
+
+	// A map storing all volumes with ongoing operations so that additional operations
+	// for that same volume (as defined by VolumeID) return an Aborted error
+	volumes sync.Map
 }
 
 var _ csi.NodeServer = &GCENodeServer{}
@@ -51,8 +53,6 @@ const (
 )
 
 func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
-	ns.mux.Lock()
-	defer ns.mux.Unlock()
 	klog.V(4).Infof("NodePublishVolume called with req: %#v", req)
 
 	// Validate Arguments
@@ -73,6 +73,11 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 	if volumeCapability == nil {
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Volume Capability must be provided")
 	}
+
+	if _, alreadyExists := ns.volumes.LoadOrStore(volumeID, true); alreadyExists {
+		return nil, status.Error(codes.Aborted, fmt.Sprintf("An operation with the given Volume ID %s already exists", volumeID))
+	}
+	defer ns.volumes.Delete(volumeID)
 
 	if err := validateVolumeCapability(volumeCapability); err != nil {
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("VolumeCapability is invalid: %v", err))
@@ -180,9 +185,8 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 }
 
 func (ns *GCENodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpublishVolumeRequest) (*csi.NodeUnpublishVolumeResponse, error) {
-	ns.mux.Lock()
-	defer ns.mux.Unlock()
 	klog.V(4).Infof("NodeUnpublishVolume called with args: %v", req)
+
 	// Validate Arguments
 	targetPath := req.GetTargetPath()
 	volID := req.GetVolumeId()
@@ -193,6 +197,11 @@ func (ns *GCENodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeU
 		return nil, status.Error(codes.InvalidArgument, "NodeUnpublishVolume Target Path must be provided")
 	}
 
+	if _, alreadyExists := ns.volumes.LoadOrStore(volID, true); alreadyExists {
+		return nil, status.Error(codes.Aborted, fmt.Sprintf("An operation with the given Volume ID %s already exists", volID))
+	}
+	defer ns.volumes.Delete(volID)
+
 	err := mount.CleanupMountPoint(targetPath, ns.Mounter.Interface, false /* bind mount */)
 	if err != nil {
 		return nil, status.Error(codes.Internal, fmt.Sprintf("Unmount failed: %v\nUnmounting arguments: %s\n", err, targetPath))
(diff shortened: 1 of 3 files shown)


msau42:
I like this approach. The only enhancement I can think of is hung/long operation detection.  We may want to track operation start time, and log a message/metric if some operation is taking too long.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -38,7 +38,7 @@ type GCENodeServer struct {
 	DeviceUtils     mountmanager.DeviceUtils
 	MetadataService metadataservice.MetadataService
 	// TODO: Only lock mutually exclusive calls and make locking more fine grained
```
davidz627: think this comment can be removed

pkg/gce-pd-csi-driver/node.go
```diff
@@ -194,6 +195,11 @@ func (ns *GCENodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeU
 		return nil, status.Error(codes.InvalidArgument, "NodeUnpublishVolume Target Path must be provided")
 	}
 
+	if _, alreadyExists := ns.volumes.LoadOrStore(volID, true); alreadyExists {
```
davidz627: possible to unit test? Test stores a vol id, calls one of these functions, make sure we get an `AlreadyExists`. Test stores an unrelated vol id, make sure there is no issue calling a function.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -194,6 +195,11 @@ func (ns *GCENodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeU
...
+		return nil, status.Error(codes.AlreadyExists, "An operation with the given Volume ID already exists")
```
davidz627: might be worth returning the volume ID in question as part of the error string

pkg/gce-pd-csi-driver/node.go
```diff
@@ -38,7 +38,7 @@ type GCENodeServer struct {
 	DeviceUtils     mountmanager.DeviceUtils
 	MetadataService metadataservice.MetadataService
 	// TODO: Only lock mutually exclusive calls and make locking more fine grained
-	mux sync.Mutex
+	volumes sync.Map
```
davidz627: Lets just add a comment here explaining what this sync achieves. Something like "locks so that only one node operation can be done on a single volume (defined by a volume ID) at a time, any concurrent calls will return an AlreadyExists error".

davidz627:
> I like this approach. The only enhancement I can think of is hung/long operation detection. We may want to track operation start time, and log a message/metric if some operation is taking too long.

@hantaowang can you create an `Issue` on the driver for this. We can do it as a follow up but it doesn't need to be a part of this PR

pkg/gce-pd-csi-driver/node.go
```diff
@@ -221,6 +225,11 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
 		return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Volume Capability must be provided")
 	}
 
+	if _, alreadyExists := ns.volumes.LoadOrStore(volumeID, true); alreadyExists {
+		return nil, status.Error(codes.AlreadyExists, "An operation with the given Volume ID already exists")
```
davidz627: the correct error code according to the spec is actually: `10 ABORTED`

https://github.com/container-storage-interface/spec/blob/master/spec.md

hantaowang:
> > I like this approach. The only enhancement I can think of is hung/long operation detection. We may want to track operation start time, and log a message/metric if some operation is taking too long.
> 
> @hantaowang can you create an `Issue` on the driver for this. We can do it as a follow up but it doesn't need to be a part of this PR

Is this [issue](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/136) that @msau42 created the same one?

msau42:
I think specifically an issue for hung/long operation detection

pkg/mount-manager/fake-safe-mounter.go
```diff
@@ -38,10 +38,38 @@ func NewFakeSafeMounter() *mount.SafeFormatAndMount {
...
 }
+
+type FakeBlockingMounter struct {
+	*mount.FakeMounter
+	OkToRun   chan struct{}
+	OpStarted chan struct{}
+}
+
+func (mounter *FakeBlockingMounter) Mount(source string, target string, fstype string, options []string) error {
+	mounter.OpStarted <- struct{}{}
```
davidz627: nit: `mountOpStarted`? I think the mounter also does the unmount (maybe in 1.14+ only) so we should disambiguate

pkg/mount-manager/fake-safe-mounter.go
```diff
@@ -38,10 +38,38 @@ func NewFakeSafeMounter() *mount.SafeFormatAndMount {
...
+	<-mounter.OkToRun
+	return mounter.FakeMounter.Mount(source, target, fstype, options)
+}
+
+func NewFakeSafeBlockingMounter(okToTest, opStarted chan struct{}) *mount.SafeFormatAndMount {
+	fakeBlockingMounter := &FakeBlockingMounter{
+		FakeMounter: fakeMounter,
+		OkToRun:     okToTest,
```
davidz627: nit: `OkToRun` or `OkToTest`?

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -378,3 +387,75 @@ func TestNodeGetCapabilities(t *testing.T) {
...
+		}
+	} else {
+		t.Fatalf("Expected error: %v, got no error", codes.Aborted)
+	}
+
+	// Start second valid request vol2PublishTargetCReq
+	go runRequestInBackground(vol2PublishTargetCReq)
+
+	// Check that vol2PublishTargetCReq is able to run
+	<-opStarted
```
davidz627: in the case that `vol2PublishTargetCReq` is actually unable to run and fails for whatever reason won't this block forever?

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -378,3 +387,75 @@ func TestNodeGetCapabilities(t *testing.T) {
...
+
+	// Allow both operations to complete
+	okToTest <- struct{}{}
```
davidz627: i see now this is more like `runMountToComplete` or `continueMountOp` would renaming it be appropriate?

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -378,3 +387,75 @@ func TestNodeGetCapabilities(t *testing.T) {
...
+		TargetPath:        defaultTargetPath + "c",
+		StagingTargetPath: defaultStagingPath + "2",
+		Readonly:          false,
+		VolumeCapability:  stdVolCap,
+	}
+
+	runRequestInBackground := func(req *csi.NodePublishVolumeRequest) {
+		_, err := ns.NodePublishVolume(context.Background(), req)
+		if err != nil {
+			t.Fatalf("Unexpedted error: %v", err)
```
davidz627: nit: spellign

davidz627: did you confirm that if we hit the `Fatalf` inside a goroutine the test actually fails? I think go testing has some subtly weird things when it comes to goroutines

pkg/gce-pd-csi-driver/node_test.go
hantaowang: See the new names if they are better.

pkg/gce-pd-csi-driver/node_test.go
hantaowang: I changed it so that failures should not block anymore.

davidz627:
/assign @jingxu97 
please review

hantaowang:
```release-note
Concurrent operations for the same volume in the controller are now serialized.
```

PR #310, opened 2019-06-20
Updated Kubernetes dependency to v1.15 and fixed all dep errors

davidz627:
The `Gopkg.toml` file is now sorted and commented and all makes sense.

/assign @msau42


Gopkg.toml
```diff
@@ -24,65 +24,67 @@
...
 
 [[constraint]]
-  name = "google.golang.org/grpc"
-  version = "1.10.0"
+  version = "v1.15.0"
+  name = "k8s.io/kubernetes"
 
 [[constraint]]
-  branch = "release-1.11"
+  version = "kubernetes-1.15.0-alpha.3"
```
msau42: Is there an official 1.15 tag we can use now

davidz627: yeah looks like they were just published

davidz627:
@msau42 I believe I've resolved all remaining dep issues (mainly around the `klog` import) and all tests should be passing soon :)

Issue #307, opened 2019-06-20
Serialize controller operations on the same volume


msau42:
ie attach and detach should not be allowed to occur in parallel for the same volume

cc @davidz627

msau42:
I haven't thought much about if all of the controller operations, ie create/delete/expand also all need to be serialized together.

msau42:
For attach and detach, we could get into strange states like returning success from Detach while an Attach to the same node is in progress.

hantaowang:
If the problem is that there is a race condition on Attach ops for the same volume and the GCP layer does not handle this, then we can look into applying similar changes as #303 and serialize ops on the same volume.

msau42:
GCP layer does handle this, however, I think our CSI driver layer does not.

ie, 2 threads are processing in parallel, an attach and a detach.  One of them is going to be waiting for the other, but ordering of operation calls and checks between the two operations are not guaranteed at the moment.

PR #316, opened 2019-06-24
Add volume level serialization for controller operations

hantaowang:
Per Issue [307](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/307) using logic similar to #303 

Fixes: #307 

```release-note
Operations in the controller driver on a volume already being operated on now return an error.
```

### DIFF
--- a/pkg/common/volume_lock.go
+++ b/pkg/common/volume_lock.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	VolumeOperationAlreadyExistsFmt = "An operation with the given Volume ID %s already exists"
+)
+
+// VolumeLocks implements a map with atomic operations. It stores a set of all volume IDs
+// with an ongoing operation.
+type VolumeLocks struct {
+	locks sets.String
+	mux   sync.Mutex
+}
+
+func NewVolumeLocks() *VolumeLocks {
+	return &VolumeLocks{
+		locks: sets.NewString(),
+	}
+}
+
+// TryAcquire tries to acquire the lock for operating on volumeID and returns true if successful.
+// If another operation is already using volumeID, returns false.
+func (vl *VolumeLocks) TryAcquire(volumeID string) bool {
+	vl.mux.Lock()
+	defer vl.mux.Unlock()
+	if vl.locks.Has(volumeID) {
+		return false
+	}
+	vl.locks.Insert(volumeID)
+	return true
+}
+
+func (vl *VolumeLocks) Release(volumeID string) {
+	vl.mux.Lock()
+	defer vl.mux.Unlock()
+	vl.locks.Delete(volumeID)
+}
--- a/pkg/gce-cloud-provider/compute/fake-gce.go
+++ b/pkg/gce-cloud-provider/compute/fake-gce.go
@@ -49,7 +49,7 @@ type FakeCloudProvider struct {
 
 var _ GCECompute = &FakeCloudProvider{}
 
-func FakeCreateCloudProvider(project, zone string, cloudDisks []*CloudDisk) (*FakeCloudProvider, error) {
+func CreateFakeCloudProvider(project, zone string, cloudDisks []*CloudDisk) (*FakeCloudProvider, error) {
 	fcp := &FakeCloudProvider{
 		project:   project,
 		zone:      zone,
@@ -61,7 +61,6 @@ func FakeCreateCloudProvider(project, zone string, cloudDisks []*CloudDisk) (*Fa
 		fcp.disks[d.GetName()] = d
 	}
 	return fcp, nil
-
 }
 
 func (cloud *FakeCloudProvider) RepairUnderspecifiedVolumeKey(ctx context.Context, volumeKey *meta.Key) (*meta.Key, error) {
@@ -420,6 +419,22 @@ func (cloud *FakeCloudProvider) getGlobalSnapshotURI(snapshotName string) string
 		snapshotURITemplateGlobal,
 		cloud.project,
 		snapshotName)
+}
+
+type FakeBlockingCloudProvider struct {
+	*FakeCloudProvider
+	ReadyToExecute chan chan struct{}
+}
+
+// FakeBlockingCloudProvider's method adds functionality to finely control the order of execution of CreateSnapshot calls.
+// Upon starting a CreateSnapshot, it passes a chan 'executeCreateSnapshot' into readyToExecute, then blocks on executeCreateSnapshot.
+// The test calling this function can block on readyToExecute to ensure that the operation has started and
+// allowed the CreateSnapshot to continue by passing a struct into executeCreateSnapshot.
+func (cloud *FakeBlockingCloudProvider) CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*compute.Snapshot, error) {
+	executeCreateSnapshot := make(chan struct{})
+	cloud.ReadyToExecute <- executeCreateSnapshot
+	<-executeCreateSnapshot
+	return cloud.FakeCloudProvider.CreateSnapshot(ctx, volKey, snapshotName)
 }
 
 func notFoundError() *googleapi.Error {
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -42,6 +42,10 @@ type GCEControllerServer struct {
 	Driver          *GCEDriver
 	CloudProvider   gce.GCECompute
 	MetadataService metadataservice.MetadataService
+
+	// A map storing all volumes with ongoing operations so that additional operations
+	// for that same volume (as defined by Volume Key) return an Aborted error
+	volumeLocks *common.VolumeLocks
 }
 
 var _ csi.ControllerServer = &GCEControllerServer{}
@@ -139,6 +143,15 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", replicationType))
 	}
 
+	volumeID, err := common.KeyToVolumeID(volKey, gceCS.MetadataService.GetProject())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to convert volume key to volume ID: %v", err)
+	}
+	if acquired := gceCS.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer gceCS.volumeLocks.Release(volumeID)
+
 	// Validate if disk already exists
 	existingDisk, err := gceCS.CloudProvider.GetDisk(ctx, volKey)
 	if err != nil {
@@ -222,6 +235,11 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find volume with ID %v: %v", volumeID, err))
 	}
 
+	if acquired := gceCS.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer gceCS.volumeLocks.Release(volumeID)
+
 	err = gceCS.CloudProvider.DeleteDisk(ctx, volKey)
 	if err != nil {
 		return nil, status.Error(codes.Internal, fmt.Sprintf("unknown Delete disk error: %v", err))
@@ -257,6 +275,14 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 	if err != nil {
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find volume with ID %v: %v", volumeID, err))
 	}
+
+	// Acquires the lock for the volume on that node only, because we need to support the ability
+	// to publish the same volume onto different nodes concurrently
+	lockingVolumeID := fmt.Sprintf("%s/%s", nodeID, volumeID)
+	if acquired := gceCS.volumeLocks.TryAcquire(lockingVolumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, lockingVolumeID)
+	}
+	defer gceCS.volumeLocks.Release(lockingVolumeID)
 
 	// TODO(#253): Check volume capability matches for ALREADY_EXISTS
 	if err = validateVolumeCapability(volumeCapability); err != nil {
@@ -343,6 +369,14 @@ func (gceCS *GCEControllerServer) ControllerUnpublishVolume(ctx context.Context,
 		return nil, err
 	}
 
+	// Acquires the lock for the volume on that node only, because we need to support the ability
+	// to unpublish the same volume from different nodes concurrently
+	lockingVolumeID := fmt.Sprintf("%s/%s", nodeID, volumeID)
+	if acquired := gceCS.volumeLocks.TryAcquire(lockingVolumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, lockingVolumeID)
+	}
+	defer gceCS.volumeLocks.Release(lockingVolumeID)
+
 	instanceZone, instanceName, err := common.NodeIDToZoneAndName(nodeID)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("could not split nodeID: %v", err))
@@ -389,6 +423,12 @@ func (gceCS *GCEControllerServer) ValidateVolumeCapabilities(ctx context.Context
 	if err != nil {
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Volume ID is of improper format, got %v", volumeID))
 	}
+
+	if acquired := gceCS.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer gceCS.volumeLocks.Release(volumeID)
+
 	_, err = gceCS.CloudProvider.GetDisk(ctx, volKey)
 	if err != nil {
 		if gce.IsGCEError(err, "notFound") {
@@ -495,6 +535,11 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 	if err != nil {
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find volume with ID %v: %v", volumeID, err))
 	}
+
+	if acquired := gceCS.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer gceCS.volumeLocks.Release(volumeID)
 
 	// Check if snapshot already exists
 	var snapshot *compute.Snapshot
--- a/pkg/gce-pd-csi-driver/controller_test.go
+++ b/pkg/gce-pd-csi-driver/controller_test.go
@@ -1413,3 +1413,70 @@ func TestPickRandAndConsecutive(t *testing.T) {
 
 	}
 }
+
+func TestVolumeOperationConcurrency(t *testing.T) {
+	readyToExecute := make(chan chan struct{}, 1)
+	gceDriver := initBlockingGCEDriver(t, nil, readyToExecute)
+	cs := gceDriver.cs
+
+	vol1CreateSnapshotAReq := &csi.CreateSnapshotRequest{
+		Name:           name + "1A",
+		SourceVolumeId: testVolumeId + "1",
+	}
+	vol1CreateSnapshotBReq := &csi.CreateSnapshotRequest{
+		Name:           name + "1B",
+		SourceVolumeId: testVolumeId + "1",
+	}
+	vol2CreateSnapshotReq := &csi.CreateSnapshotRequest{
+		Name:           name + "2",
+		SourceVolumeId: testVolumeId + "2",
+	}
+
+	runRequest := func(req *csi.CreateSnapshotRequest) <-chan error {
+		response := make(chan error)
+		go func() {
+			_, err := cs.CreateSnapshot(context.Background(), req)
+			response <- err
+		}()
+		return response
+	}
+
+	// Start first valid request vol1CreateSnapshotA and block until it reaches the CreateSnapshot
+	vol1CreateSnapshotAResp := runRequest(vol1CreateSnapshotAReq)
+	execVol1CreateSnapshotA := <-readyToExecute
+
+	// Start vol1CreateSnapshotB and allow it to execute to completion. Then check for Aborted error.
+	// If a non Abort error is received or if the operation was started, then there is a problem
+	// with volume locking
+	vol1CreateSnapshotBResp := runRequest(vol1CreateSnapshotBReq)
+	select {
+	case err := <-vol1CreateSnapshotBResp:
+		if err != nil {
+			serverError, ok := status.FromError(err)
+			if !ok {
+				t.Fatalf("Could not get error status code from err: %v", err)
+			}
+			if serverError.Code() != codes.Aborted {
+				t.Errorf("Expected error code: %v, got: %v. err : %v", codes.Aborted, serverError.Code(), err)
+			}
+		} else {
+			t.Errorf("Expected error: %v, got no error", codes.Aborted)
+		}
+	case <-readyToExecute:
+		t.Errorf("The operation for vol1CreateSnapshotB should have been aborted, but was started")
+	}
+
+	// Start vol2CreateSnapshot and allow it to execute to completion. Then check for success.
+	vol2CreateSnapshotResp := runRequest(vol2CreateSnapshotReq)
+	execVol2CreateSnapshot := <-readyToExecute
+	execVol2CreateSnapshot <- struct{}{}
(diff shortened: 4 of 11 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -138,6 +143,11 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", replicationType))
 	}
 
+	if _, alreadyExists := gceCS.volumes.LoadOrStore(volKey.String(), true); alreadyExists {
```
msau42: It might be nice to wrap this into a new Interface instead of operating directly on sync.Map.  That way in the future, we can enhance this (to add things like timing/metrics measurements and long operation detection).

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -249,6 +264,11 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find volume with ID %v: %v", volumeID, err))
 	}
 
+	if _, alreadyExists := gceCS.volumes.LoadOrStore(volKey.String(), true); alreadyExists {
```
msau42: This gets a little tricky.  We actually want to serialize attach/detach operations on the same volume.  But we want to allow attach on the same volume in parallel if they're different nodes. (for rox access mode)

hantaowang: Are you then worried about seralizating volume operations for other calls like CreateSnapshot or DeleteSnapshot? I added it here just in case but I can remove those.

msau42: I think seralizing amongst (un)provision/attach/snapshot for the same volume is fine. Let's be safe for now, and we can considering relaxing it in the future if we find it's an issue. The only corner case is the attaching from ROX to multiple nodes. That should be allowed to happen in parallel.

hantaowang:
I created a VolumeLock struct that abstracts away the logic of the synchronized map. The earlier node level parallelization use it too now. Let me know if you think an actual Interface is required too.

pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -422,6 +421,42 @@ func (cloud *FakeCloudProvider) getGlobalSnapshotURI(snapshotName string) string
 		snapshotName)
 }
 
+type FakeBlockingCloudProvider struct {
+	*FakeCloudProvider
+	snapshotToExec chan SnapshotSourceAndTarget
+	readyToExecute chan struct{}
```
davidz627: nit: naming. `readyToExecute` doesn't really tell me much about who is ready to execute what. Maybe `createSnapshotCalled` is more accurate?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,76 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+
+	// Allow the vol2CreateSnapshotReq to complete, which it can concurrently with vol1CreateSnapshotReqA
+	snapshotToExecute <- sourceAndTargetForRequest(vol2CreateSnapshotReq)
+	if err = <-reqFinished; err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	// To clean up, allow the vol1CreateSnapshotReqA to complete
+	snapshotToExecute <- sourceAndTargetForRequest(vol1CreateSnapshotReqA)
+	if err = <-reqFinished; err != nil {
```
davidz627: we should probably loop through `reqFinished` until completion at the end. In general I want to avoid code where the implementer/reviewer has to carefully count the number of sends on a channel and hand-write a corresponding number of receives on a channel. This is prone to error and very difficult to reason about

pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -422,6 +421,42 @@ func (cloud *FakeCloudProvider) getGlobalSnapshotURI(snapshotName string) string
...
+// FakeBlockingCloudProvider's method adds two channels to the CreateSnapshot process in order to provide functionality to finely
+// control the order of execution of volume operations. readyToExecute signals that an CreateSnapshot operation has been called.
+// Then it cycles through the snapshotToExec channel, waiting for permission to actually create the snapshot.
+func (cloud *FakeBlockingCloudProvider) CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*compute.Snapshot, error) {
+	cloud.readyToExecute <- struct{}{}
+	for snapshotToExec := range cloud.snapshotToExec {
+		if snapshotToExec.Source == volKey.String() && snapshotToExec.Target == snapshotName {
+			break
+		} else {
+			cloud.snapshotToExec <- snapshotToExec
```
davidz627: this puts the function into a tight loop when there are items in `snapshotToExec` but they are not for this function. I would like to avoid this, but I've been thinking about it for a while and im honestly not sure how

/cc @misterikkit @msau42

davidz627: we could make this less general and just have 3 channels in the blocking cloud provider, one for each specific snapshot we can execute and just case on those based on which invocation of `CreateSnapshot` we're in. That would get rid of the tight loop as well as the requirement for the exported struct `SnapshotSourceAndTarget`. But that solution is also not ideal as it's definitely pretty janky and requires more channels making it more confusing in a different way.

Either way, don't spend too long thinking about this since it's test code and probably ok as is but its actually kind of an interesting problem. Maybe Jon or Michelle will have some ideas

pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -422,6 +421,42 @@ func (cloud *FakeCloudProvider) getGlobalSnapshotURI(snapshotName string) string
 		snapshotName)
 }
 
+type FakeBlockingCloudProvider struct {
+	*FakeCloudProvider
+	snapshotToExec chan SnapshotSourceAndTarget
```
misterikkit: Okay, I understand what your test is doing now.
1. make concurrent API calls
2. assert that conflicting operations do not happen concurrently

The fact that you return an aborted error makes the test a bit easier to write. If you assert an aborted error, you know that (a) the race condition is prevented, and (b) the test actually covered the race condition.

Synchronizing a test like this with channels is the right approach IMO, but it can be structured in a way that is easier to read and removes the tight poll loop.

My number one piece of advice is **prefer callbacks instead of channels in APIs**. (There's a really good gophercon talk about this, which I can't find at the moment.) An API that uses channels can have many ways to mis-use it. Should API users read, write, or close the channels? Do callers have control over the buffer size? Buffered vs unbuffered can make a semantic difference. It's difficult to picture how `FakeBlockingCloudProvider` would be used by a different unit test.

An API that uses channels to communicate can be transformed into a callback-based API. These are much more test friendly because callbacks can be wrapped in funcs that can interact with the test environment before and after the actual callback runs.

In this specific example, you could have a func value in the struct called `beforeSnapshot` which is called before you delegate into `FakeCloudProvider.CreateSnapshot`. That func can come from your test, where it is able to interact with local state, such as synchronization channels.

pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -422,6 +421,42 @@ func (cloud *FakeCloudProvider) getGlobalSnapshotURI(snapshotName string) string
...
+		if snapshotToExec.Source == volKey.String() && snapshotToExec.Target == snapshotName {
+			break
+		} else {
+			cloud.snapshotToExec <- snapshotToExec
+		}
+	}
+	return cloud.FakeCloudProvider.CreateSnapshot(ctx, volKey, snapshotName)
+}
+
+func FakeCreateBlockingCloudProvider(project, zone string, cloudDisks []*CloudDisk, snapshotToExec chan SnapshotSourceAndTarget, readyToExecute chan struct{}) (*FakeBlockingCloudProvider, error) {
```
misterikkit: It looks like you're following the style in this file, but `FakeCreate` sounds much weirder than `CreateFake` to me. (:

pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -422,6 +421,42 @@ func (cloud *FakeCloudProvider) getGlobalSnapshotURI(snapshotName string) string
...
+	fcp, err := FakeCreateCloudProvider(project, zone, cloudDisks)
+	fbcp := &FakeBlockingCloudProvider{
```
misterikkit: does this really need a constructor? Since this is just filling in struct fields, why not have the caller do that?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,76 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+func TestVolumeOperationConcurrency(t *testing.T) {
+	snapshotToExecute := make(chan gce.SnapshotSourceAndTarget, 3)
+	readyToExecute := make(chan struct{}, 2)
+	reqFinished := make(chan error, 2)
+
+	gceDriver := initBlockingGCEDriver(t, nil, snapshotToExecute, readyToExecute)
+	cs := gceDriver.cs
+
+	vol1CreateSnapshotReqA := &csi.CreateSnapshotRequest{
+		Name:           name + "1A",
```
misterikkit: What is `name` and why is it important for this test? I haven't read the other tests in this file, so you may be just going with the convention. From a github PR, finding the definition of `name` is a nuisance. I try to follow this testing philosophy when possible: https://abseil.io/tips/122

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,76 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+	vol1CreateSnapshotReqB := &csi.CreateSnapshotRequest{
+		Name:           name + "1B",
+		SourceVolumeId: testVolumeId + "1",
+	}
+	vol2CreateSnapshotReq := &csi.CreateSnapshotRequest{
+		Name:           name + "2",
+		SourceVolumeId: testVolumeId + "2",
+	}
+
+	runRequestInBackground := func(req *csi.CreateSnapshotRequest) {
```
misterikkit: I'd drop the "InBackground" from this func name. Prefer to name things after what they are or what they do, rather than name them after how they are used.

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,76 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+		}
+	}
+
+	// Start first valid request vol1CreateSnapshotReqA and block until it reaches the CreateSnapshot
+	go runRequestInBackground(vol1CreateSnapshotReqA)
+	<-readyToExecute
+
+	// Check that vol1CreateSnapshotReqB is rejected, due to same volume ID
+	// Also allow vol1CreateSnapshotReqB to complete, in case it is allowed to Mount
+	snapshotToExecute <- sourceAndTargetForRequest(vol1CreateSnapshotReqB)
```
misterikkit: The pattern you are using here can be described as, "Signal on a channel once the request has begun, and wait for a return signal before allowing the request to complete."

I like this pattern.

A lot of the complexity in this test comes from the fact that you are using two independent channels to do this. An alternative is to use a channel of channels. That is, each request starts by creating a reply channel and sending it to the unit test through the "main" channel. Then each request waits for a reply on its own channel, and the test has clearer control over which request is allowed to proceed.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -139,6 +143,11 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", replicationType))
 	}
 
+	if acquired := gceCS.volumes.TryAcquire(volKey.String()); !acquired {
```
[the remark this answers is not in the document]
hantaowang: Main point was because in CreateVolume I only have access to the key, unless I do something to convert the key to the ID.

pkg/gce-cloud-provider/compute/fake-gce.go
davidz627: agreed, I think `FakeCreate` was a brain fart

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -496,6 +528,11 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find volume with ID %v: %v", volumeID, err))
 	}
 
+	if acquired := gceCS.volumes.TryAcquire(volKey.String()); !acquired {
+		return nil, status.Error(codes.Aborted, fmt.Sprintf("An operation with the given Volume Key %s already exists", volKey.String()))
```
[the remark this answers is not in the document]
hantaowang: I think the repo style does not extract common error strings for anything else, so I'm going to leave that in there.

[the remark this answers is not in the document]
hantaowang: Actually I changed it. Be the change you want to see.

pkg/gce-pd-csi-driver/controller_test.go
hantaowang: I just tried implementing this, but it actually don't because of how the call is set up. The test interface blocks in the cloudprovider function call, while logic that aborts the operation occurs in the layer above that. So if the operation is aborted, it will never make it to the cloud provider call. This means that the channel of channels won't get a chance to create and send the reply channel.

hantaowang: The problem arises from the fact that merging the two independent channels means that the caller will need to block on the "main" channel in order to confirm that the operation has started. In the current setup, the operation that should be aborted has the "go ahead" SnapshotSourceAndTarget prequeued so that we can directly just go and check the error channel. In this set up, in order to give the go ahead we need the reply channel that is generated by the function call, but because that call is aborted, the reply channel isn't created.

pkg/gce-pd-csi-driver/controller_test.go
hantaowang: Name is a predefined name for the snapshot. I'm just appending an identifier to the end to get unique names with the same prefix. In actuality what name really is doesn't really matter, but I'm just sticking to the convention of using name for the snapshot name as much as possible.

pkg/gce-cloud-provider/compute/fake-gce.go
hantaowang: I like constructors, but am ok with removing this too.

pkg/gce-cloud-provider/compute/fake-gce.go
hantaowang: The thing with call backs is that while I guess it makes it more generic, it just moves the same logic that was in the BlockingCloudProvider struct into the test. I don't think it removes any of the complexity that's the issue,.

misterikkit: Splitting test logic across two files *is* complexity. As we discussed, I don't think your wrapper around the fake cloud provider needs to be in a separate file.

pkg/gce-pd-csi-driver/controller_test.go
misterikkit: I think you can still cause the first call to block while the second call gets an abort error. You don't need to block or unblock any threads for that. I'm a little sleep deprived, so maybe not understanding the concurrency correctly.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -60,6 +60,12 @@ type GCECompute interface {
 	GetSnapshot(ctx context.Context, snapshotName string) (*compute.Snapshot, error)
 	CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*compute.Snapshot, error)
 	DeleteSnapshot(ctx context.Context, snapshotName string) error
+	// Metadata
+	GetProject() string
```
davidz627: This breaks the abstraction, the project comes from the metadata service not the Compute service

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -118,32 +122,41 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
-			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume failed to get region from zones: %v", err))
+			return nil, status.Errorf(codes.InvalidArgument, "CreateVolume failed to get region from zones: %v", err)
 		}
 		volKey = meta.RegionalKey(name, region)
 	default:
-		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", replicationType))
+		return nil, status.Errorf(codes.InvalidArgument, "CreateVolume replication type '%s' is not supported", replicationType)
+	}
+
+	volumeID, err := common.KeyToVolumeID(volKey, gceCS.CloudProvider.GetProject())
```
davidz627: `gceCS.MetadataService.GetProject()`

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -293,12 +317,12 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
...
 	deviceName, err := common.GetDeviceName(volKey)
 	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("error getting device name: %v", err))
+		return nil, status.Errorf(codes.Internal, "error getting device name: %v", err)
 	}
 
 	attached, err := diskIsAttachedAndCompatible(deviceName, instance, volumeCapability, readWrite)
 	if err != nil {
-		return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("Disk %v already published to node %v but incompatbile: %v", volKey.Name, nodeID, err))
+		return nil, status.Errorf(codes.AlreadyExists, "Disk %v already published to node %v but incompatbile: %v", volKey.Name, nodeID, err)
```
davidz627: I'm not gonna make you do this now but in the future this type of unrelated change should go in a seperate PR

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -343,9 +367,15 @@ func (gceCS *GCEControllerServer) ControllerUnpublishVolume(ctx context.Context,
 		return nil, err
 	}
 
+	lockingVolumeID := fmt.Sprintf("%s/%s", nodeID, volumeID)
```
davidz627: add a comment here explaining why we allow concurrent different nodes

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,70 @@ func TestPickRandAndConsecutive(t *testing.T) {
 
 	}
 }
+
+func TestVolumeOperationConcurrency(t *testing.T) {
```
davidz627: cool! This is way more readable. I can figure out roughly whats going on in the first read through. Thanks for the review @misterikkit and great work refactoring @hantaowang

pkg/gce-pd-csi-driver/gce-pd-driver_test.go
```diff
@@ -22,13 +22,29 @@ import (
...
-	gceDriver := GetGCEDriver()
-	fakeCloudProvider, err := gce.FakeCreateCloudProvider(project, zone, cloudDisks)
+	fakeCloudProvider, err := gce.CreateFakeCloudProvider(project, zone, cloudDisks)
+	if err != nil {
+		t.Fatalf("Failed to create fake cloud provider: %v", err)
+	}
+	return initGCEDriverWithCloudProvider(t, fakeCloudProvider)
+}
+
+func initBlockingGCEDriver(t *testing.T, cloudDisks []*gce.CloudDisk, readyToExecute chan chan struct{}) *GCEDriver {
```
davidz627: nit: I'm still not a fan of `readyToExecute` as a name. I feel like the actual meaning of this channel is not that.

pkg/common/volume_lock.go
```diff
@@ -0,0 +1,56 @@
...
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	VolumeOperationAlreadyExistsFmt = "An operation with the given Volume ID %s already exists"
+)
+
+// VolumeLocks implements a map with atomic operations. It stores a set of all volume IDs
```
misterikkit: Thanks for adding the comment. It looks good as-is, but the second sentence seems like the more important part.

pkg/common/volume_lock.go
```diff
@@ -0,0 +1,56 @@
...
+	mux   sync.Mutex
+}
+
+func NewVolumeLocks() *VolumeLocks {
+	return &VolumeLocks{
+		locks: sets.NewString(),
+	}
+}
+
+func (vl *VolumeLocks) TryAcquire(volumeID string) bool {
```
misterikkit: This needs a docstring.

pkg/gce-pd-csi-driver/controller.go
misterikkit: +1

The added diffs make it harder to review this PR.

pkg/gce-pd-csi-driver/gce-pd-driver_test.go
hantaowang: What do you think a better the name would be?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,70 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+	vol1CreateSnapshotBReq := &csi.CreateSnapshotRequest{
+		Name:           name + "1B",
+		SourceVolumeId: testVolumeId + "1",
+	}
+	vol2CreateSnapshotReq := &csi.CreateSnapshotRequest{
+		Name:           name + "2",
+		SourceVolumeId: testVolumeId + "2",
+	}
+
+	runRequest := func(req *csi.CreateSnapshotRequest) chan error {
```
misterikkit: nit: to make this a little more readable, make the return type `<-chan error`

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,70 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+	// Start vol1CreateSnapshotB and allow it to execute to completion. Then check for Aborted error.
+	// If a non Abort error is received or if the operation was started, then there is a problem
+	// with volume locking
+	vol1CreateSnapshotBResp := runRequest(vol1CreateSnapshotBReq)
+	select {
+	case err := <-vol1CreateSnapshotBResp:
+		if err != nil {
+			serverError, ok := status.FromError(err)
+			if !ok {
+				t.Fatalf("Could not get error status code from err: %v", err)
```
misterikkit: Are all these `Fatalf` errors really fatal? In general, you should only use fatal if the test cannot proceed from that error.

pkg/gce-pd-csi-driver/controller_test.go
misterikkit: Yeah, thanks for putting in the effort. I think there are more potential gains on the readability of this test, but we're entering the point of diminishing returns.

pkg/gce-pd-csi-driver/gce-pd-driver_test.go
misterikkit: Maybe `snapshotRequests`? If the channel variable were purely local to the unit test code, it could have a more specific name like `snapshotStarted` or something like that.

davidz627: `executeCreateSnapshotCallback`, `continueCreateSnapshotCallback`? If you wan't it more general `continueCallback`? I'm honestly not sure so you can stick with `readyToExecute` if that makes the most sense.

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1413,3 +1413,70 @@ func TestPickRandAndConsecutive(t *testing.T) {
...
+	// Start vol1CreateSnapshotB and allow it to execute to completion. Then check for Aborted error.
+	// If a non Abort error is received or if the operation was started, then there is a problem
+	// with volume locking
+	vol1CreateSnapshotBResp := runRequest(vol1CreateSnapshotBReq)
+	select {
+	case err := <-vol1CreateSnapshotBResp:
+		if err != nil {
+			serverError, ok := status.FromError(err)
+			if !ok {
+				t.Errorf("Could not get error status code from err: %v", err)
```
davidz627: I think these are actually fatal errors. if we get `!ok` here that means `serverError` could be nil and we immediately have a `nil` dereference on the next line. I'd rather that the test stop and we exit cleanly and run all defers (with `t.Fatal`) rather than crash and burn

pkg/common/volume_lock.go
```diff
@@ -0,0 +1,58 @@
...
+	mux   sync.Mutex
+}
+
+func NewVolumeLocks() *VolumeLocks {
+	return &VolumeLocks{
+		locks: sets.NewString(),
+	}
+}
+
+// Tries to acquire the lock for operating on volumeID and returns true if successful.
```
davidz627: nit: docstring should start with Method name

pkg/mount-manager/fake-safe-mounter.go
```diff
@@ -52,35 +52,24 @@ func NewFakeSafeMounter() *mount.SafeFormatAndMount {
...
 
-type MountSourceAndTarget struct {
-	Source string
-	Target string
-}
-
-// FakeBlockingMounter's method adds two channels to the Mount process in order to provide functionality to finely
-// control the order of execution of Mount calls. readToMount signals that a Mount operation has been called.
-// Then it cycles through the mountToRun channel, waiting for permission to actually make the mount operation.
+// FakeBlockingMounter's method adds functionality to finely control the order of execution of Mount calls.
```
davidz627: nit: Godoc should begin with name of the element it describes (in this case `Mount`)

pkg/gce-pd-csi-driver/node.go
```diff
@@ -39,7 +38,7 @@ type GCENodeServer struct {
 
 	// A map storing all volumes with ongoing operations so that additional operations
```
jingxu97: update the comments after changing to volumeLocks?

pkg/gce-pd-csi-driver/node.go
```diff
@@ -74,10 +73,10 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume Volume Capability must be provided")
 	}
 
-	if _, alreadyExists := ns.volumes.LoadOrStore(volumeID, true); alreadyExists {
-		return nil, status.Error(codes.Aborted, fmt.Sprintf("An operation with the given Volume ID %s already exists", volumeID))
+	if acquired := ns.volumeLocks.TryAcquire(volumeID); !acquired {
```
jingxu97: do we allow to mount the same volume to different paths concurrently?

hantaowang: No, in our final design we decided not to support that to be safe because the cost of a bind mount is very small.

pkg/gce-pd-csi-driver/node.go
hantaowang: I'm not sure that a comment is actually even needed here, as what VolumeLocks does is explained in by common.VolumeLocks

Issue #306, opened 2019-06-19
Add test configs on Ubuntu


msau42:
So we can test xfs

/kind test

davidz627:
I was just in the middle of creating this issue when I found this. Definitely much needed. Should be an easy test config change (change the image family)

davidz627:
/remove-lifecycle stale
xref: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/305

msau42:
/reopen
The ability to test on ubuntu was added, but we don't have an actual config yet.

davidz627:
On my list of things to do this week - just configured the description incorrectly

davidz627:
Test configs added with GKE internally - still have some issues but close to resolution.

Issue #305, opened 2019-06-19
Add xfs utilities to driver


msau42:
They're missing

davidz627:
/remove-lifecycle stale
should probably do this ASAP

PR #317, opened 2019-06-24
Add support for Resize (Expand Volume) on the driver

### LINKED ISSUE #227: Support expand volume
CSI v1.1.0 added support for `ControllerExpandVolume` and `NodeExpandVolume`. This driver should implement them

davidz627:
/kind feature

Fixes #227

/assign @msau42 @saad-ali 
/cc @gnufied

```release-note
Adds support for ControllerExpandVolume and NodeExpandVolume
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,14 @@ This plugin is compatible with CSI versions [v1.0.0](https://github.com/containe
 
 ### Kubernetes Compatibility
 
-| GCE PD CSI Driver\Kubernetes Version | 1.10.5 - 1.11 | 1.12 | 1.13 | 1.14+
-|--------------------------------------|---------------|------|------|------|
-| v0.1.x.alpha                         | yes           | no   | no   | no   |
-| v0.2.x (alpha)                       | no            | yes  | no   | no   |
-| v0.3.x (beta)                        | no            | no   | yes  | yes  |
-| v0.4.x (beta)                        | no            | no   | yes  | yes  |
-| v0.5.x (beta)                        | no            | no   | no   | yes  |
-| dev                                  | no            | no   | no   | yes  |
+| GCE PD CSI Driver\Kubernetes Version | 1.10.5 - 1.11 | 1.12 | 1.13 | 1.14 | 1.15+|
+|--------------------------------------|---------------|------|------|------|------|
+| v0.1.x.alpha                         | yes           | no   | no   | no   | no   |
+| v0.2.x (alpha)                       | no            | yes  | no   | no   | no   |
+| v0.3.x (beta)                        | no            | no   | yes  | yes  | yes  |
+| v0.4.x (beta)                        | no            | no   | yes  | yes  | yes  |
+| v0.5.x (beta)                        | no            | no   | no   | yes  | yes  |
+| dev                                  | no            | no   | no   | no   | yes  |
 
 ### Known Issues
 
@@ -62,14 +62,13 @@ This driver supports only one topology key:
 `topology.gke.io/zone`
 that represents availability by zone.
 
-### Kubernetes Beta Features
+### Features in Development
 
-* Topology: Requires K8s 1.14+ on Master and Nodes and PD driver v0.5.0+
-
-### Kubernetes Alpha Features
-
-* Snapshots: Requires K8s 1.13+ on Master and PD driver v0.3.0+ with the alpha
-  overlay
+| Feature         | Stage | Min Kubernetes Master Version | Min Kubernetes Nodes Version | Min Driver Version | Deployment Overlay |
+|-----------------|-------|-------------------------------|------------------------------|--------------------|--------------------|
+| Topology        | Beta  | 1.14                          | 1.14                         | v0.5.0             | Stable             |
+| Snapshots       | Alpha | 1.13                          | Any                          | v0.3.0             | Alpha              |
+| Resize (Expand) | Alpha | 1.14                          | 1.14                         | dev                | Alpha              |
 
 ### Future Features
 
--- a/deploy/kubernetes/overlays/alpha/controller_add_resizer.yaml
+++ b/deploy/kubernetes/overlays/alpha/controller_add_resizer.yaml
@@ -1,0 +1,17 @@
+kind: StatefulSet
+apiVersion: apps/v1
+metadata:
+  name: csi-gce-pd-controller
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-resizer
+          imagePullPolicy: Always
+          image: quay.io/k8scsi/csi-resizer:canary
+          args:
+            - "--v=5"
+            - "--csi-address=/csi/csi.sock"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
--- a/deploy/kubernetes/overlays/alpha/rbac_add_resizer.yaml
+++ b/deploy/kubernetes/overlays/alpha/rbac_add_resizer.yaml
@@ -1,0 +1,32 @@
+# Resizer must be able to work with PVCs, PVs, SCs.
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: external-resizer-role
+rules:
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims/status"]
+    verbs: ["update", "patch"]
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["list", "watch", "create", "update", "patch"]
+
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: csi-resizer-role
+subjects:
+  - kind: ServiceAccount
+    name: csi-controller-sa
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: external-resizer-role
+  apiGroup: rbac.authorization.k8s.io
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -50,6 +50,7 @@ type GCECompute interface {
 	GetDiskSourceURI(volKey *meta.Key) string
 	GetDiskTypeURI(volKey *meta.Key, diskType string) string
 	WaitForAttach(ctx context.Context, volKey *meta.Key, instanceZone, instanceName string) error
+	ResizeDisk(ctx context.Context, volKey *meta.Key, requestBytes int64) (int64, error)
 	// Regional Disk Methods
 	GetReplicaZoneURI(zone string) string
 	// Instance Methods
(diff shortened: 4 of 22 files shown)


README.md
```diff
@@ -70,6 +70,8 @@ that represents availability by zone.
 
 * Snapshots: Requires K8s 1.13+ on Master and PD driver v0.3.0+ with the alpha
   overlay
+* Resize: Requires K8s 1.14+ on Master and Nodes and PD driver v0.6.0+ with the
```
msau42: Would the alpha/beta features section be better organized by a table?

msau42: Also add 1.15 to the table above?

deploy/kubernetes/overlays/alpha/controller_add_resizer.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: apps/v1
+metadata:
+  name: csi-gce-pd-controller
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-resizer
+          imagePullPolicy: Always
+          image: quay.io/k8scsi/csi-resizer:canary
```
msau42: We should work on publishing our own resizer image

deploy/kubernetes/overlays/alpha/rbac_add_resizer.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "update", "patch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims/status"]
+    verbs: ["update", "patch"]
+  - apiGroups: ["storage.k8s.io"]
```
msau42: I think storageclass is not needed anymore.
https://github.com/kubernetes-csi/external-resizer/blob/master/deploy/kubernetes/rbac.yaml

pkg/gce-pd-csi-driver/node.go
```diff
@@ -349,7 +351,36 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("ControllerExpandVolume volume ID is invalid: %v", err))
+	}
+
+	devicePath, err := ns.getDevicePath(volumeID, "")
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Error when getting device path for %s: %v", volumeID, err))
+	}
+
+	resizer := resizefs.NewResizeFs(ns.Mounter)
```
msau42: If this is a raw block volume, we should skip trying to resize the filesystem.

davidz627: `resizer.Resize` already checks for `fstype` and skips resize if no `fs` is found (block device).

deploy/kubernetes/overlays/alpha/controller_add_resizer.yaml
davidz627: yeah, I can contact releng about this. Lets track seperatly from this PR though

davidz627:
@msau42 addressed your comments

pkg/gce-pd-csi-driver/node.go
msau42: Are we able to unit test this? If not, e2e test?

davidz627: wrote a unit test, it's kind of insane. suggestions welcome lol

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -328,6 +341,87 @@ func TestNodeStageVolume(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Logf("Test case: %s", tc.name)
+
+		execCallback := func(cmd string, args ...string) ([]byte, error) {
+			if cmd == "blkid" {
+				if tc.blockDevice {
+					// blkid returns exit code 2 when run on unformatted device
+					return nil, utilexec.CodeExitError{
+						Err:  errors.New("this is an exit error"),
+						Code: 2,
```
msau42: hopefully it doesn't overload 2 with any other error

pkg/gce-pd-csi-driver/node.go
msau42: Thanks for adding a unit test. We should probably add a e2e test anyway, so that we can actually catch potential distro issues too.

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -328,6 +341,87 @@ func TestNodeStageVolume(t *testing.T) {
...
+		{
+			name: "block device expand",
+			req: &csi.NodeExpandVolumeRequest{
+				VolumeId:   volumeID,
+				VolumePath: "some-path",
+				CapacityRange: &csi.CapacityRange{
+					RequiredBytes: resizedBytes,
+				},
+			},
+			blockDevice: true,
```
msau42: does expRespBytes need to be set?

If for some reason, nodeexpand got called before controllerexpand succeeded, do we want to blindly return success?

pkg/gce-pd-csi-driver/node.go
```diff
@@ -365,7 +367,38 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
+	resized, err := resizer.Resize(devicePath, "")
```
msau42: what happens if it's already resized?

pkg/gce-pd-csi-driver/node.go
```diff
@@ -365,7 +367,38 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("Error when resizing volume %s: %v", volKey.String(), err))
+
+	}
+
+	resp := &csi.NodeExpandVolumeResponse{}
+	if resized {
+		resp.CapacityBytes = reqBytes
```
msau42: do we need to validate that the new size is actually reqBytes?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -593,6 +594,65 @@ func (cloud *CloudProvider) CreateSnapshot(ctx context.Context, volKey *meta.Key
...
+	if err != nil {
+		return -1, fmt.Errorf("failed to resize zonal volume %v: %v", volKey.String(), err)
+	}
+
+	err = cloud.waitForZonalOp(ctx, op, volKey.Zone)
+	if err != nil {
+		return -1, fmt.Errorf("failed waiting for op for zonal resize for %s: %v", volKey.String(), err)
+	}
+
+	return requestGb, nil
```
msau42: do we want to validate the size after success?

pkg/gce-pd-csi-driver/identity.go
```diff
@@ -58,6 +59,13 @@ func (gceIdentity *GCEIdentityServer) GetPluginCapabilities(ctx context.Context,
 					},
 				},
 			},
+			{
+				Type: &csi.PluginCapability_VolumeExpansion_{
+					VolumeExpansion: &csi.PluginCapability_VolumeExpansion{
+						Type: csi.PluginCapability_VolumeExpansion_ONLINE,
+					},
+				},
+			},
```
gnufied: We probably want to add both ONLINE and OFFLINE capabilities here.

pkg/gce-pd-csi-driver/node.go
gnufied: This will not work for file systems that require device mount path though.. I think you will have to pass deviceMountPath to the `Resize` function as well. it can be extracted from node expand request.

gnufied: *cough* *cough* - xfs

msau42: also semi related: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/305 and https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/306

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -328,6 +341,87 @@ func TestNodeStageVolume(t *testing.T) {
...
+	volumeID := "project/test001/zones/c1/disks/testDisk"
+	testCases := []struct {
+		name         string
+		req          *csi.NodeExpandVolumeRequest
+		blockDevice  bool
+		expRespBytes int64
+		expErrCode   codes.Code
+	}{
+		{
+			name: "ext4 fs expand",
```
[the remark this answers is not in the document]
davidz627: `blkid` returns some fs config information that contains the string `ext4` for the `ext4` case. I can actually augment the test to do `xfs` as well

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -328,6 +341,87 @@ func TestNodeStageVolume(t *testing.T) {
...
+				},
+			},
+			blockDevice: true,
+		},
+	}
+	for _, tc := range testCases {
+		t.Logf("Test case: %s", tc.name)
+
+		execCallback := func(cmd string, args ...string) ([]byte, error) {
+			if cmd == "blkid" {
```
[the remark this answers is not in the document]
davidz627: not that I am aware of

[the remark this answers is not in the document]
davidz627: I just kind of wrote it, I could contribute it back upstream later

pkg/gce-pd-csi-driver/node_test.go
davidz627: > If the specified token was not found, or no (specified) devices could be identified, an exit code of 2 is returned. 

we search specifically for tokens `TYPE` and `PTTYPE`. I guess this error could also occur for any other reason the tokens don't exist (of which I can't think of any currently).

> For usage or other errors, an exit code of 4 is returned.

pkg/gce-pd-csi-driver/node_test.go
davidz627: I was attempting to interpret the spec (specifically the `OPTIONAL` bit):
> // The capacity of the volume in bytes. This field is OPTIONAL.
  int64 capacity_bytes = 1;

In that I think since after the `NodeExpandVolume` the `volume` which for `NodeExpandVolume` I feel refers to the fs, the block volume has no `capacity_bytes`. I could be easily convinced otherwise though, this is just my best effort guess

pkg/gce-pd-csi-driver/node.go
davidz627: > what happens if it's already resized?

I believe `resize2fs` just returns success and therefore this `resizer.Resize` will return success. Which is pretty much what we expect.
Not sure about the behavior of the `xfs` tool but I assume its the same/similar since it also doesn't take `size` as an arg, it just resizes to the size of the underlying partition

pkg/gce-pd-csi-driver/node.go
davidz627: sure, I guess we could do a `df` but is that really necessary?

pkg/gce-cloud-provider/compute/gce-compute.go
davidz627: The `ControllerExpandVolume` layer could do a `GetDisk` and check the size. Same question on the node side, if the GCE backend returns success do we really need to double check? I don't think we do this for any of the other calls

pkg/gce-pd-csi-driver/node.go
davidz627: sure, we can't do `xfs` just yet since we don't bring up our e2e tests with an `xfs` distro but that's a seperate issue. I can add one for the block device case

pkg/gce-pd-csi-driver/node_test.go
msau42: cc @gnufied for spec interpretation and why is capacity in the response optional?  Does the resizer use the returned capacity to set the PVC capacity? I don't think it specifically refers to fs/block anywhere in NodeExpandVolume. fs was used as an example.

pkg/gce-pd-csi-driver/node.go
msau42: Since we don't actually pass in a size to resizefs, how do we know the size we return is actually what was requested?

pkg/gce-cloud-provider/compute/gce-compute.go
msau42: For controller resize, I don't think it's as necessary since we pass in the size to the operation

pkg/gce-pd-csi-driver/node.go
davidz627: added validation

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -199,7 +199,7 @@ var _ = Describe("GCE PD CSI Driver", func() {
 		Expect(cloudDisk.SizeGb).To(Equal(newSizeGb))
 
 		// Resize node
-		err = client.NodeExpandVolume(volID, publishDir, newSizeGb)
+		_, err = client.NodeExpandVolume(volID, publishDir, newSizeGb)
```
msau42: We may also want an "offline" test case where NodeExpand is called after NodeStage, but before NodePublish

davidz627: done

pkg/gce-pd-csi-driver/node_test.go
gnufied: `capacity_bytes` in response and `capacity_range` in request is optional for `NodeExpandVolume` and the reason this is optional is because a CO may choose to expand the volume to full available size for fs expansion (if applicable). Obviously knowing whether to set `capacity_range` or not depends on ability of caller to determine what kind of volume it is dealing with when making `NodeExpandVolume` call. For a plugin author though - it always a good idea to set `capacity_bytes` in response whenever possible.

Currently we aren't using the return size from this call for setting `pvc.Status.Capacity` but another CO or a different version of k8s may use the returned size.

gnufied: One potential problem with setting return value of this call to `PVC.Status.Capacity` is - after formatting with a FS, the result disk size might be smaller than what user requested but we do not want to trigger another resize of underlying PV for handling such a case.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -365,7 +368,70 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
+			return nil, status.Errorf(codes.Internal, "ControllerExpandVolume resize requested for %v but after resize volume was size %v", common.BytesToGb(reqBytes), gotSizeGb)
+		}
+	} else {
+		// Some sort of formatted volume
+		gotSizeGb, err := ns.getFSSizeGb(devicePath)
+
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "ControllerExpandVolume resize could not get fs size of %s: %v", volumePath, err)
+		}
+		if gotSizeGb != common.BytesToGb(reqBytes) {
```
gnufied: exact equality is going to be a problem, also as I mentioned somewhere below formatted disk might be smaller than underlying volume.

davidz627: changed to only check exact equality for the actual block device: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/317/commits/3a7e29fe2421c66f6d22d73eb6d06e48242f6ad5#diff-37c748634a23b1ed91abeaa8a6d21525R405

Ignoring fs size for now

pkg/gce-pd-csi-driver/node_test.go
davidz627: got it. Setting `response` size to equal `request` if successful resize finished. We also ignore the capacity range in the request for now but I've created a tracking issue for that

davidz627:
@gnufied @msau42 
made requested changes

pkg/gce-pd-csi-driver/node.go
```diff
@@ -365,7 +368,72 @@ func (ns *GCENodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGe
...
+		return nil, status.Error(codes.Internal, fmt.Sprintf("ControllerExpandVolume error when resizing volume %s: %v", volKey.String(), err))
+
+	}
+
+	// Check the block size
+	gotBlockSizeBytes, err := ns.getBlockSizeBytes(devicePath)
+	if err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("ControllerExpandVolume error when getting size of block volume at path %s: %v", devicePath, err))
+	}
+	if gotBlockSizeBytes != reqBytes {
```
msau42: Maybe this should be got >= req?  I believe GCE rounds up (?) to the nearest Gb?

gnufied: I am not sure if we need to check disk size at all here.

davidz627: think im going to continue to verify disk size but just check for `got >= req`. Just as a quick sanity check. It's not super necessary but might catch errors if for some reason the disk is not resized properly

PR #319, opened 2019-06-25
[WIP] Gke integration tests

hantaowang:
This PR clarifies versioning with the deployment strategies gce and gke. 
- test-version is used to determine the k8s version to test with. 
- kube-version and gke-cluster-version is used to determine the cluster version to boot, depeonding on the deployment strategy.
- local-k8s-dir points to a rebuilt kubernetes directory. If this flag is set, then neither kube-version nor test-version should be because the local version is used for both cluster and tests.
- if using GKE: either set (gke-cluster-version and test-version) or (gke-cluster-version and local-k8s-dir)
- if using GCE: either set (local-k8s-dir) or (kube-version and test-version)

It also cleans up some code with the `ensureVariable`, `ensureFlag`, `getTempDir`, `removeTempDir` helper functions.

It also makes certain time consuming tasks run concurrently. Tell me if this is not desired.

```release-note
Changed how versioning flags work with the deployment strategy and adds a test-version flag to separate from test and cluster versioning.
```


test/k8s-integration/main.go
```diff
@@ -36,11 +37,12 @@ var (
 	teardownDriver   = flag.Bool("teardown-driver", true, "teardown the driver after the e2e test")
 	bringupCluster   = flag.Bool("bringup-cluster", true, "build kubernetes and bringup a cluster")
 	gceZone          = flag.String("gce-zone", "", "zone that the gce k8s cluster is created/found in")
-	kubeVersion      = flag.String("kube-version", "master", "version of Kubernetes to download and use")
+	kubeVersion      = flag.String("kube-version", "", "version of Kubernetes to download and use for the cluster")
+	testVersion      = flag.String("test-version", "", "version of Kubernetes to download and use for tests")
 	kubeFeatureGates = flag.String("kube-feature-gates", "", "feature gates to set on new kubernetes cluster")
 	localK8sDir      = flag.String("local-k8s-dir", "", "local kubernetes/kubernetes directory to run e2e tests from")
```
davidz627: can you improve the comment here to specify that when using local k8s dir it will use this for both `kube-version` and `test-version`

test/k8s-integration/main.go
```diff
@@ -157,69 +178,134 @@ func handle() error {
...
 
+	// Create temporary directories for kubernetes builds
+	k8sParentDir := getTempDir()
+	k8sDir := filepath.Join(k8sParentDir, "kubernetes")
+	testParentDir := getTempDir()
+	testDir := filepath.Join(testParentDir, "kubernetes")
+	defer removeTempDir(k8sParentDir)
+	defer removeTempDir(testParentDir)
+
+	numTasks := 4
```
davidz627: Can it be a `const`? Why is the numTasks=4, could probably use a comment

test/k8s-integration/main.go
```diff
@@ -157,69 +178,134 @@ func handle() error {
...
+		k8sDir = *localK8sDir
+		k8sDependencyChan <- true
+	}
+
+	// If test version is set, then download and build Kubernetes to run K8s tests
+	// Otherwise, either kube version is set (which implies GCE) or a local K8s dir is being used
+	if len(*testVersion) != 0 && *testVersion != *kubeVersion {
+		go func() {
+			// TODO: Build only the tests
+			err := downloadKubernetesSource(pkgDir, testParentDir, *testVersion)
```
davidz627: can we download only one version of the source and just copy it if we need it for test dir

test/k8s-integration/main.go
```diff
@@ -157,69 +178,134 @@ func handle() error {
...
+		errChan <- nil
+		k8sDir = *localK8sDir
+		k8sDependencyChan <- true
+	}
+
+	// If test version is set, then download and build Kubernetes to run K8s tests
+	// Otherwise, either kube version is set (which implies GCE) or a local K8s dir is being used
+	if len(*testVersion) != 0 && *testVersion != *kubeVersion {
+		go func() {
+			// TODO: Build only the tests
```
davidz627: `make WHAT=test/e2e/e2e.test`

test/k8s-integration/main.go
```diff
@@ -157,69 +178,134 @@ func handle() error {
...
+			err = buildKubernetes(testDir)
+			if err != nil {
+				errChan <- fmt.Errorf("failed to build Kubernetes: %v", err)
+				return
+			}
+			errChan <- nil
+		}()
+	} else {
+		testDir = k8sDir
+		errChan <- nil
```
davidz627: kind of odd to me that we need to push nil errors onto the `errChan` like this. Feel like we should be able to have something watch the `errChan` for any actual errors and halt the entire test suite (and execute all defers for cleanup).

Maybe something like a watcher that watches the error channel that pushes a cancel signal to a context that all goroutines take in. Let me know if this would make it more complicated, but I feel like it's more extensible so we don't have to do an error prone counting of errChan signals as we add/remove things from the integration testing

test/k8s-integration/main.go
```diff
@@ -157,69 +178,134 @@ func handle() error {
...
+	k8sParentDir := getTempDir()
+	k8sDir := filepath.Join(k8sParentDir, "kubernetes")
+	testParentDir := getTempDir()
+	testDir := filepath.Join(testParentDir, "kubernetes")
+	defer removeTempDir(k8sParentDir)
+	defer removeTempDir(testParentDir)
+
+	numTasks := 4
+	errChan := make(chan error, numTasks)
+	k8sDependencyChan := make(chan bool, 1)
```
davidz627: can we use named waitgroups to capture dependency chains where certain operations have to wait for groups of other operations to complete before we can proceed.

test/k8s-integration/main.go
hantaowang: I think it makes it a lot more complicated dependency wise, so I'll think about it.

hantaowang:
I'll let this PR sit for now, but I've already moved some of the small changes into other PRs. See #322 for the utility changes and #323 for the test-version flag related changes. When those are merged I will open a PR to break apart the huge `main.go`. Then I will look into the concurrency changes - if time permits.

PR #322, opened 2019-06-26
add flag checking and temp dir utilities

hantaowang:
Per #319, break up the PR into small ones. This first one adds some utilities that make the flag checking process more readable and uses a cleaner temp directory creation and deletion system.

```release-note
None
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


test/k8s-integration/main.go
```diff
@@ -70,45 +71,40 @@ func init() {
...
-	}
+	ensureVariable(saFile, true, "service-account-file is a required flag")
+	ensureVariable(deployOverlayName, true, "deploy-overlay-name is a required flag")
+	ensureVariable(testFocus, true, "test-focus is a required flag")
+	ensureVariable(gceZone, true, "gce-zone is a required flag")
 
-	if !*bringupCluster && len(*kubeFeatureGates) > 0 {
-		klog.Fatalf("kube-feature-gates set but not bringing up new cluster")
+	if *migrationTest {
+		ensureVariable(storageClassFile, false, "storage-class-file and migration-test cannot both be set")
```
davidz627: will adding a helper functions for `XOR` of variable, `OR` of variable,  `AND` of variables simplify this?

like migration and storage class could be
`ensureVariablesXOR(storageClassFile, migrationTest)` <- would be nice if it can somehow reflect the flag name for a standard error message for `XOR` type variables

test/k8s-integration/main.go
```diff
@@ -602,3 +597,32 @@ func deleteImage(stagingImage, stagingVersion string) error {
 	}
 	return nil
 }
+
+func getTempDir() string {
+	dir, err := ioutil.TempDir("", "")
```
davidz627: can we prefix or store all these tmp dirs in a way that identifies this as from this test framework. like
`gcp-pd-driver-tmp` or something. Not sure what a canonical name will look like

test/k8s-integration/main.go
```diff
@@ -602,3 +597,32 @@ func deleteImage(stagingImage, stagingVersion string) error {
 	}
 	return nil
 }
+
+func getTempDir() string {
```
davidz627: nit: is this more of a `generateUniqueTmpDir`

test/k8s-integration/main.go
```diff
@@ -602,3 +597,32 @@ func deleteImage(stagingImage, stagingVersion string) error {
...
+	dir, err := ioutil.TempDir("", "")
+	if err != nil {
+		klog.Fatalf("Error creating temp dir: %v", err)
+	}
+	return dir
+}
+
+func removeTempDir(dir string) {
```
davidz627: nit: this is just a `removeDir`

test/k8s-integration/main.go
```diff
@@ -602,3 +597,32 @@ func deleteImage(stagingImage, stagingVersion string) error {
...
+	err := os.RemoveAll(dir)
+	if err != nil {
+		klog.Fatalf("Error removing temp dir: %v", err)
+	}
+}
+
+func ensureVariable(v *string, set bool, msgOnError string) {
```
davidz627: nit: if we're going to do more helper function for the flag handling we could break all of those out into a separate file

hantaowang: I'll do this as part of the break things apart PR.

test/k8s-integration/main.go
hantaowang: Hmm, I'm not convinced that this would make things easier to understand. It will be cleaner as in the flag handling part will have less lines of code, but I feel like it will be hard to understand whats going on when you have to think about boolean operators.

davidz627: talked offline, lets think about this some more and maybe do it in a follow-up PR if time permits

PR #323, opened 2019-06-26
Separate test k8s

hantaowang:
Per #319, break up the PR into small ones. This one builds onto #322 and adds the test-version flag. It also separates k8s building and cluster booting, so that you can boot a cluster without having to build k8s. The rules for how flags can be used together is defined here:
- test-version is used to determine the k8s version to test with. 
- kube-version and gke-cluster-version is used to determine the cluster version to boot, depeonding on the deployment strategy.
- local-k8s-dir points to a rebuilt kubernetes directory. If this flag is set, then neither kube-version nor test-version should be because the local version is used for both cluster and tests.
- if using GKE: either set (gke-cluster-version and test-version) or (gke-cluster-version and local-k8s-dir)
- if using GCE: either set (local-k8s-dir) or (kube-version and test-version)

Do not merge until after #322 is merged. Then master will be rebased into this PR.

```release-note
None
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -37,12 +37,12 @@ var (
 	teardownDriver   = flag.Bool("teardown-driver", true, "teardown the driver after the e2e test")
 	bringupCluster   = flag.Bool("bringup-cluster", true, "build kubernetes and bringup a cluster")
 	gceZone          = flag.String("gce-zone", "", "zone that the gce k8s cluster is created/found in")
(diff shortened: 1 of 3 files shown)


davidz627:
/lgtm
/hold
since this depends on the commit from the other PR which has not yet merged

davidz627:
/approve
Please remove integration test binary from the commit and add to `.gitignore` for lgtm :)

PR #330, opened 2019-06-28
update gofmt to not search unnecessary places and report files changed

hantaowang:
```release-note
None
```

### DIFF
--- a/hack/update-gofmt.sh
+++ b/hack/update-gofmt.sh
@@ -18,4 +18,4 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-find . -name "*.go" | grep -v "\/vendor\/" | xargs gofmt -s -w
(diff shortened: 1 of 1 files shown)


hack/update-gofmt.sh
```diff
@@ -18,4 +18,4 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-find . -name "*.go" | grep -v "\/vendor\/" | xargs gofmt -s -w
+find . -not -path "*/vendor/*" -not -path "./test/k8s-integration/src/*" -name "*.go" | xargs gofmt -s -w -l
```
davidz627: > To ignore a whole directory tree, use -prune rather than checking every file in the tree. For example, to skip the directory 'src/emacs' and all files and directories under it, and print the names of the other files found, do something like this:
```find . -path ./src/emacs -prune -o -print```

> Note that the pattern match test applies to the whole file name, starting from one of the start points named on the command line. It would only make sense to use an absolute path name here if the relevant start point is also an absolute path. This means that this command will never match anything:

lol

PR #331, opened 2019-07-04
LOW PRIORITY Fix random lint errors

davidz627:
/kind cleanup

For `Id -> ID` see: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Fixes a bunch of random lint errors because its late and I don't want to think but I'm waiting for a ride.

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
+# GCP Compute Persistent Disk CSI Driver
+
 WARNING: This driver is beta and should not be used in performance critical applications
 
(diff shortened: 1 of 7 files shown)


hantaowang:
The test has failed twice now:
https://prow.k8s.io/?repo=kubernetes-sigs%2Fgcp-compute-persistent-disk-csi-driver&pull=331&author=davidz627
on different tests, with the rest skipped.

PR #339, opened 2019-07-12
bugfix-222 Driver should check socket parent directory before trying …

### LINKED ISSUE #222: Driver should check socket parent directory before trying to bind it
While testing the driver, we found that in the server's code https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-pd-csi-driver/server.go#L85,  if the socket parent directory does not exit, the driver does not check it and just tries to call Listen to the address. So the error message returned from Listen call does not give the clear reason why it fail. It would be better to check whether the parent directory exist or not before listening to the socket address.

zhucan:
**What type of PR is this?**
/kind bug

if the socket parent directory does not exit, the driver does not check it and just tries to call Listen to the address. So the error message returned from Listen call does not give the clear reason why it fail. It would be better to check whether the parent directory exist or not before listening to the socket address.

Fixes #222

@jingxu97 
@davidz627 

```release-note
Driver should check socket parent directory before trying to bind it
```

### DIFF
--- a/pkg/gce-pd-csi-driver/server.go
+++ b/pkg/gce-pd-csi-driver/server.go
@@ -18,6 +18,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"path/filepath"
 	"sync"
 
 	"google.golang.org/grpc"
@@ -90,6 +91,13 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
 		addr = u.Host
 	} else {
 		klog.Fatalf("%v endpoint scheme not supported", u.Scheme)
+	}
+
+	if u.Scheme == "unix" {
+		listenDir, _ := filepath.Split(addr)
+		if _, err := os.Stat(listenDir); err != nil {
+			klog.Fatalf("Failed to stat %s, error: %s", listenDir, err.Error())
+		}
 	}
 
 	klog.V(4).Infof("Start listening with scheme %v, addr %v", u.Scheme, addr)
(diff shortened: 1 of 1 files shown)


pkg/gce-pd-csi-driver/server.go
```diff
@@ -92,6 +93,13 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
 		klog.Fatalf("%v endpoint scheme not supported", u.Scheme)
 	}
 
+	if u.Scheme == "unix" {
```
davidz627: could this just be moved into the same `if u.Scheme == "unix"` block above

pkg/gce-pd-csi-driver/server.go
```diff
@@ -92,6 +93,13 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
...
+		listenDir, _ := filepath.Split(addr)
```
davidz627: `filepath.Dir` could be better here since you drop the last output anyway

pkg/gce-pd-csi-driver/server.go
```diff
@@ -92,6 +93,13 @@ func (s *nonBlockingGRPCServer) serve(endpoint string, ids csi.IdentityServer, c
...
+		if _, err := os.Stat(listenDir); err != nil {
```
davidz627: would prefer a check for an `os.IsNotExist(err)` explicitly and the error message thrown could be more illuminating such as: "Expected Kubelet plugin watcher to create parent dir *dirname* but did not find such a di"

zhucan:
> thanks for picking this up @zhucan! Just some minor comments, feel free to chat on slack (i'm davidz627) if you have any questions

copy that, thanks

davidz627:
Thanks @zhucan

PR #344, opened 2019-07-16
fix a bug where keys are not properly translated to volume id

hantaowang:
**What type of PR is this?**
/kind failing-test

Flaky tests

@davidz627 @saad-ali 

```release-note
Fixed a bug that causes disks in the same zone/region to be provisioned serially
```

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -76,11 +76,11 @@ func VolumeIDToKey(id string) (*meta.Key, error) {
 func KeyToVolumeID(volKey *meta.Key, project string) (string, error) {
 	switch volKey.Type() {
 	case meta.Zonal:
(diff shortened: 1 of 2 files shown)


pkg/common/utils.go
```diff
@@ -76,9 +76,9 @@ func VolumeIDToKey(id string) (*meta.Key, error) {
 func KeyToVolumeID(volKey *meta.Key, project string) (string, error) {
 	switch volKey.Type() {
 	case meta.Zonal:
-		return fmt.Sprintf(volIDZonalFmt, project, volKey.Zone, volKey.Zone), nil
+		return fmt.Sprintf(volIDZonalFmt, project, volKey.Zone, volKey.Name), nil
```
davidz627: please add unit test for `KeyToVolumeID`. We should have done that when adding the function in the first place, oops

hantaowang:
@davidz627 @jingxu97 ready for review thanks!

PR #345, opened 2019-07-16
Add forced variable num-nodes so that we can give it to kubetest to force tests to run that have high node requirements

davidz627:
/kind flake

For some reason some migration tests get skipped every once in a while because "number of nodes is not enough" when it only needs 2. For some reason the "num-nodes" variable in the e2e binary that's supposed to be dynamically gathered is showing `-1` (indicating failure to dynamically get the read/schedulable nodes from api server sometimes) so to solve I provide the num nodes flag every time.

I'm currently looking into why the e2e binary is not auto-setting this variable sometimes but that might be a larger fix

/assign @msau42 
/cc @hantaowang 

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 
 	"k8s.io/klog"
 )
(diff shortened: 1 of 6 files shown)


davidz627:
What is super weird is that I'm observing `NumNodes == -1` where this loc would suggest that is impossible:

https://github.com/kubernetes/kubernetes/blob/a10eced56b9100c30489b10c7a01e685735f6f0b/test/e2e/e2e.go#L110
since it will get some non-negative len of a list that is generated from schedulable nodes.

PR #346, opened 2019-07-16
Decrease verbosity to 7 to avoid verbosity 8 REST request/response spew

davidz627:
Too many logs.

/assign @msau42 @hantaowang 

/kind cleanup

```release-note
NONE
```


davidz627:
@msau42 changed to 5 to get rid of some other client-go logs. minimum of 4 to get logs I'm interested in

PR #347, opened 2019-07-17
Give project variable to test framework so it can do things like list disks

davidz627:
/kind failing-test

A lot of the older storage tests do things like listing disks on GCE directly from the test framework, the framework needs the project ID to function as it runs directly off the test client (not the server (kubernetes)).

Note: `$PROJECT` is a default env var in the test-infra

/assign @msau42 @hantaowang 

```release-note
NONE
```


davidz627:
got a replacement on the way when I just move to kubetest this should fix everything

PR #349, opened 2019-07-17
Migrate to k8s integration tests to kubetest

davidz627:
/kind cleanup
/kind failing-test

/assign @msau42 @hantaowang 

Alright this should just fix a bunch of stuff. Once it's done and stable I think there are a bunch of flags we can probably remove that were put in to work around the fact we weren't using kubetest :(

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -187,7 +187,6 @@ func handle() error {
 	k8sDir := filepath.Join(k8sParentDir, "kubernetes")
 	testParentDir := generateUniqueTmpDir()
 	testDir := filepath.Join(testParentDir, "kubernetes")
-	k8sBuildBinDir := k8sInDockerBuildBinDir
 	defer removeDir(k8sParentDir)
 	defer removeDir(testParentDir)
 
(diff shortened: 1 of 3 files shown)


msau42:
We lost the summary of each test case in the output: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/349/pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration/1151641109189890048/

test/k8s-integration/main.go
```diff
@@ -306,48 +299,38 @@ func setEnvProject(project string) error {
...
+func runTestsWithConfig(k8sDir, testFocus, testConfigArg string) error {
+	err := os.Chdir(k8sDir)
 	if err != nil {
 		return err
 	}
 
 	homeDir, _ := os.LookupEnv("HOME")
 	os.Setenv("KUBECONFIG", filepath.Join(homeDir, ".kube/config"))
 
-	artifactsDir, _ := os.LookupEnv("ARTIFACTS")
```
msau42: We still need reportArg

PR #352, opened 2019-07-22
build kubectl for test kube directory

hantaowang:
**What type of PR is this?**
/kind failing-test

Builds k8s testing src with make quick release so that kubetest has access to the kubectl built.

```release-note
None
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -214,7 +214,7 @@ func handle() error {
 		}
 		err = buildKubernetes(testDir, "WHAT=test/e2e/e2e.test")
 		if err != nil {
-			return fmt.Errorf("failed to build Kubernetes: %v", err)
+			return fmt.Errorf("failed to build Kubernetes e2e: %v", err)
 		}
 	} else {
 		testDir = k8sDir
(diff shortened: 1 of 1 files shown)


test/k8s-integration/main.go
```diff
@@ -212,7 +212,7 @@ func handle() error {
 		if err != nil {
 			return fmt.Errorf("failed to download Kubernetes source: %v", err)
 		}
-		err = buildKubernetes(testDir, "WHAT=test/e2e/e2e.test")
+		err = buildKubernetes(testDir, "quick-release")
```
davidz627: can you add a "TODO" to parallelize these builds since they'll take a long time now

hantaowang: Thinking about this, it may not make sense to parallelize k8s builds because each quick-release is already parallelized you can check htop to confirm). You might save some time on the go builds for certain components that are not, but then suffer from the context switching. 

This is also why my desktop was dying earlier - I was parallelizing two builds.

hantaowang: I changed the PR so it builds kubectl specificlly (I figured out how). Currently testing to see if this solves it.

davidz627: ping me when ready for review

hantaowang:
/close
Use https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/358 instead

Issue #356, opened 2019-07-26
Set pod CPU & mem Requests for driver Container & Node components


davidz627:
I did some stress testing of the driver:

![GKE Container - CPU usage for Controller](https://user-images.githubusercontent.com/5459214/61985942-ba8e9980-afc0-11e9-808a-dfb7c7848323.png)

![GKE Container - Memory usage for Controller](https://user-images.githubusercontent.com/5459214/61985945-be222080-afc0-11e9-811e-bec5a9525fe9.png)

![GKE Container - CPU usage for NODE](https://user-images.githubusercontent.com/5459214/61985953-c2e6d480-afc0-11e9-834b-c50b508ec45c.png)

![GKE Container - Memory usage for NODE](https://user-images.githubusercontent.com/5459214/61985955-c5e1c500-afc0-11e9-93d3-b8888aa191d7.png)

Timeline:
3:35: I create 50 pods on one node
3:45: Delete 50 pods on one node
3:53 Start 50 pods again
3:59 Delete 50 pods again
4:05 Start 100 pods on one node
4:15 Delete 100 pods

Configuration used:
- GKE cluster with 1 n1-standard-32 node
- Stable driver v0.5.1
- Disks are 6Gi

Limitation of this test:
- Only used one node so probably not the max stress possible on the controller. Looks like we didn't hit the limit in terms of controller ops with the max pods per node because the CPU was still scaling linearly.
- Node backoff is probably not optimized so we probably didn't hit the max stress possible on the node (however, with the current version of the driver/csi-sidecars with its current backoff settings I believe we hit the max stress)
- 6Gi disks probably smaller than many production disks, larger disks may take longer/more cpu/mem to format

/cc @verult @msau42

misterikkit:
Looks like we could comfortably set the Node CPU request at `40m` or `60m`. Setting a CPU limit is not recommended.

msau42:
It would be interesting to see if go has profiling tools to get a breakdown of memory usage.

hantaowang:
cc @jingxu97

misterikkit:
Based on discussion, it would be interesting to,
- run the Controller on a separate node from the load test
- get some CPU & memory profiles from the Controller & Node components

davidz627:
## Follow-Up No.1
- Tested with same configuration as original but set disk size to "500Gi"

[...]

PR #366, opened 2019-08-06
Set requests for CPU and MEM for controller containers

davidz627:
As it turns out CPU and MEM requests are container based, so I took the MAX values of each for each container throughout my test runs and set those as requests. Not sure if I should do average or something else?

Based on results from #356 

/assign @msau42 @misterikkit
/cc @verult 

/kind feature
/kind flake

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -25,6 +25,10 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+          resources:
+            requests:
+              cpu: "25m"
+              memory: "25Mi"
         - name: csi-attacher
           image: gke.gcr.io/csi-attacher
           args:
@@ -33,6 +37,10 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+          resources:
+            requests:
+              cpu: "25m"
+              memory: "25Mi"
         - name: csi-resizer
           image: gke.gcr.io/csi-resizer
           args:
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/base/controller.yaml
```diff
@@ -24,6 +24,10 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+          resources:
+            requests:
+              cpu: "25m"
+              memory: "23Mi"
```
msau42: Was there some best practice about not setting memory requests?  Or was that for limits?

davidz627: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/356#issuecomment-515633418

CPU limit not recommended AFAIK. I think requests seem fine, don't think it will affect throttling just pre-emptability and resource packing.

davidz627:
/hold
we should not set based on max, instead should be based on some reasonable amount above baseline

verult:
Is this PR blocked on experiments in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/356?

davidz627:
/hold cancel
Since this is a system critical pod and the max CPU usage is actually not very high we've set the CPU Request to be around the max usage for each container `25m` and the controller memory usage for each container is actually fairly steady around `25MiB` so each container has been set to this.

Also the experiments were performed with 100 concurrent volumes but since the controller is per-cluster there could easily be this many or more so it's better to err on the higher side for requests.

This puts the entire controller Pod at `100m` CPU and `100MiB` memory
/cc @msau42 @verult 
Please LGTM if this seems ok.

msau42:
We are really tight on controller cpu. I'd rather request a lower number for "avg" use instead of max/stress, and allow things to burst, rather than reserve too much cpu that may largely stay idle.

Issue #367, opened 2019-08-06
Make sure all errors returned in CSI response is wrapped


msau42:
For example, [this](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-pd-csi-driver/controller.go#L383) directly returns cloud provider error.  And for this particular case,we should be returning NotFound according to the CSI Spec (pending discussion in https://github.com/container-storage-interface/spec/issues/373)

/kind bug

PR #368, opened 2019-08-06
Wrap all GRPC errors in status, fix semantics of NotFound errors

davidz627:
Fixes: #367 
/assign @msau42 @jsafrane 
/cc @jingxu97 

/kind bug
/kind cleanup

```release-note
ControllerUnpublishVolume now returns success when the Node is GCE API NotFound.
Invalid format VolumeID is now GRPC InvalidArgument error instead of GRPC NotFound.
Underspecified disks not found in any zone now return GRPC NotFound.
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -232,7 +232,7 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 
 	volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, volKey)
 	if err != nil {
-		klog.Warningf("Treating volume as deleted because cannot find volume %v: %v", volKey.String(), err)
+		klog.Warningf("Treating volume as deleted because cannot find volume %v: %v", volumeID, err)
 		return &csi.DeleteVolumeResponse{}, nil
 	}
 
@@ -546,6 +546,15 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, volumeID)
 	}
 	defer gceCS.volumeLocks.Release(volumeID)
+
+	// Check if volume exists
+	_, err = gceCS.CloudProvider.GetDisk(ctx, volKey)
+	if err != nil {
+		if gce.IsGCENotFoundError(err) {
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("CreateSnapshot could not find disk %v: %v", volKey.String(), err))
+		}
+		return nil, status.Error(codes.Internal, fmt.Sprintf("CreateSnapshot unknown get disk error: %v", err))
+	}
 
 	// Check if snapshot already exists
 	var snapshot *compute.Snapshot
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -226,14 +226,17 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 	if err != nil {
 		// Cannot find volume associated with this ID because VolumeID is not in
 		// correct format, this is a success according to the Spec
-		klog.Warningf("Treating volume as deleted because volume id %s is invalid: %v", volumeID, err)
+		klog.Warningf("DeleteVolume treating volume as deleted because volume id %s is invalid: %v", volumeID, err)
 		return &csi.DeleteVolumeResponse{}, nil
 	}
 
 	volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, volKey)
 	if err != nil {
-		klog.Warningf("Treating volume as deleted because cannot find volume %v: %v", volumeID, err)
-		return &csi.DeleteVolumeResponse{}, nil
+		if gce.IsGCENotFoundError(err) {
+			klog.Warningf("DeleteVolume treating volume as deleted because cannot find volume %v: %v", volumeID, err)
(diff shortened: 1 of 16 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -267,7 +267,7 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 
 	volKey, err := common.VolumeIDToKey(volumeID)
 	if err != nil {
-		return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find volume with ID %v: %v", volumeID, err))
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("ControllerPublishVolume volume ID is invalid: %v", err))
 	}
 
 	volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, volKey)
```
msau42: Do we want to check gce error codes here? What if we had temporary issues with the cloud provider, and not that the disk doesn't actually exist.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -382,7 +382,11 @@ func (gceCS *GCEControllerServer) ControllerUnpublishVolume(ctx context.Context,
 	}
 	instance, err := gceCS.CloudProvider.GetInstanceOrError(ctx, instanceZone, instanceName)
 	if err != nil {
-		return nil, err
+		if gce.IsGCENotFoundError(err) {
+			// Node not existing on GCE means that disk has been detached
```
msau42: do we want to log this?

davidz627:
/hold
using a private branch version of csi-sanity based on https://github.com/kubernetes-csi/csi-test/issues/103.
Wait till above PR merged and pull in new real CSI-Test dependency

pkg/gce-pd-csi-driver/controller.go
davidz627: there's no real way to distinguish between "disk not exists in zone because it may or may not exist in another zone but is temporarily not showing up" vs "disk does not exist in any zone because it just doesn't exist" since `RepairUnderspecifiedVolumeKey` queries all the zones in the region for the disk

msau42: you could check if the zone check fails because of "not found" error code

davidz627: right, but it may be `NotFound` is some subset of zones, and some other error in other zones. What do we do in that case?

msau42: If we looped through all zones and found it, then return success.  Otherwise, if all the zones returned notFound, then return notFound.  Otherwise return error?

jingxu97: I noticed that disk name is not unique across zones. How we can be sure the disk we found while looping through the zones is exactly the one we want since we don't pass zone information?

davidz627: we can't be sure, but I believe this is consistent with behavior of in-tree plugin. This codepath is only hit for migration - any disks managed natively through the CSI Driver have zone/region information encoded in their volume ID

msau42: To be on the safe side, can we search all zones and if we have multiple matches than return error?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -382,7 +384,12 @@ func (gceCS *GCEControllerServer) ControllerUnpublishVolume(ctx context.Context,
...
+			klog.Warningf("Treating volume %v as unpublished because node %v could not be found", volKey.String(), instanceName)
+			return &csi.ControllerUnpublishVolumeResponse{}, nil
```
msau42: Discussed this with @jsafrane this morning.  Current CSI spec clearly says to return NotFound error in this case.  Until CSI spec changes, we should probably follow it.

msau42: Sorry I'm behind the times. This was discussed in the CSI spec meeting this morning and conclusion is we should allow plugins to return success if it knows volume is definitely detached.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -224,14 +224,16 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 
 	volKey, err := common.VolumeIDToKey(volumeID)
 	if err != nil {
-		// Cannot find volume associated with this ID because can't even get the name or zone
-		// This is a success according to the spec
+		// Cannot find volume associated with this ID because VolumeID is not in
+		// correct format, this is a success according to the Spec
+		klog.Warningf("Treating volume as deleted because volume id %s is invalid: %v", volumeID, err)
```
msau42: Should it return success? Can we get into a case where user pre-provisions a PV, but gets the volume id wrong.  Then when they go and delete it they think it's successful, when it's actually not, and then we leak a volume.

davidz627: I don't agree with the return code in this case for many reasons, but this is what it says in the Spec as well as CSI Sanity.

pkg/gce-pd-csi-driver/controller.go
jingxu97: In this issue https://github.com/kubernetes/kubernetes/pull/65198, @verult mentioned that csi driver solve this issue, how it is solved?

davidz627: When using the CSI Driver natively the region/zone information is encoded in the volume ID. This unspecified/repair case is **only** for CSI Migration, in which case we will continue to have the same issue as before.

davidz627: alright i've done the original repair suggestion in a seperate commit, PTAL

davidz627:
/hold cancel
/retest
@msau42 ready for review

PR #358, opened 2019-07-29
add functionality to use gke with kubetest

hantaowang:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**:
Our new testing framework kubetest does not work out of the box with GKE.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:
/assign @davidz627 @jingxu97 

**Does this PR introduce a user-facing change?**:

```release-note
None
```

PR #370, opened 2019-08-10
Export kubectl path to test

hantaowang:
**What type of PR is this?**
/kind failing-test

We previous had a PR to export the kubectl path https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/352 that was closed in favor of another PR https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/358, but oddly the kubectl path fix was never actually included in that second PR. This adds the fix.

Fixes #

```release-note
None
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -214,12 +214,16 @@ func handle() error {
 		}
 		err = buildKubernetes(testDir, "WHAT=test/e2e/e2e.test")
 		if err != nil {
-			return fmt.Errorf("failed to build Kubernetes: %v", err)
+			return fmt.Errorf("failed to build Kubernetes e2e: %v", err)
 		}
-		// kubetest relies on ginkgo already built in the test k8s directory
+		// kubetest relies on ginkgo and kubectl already built in the test k8s directory
 		err = buildKubernetes(testDir, "ginkgo")
 		if err != nil {
 			return fmt.Errorf("failed to build gingko: %v", err)
+		}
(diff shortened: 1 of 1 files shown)


test/k8s-integration/cluster.go
```diff
@@ -193,5 +194,14 @@ func getGKEKubeTestArgs() ([]string, error) {
 		fmt.Sprintf("--gcp-project=%s", project[:len(project)-1]),
 	}
 
+	out, err := exec.Command("which", "kubectl").Output()
```
msau42: where does kubectl come from? The test container image binary?  kubectl has interesting compatibility behavior with Kubernetes versions.  The kubectl version from the container image may get stale.  How do our regular gke tests get kubectl?

hantaowang: It comes from the test image. There really isnt a way to make sure kubectl version is compatible with GKE version unless we build the exact version of kubectl for our GKE cluster version. We can get close by using kubectl built in the testing directory, so at least as long as were using the latest k8s e2e tests then kubectl version >= gke version. 

I updated the code to do this.

msau42: Is this what normal gke tests too? We should follow closely what is done for the regular tests.

hantaowang: Do you mean other tests not owned by sig storage that run on GKE? They do t try to boot their own gke cluster. They let kubetest do the cluster up. And thus they do t have to worry about kubectl as kubetest will manage that for them.

PR #378, opened 2019-08-16
Update csi-attacher to v.1.2.1 for fix to not ignore ControllerUnpublishErrors

davidz627:
/kind bug

/assign @msau42 

Hope this might actually fix some flakes in the tests :)

```release-note
Updated CSI Attacher to v1.2.1 to stop ignoring errors from ControllerUnpublish
```


msau42:
/lgtm

Do we want to update our other overlays as well?

davidz627:
> Do we want to update our other overlays as well?

AFAIK no other overlays need to be updated. Maybe release staging, but not a priority rn.

PR #364, opened 2019-08-02
Deployment spec cleanup

verult:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
* Deployment is now in a standalone namespace rather than `default` to isolate the driver from arbitrary workloads.
* Removed unnecessary driver-registrar role. The events permission is unused
* Bumping snapshotter version to latest
* Changed resizer image to gcr.io
* Added "gce-pd-csi-driver-node" and "gce-pd-csi-driver-controller" priorityClasses and set them in node DaemonSet and controller StatefulSet.
* Changed the label key to "k8s-app" to follow the GKE convention that this deployment is a system app.
* Fixed the name of the resizer RoleBinding

Fixes: #321
Fixes: #37 
Fixes: #357

```release-note
Deployment spec updates:
* The deployment is no longer in namespace `default`
* Changed "app" label key to "k8s-app"
* csi-snapshotter version has been changed to v1.2.0-gke.0
* The resizer role binding has been renamed to "csi-controller-resizer-binding"
* Removed driver-registrar role. 
```

/assign @davidz627 @msau42

PR #379, opened 2019-08-20
Fix resize test failure; enable resize tests against latest

verult:
**What type of PR is this?**
>
/kind bug

Fixes [resize tests](https://testgrid.k8s.io/sig-gcp-compute-persistent-disk-csi-driver#Migration%20Kubernetes%20Master%20Driver%20Latest) that was broken by https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/364
* Replaced snapshotter and resizer RBAC namespace

Enabling resize tests against the latest driver in Testgrid:
* Added resize capabilities to test config
* Enabled `ExpandCSIVolumes` feature gate for test cluster creation.

Fixes #

Locally verified with k8s master, with `ExpandCSIVolumes` feature gate enabled. Also manually verified resize operation with dev deployment.

```release-note
NONE
```


test/run-k8s-integration.sh
```diff
@@ -23,6 +23,7 @@ export GCE_PD_VERBOSITY=9
 make -C ${PKGDIR} test-k8s-integration
 
 base_cmd="${PKGDIR}/bin/k8s-integration-test \
+            --kube-feature-gates=\"ExpandCSIVolumes=true\"
```
davidz627: can you add this to the local script too (or a comment saying to bring up your cluster with this feature gate or it wont work if the script doesn't bring up a cluster)

verult: I'm actually about to remove this, since we don't want resize test in latest yet

PR #397, opened 2019-10-09
Bumping sidecars to latest versions for 0.6.0 staging RC

verult:
```release-note
NONE
```

/assign @davidz627 @msau42


deploy/kubernetes/overlays/prow-gke-release-staging-rc/kustomization.yaml
```diff
@@ -8,10 +8,10 @@ images:
   newTag: "v0.6.0-rc1"
```
msau42: do we need a new pd driver rc too? @davidz627

verult: Updated

Issue #388, opened 2019-09-19
Add documentation on how to create a PV with a pre-existing PD


msau42:
/kind docs

PR #403, opened 2019-10-14
Add user guide for pre-provisioned volumes

davidz627:
Fixes: #388 

/assign @msau42 
/kind documentation

```release-note
NONE
```

### DIFF
--- a/docs/kubernetes/user-guides/pre-provisioned.md
+++ b/docs/kubernetes/user-guides/pre-provisioned.md
@@ -1,0 +1,90 @@
+# Kubernetes Pre-Provisioned Disks User Guide
+
+This guide gives a simple example on how to use this driver with disks that have
+been pre-provisioned by an administrator.
+
+## Install Driver
+
+See [instructions](driver-install.md)
+
+## Pre-Provision a Disk
+
+If you have not already pre-provisioned a disk on GCP you can do that now.
+
+1. Create example PD
+
+```bash
+gcloud compute disks create test-disk --zone=us-central1-c --size=200Gi
+```
+
+## Create Persistent Volume for Disk
+
+1. Create example Storage Class
+
+```bash
+kubectl apply -f ./examples/kubernetes/demo-zonal-sc.yaml
+```
+
+2. Create example Persistent Volume
+
+**Note:** The `volumeHandle` and `nodeSelectorTerms` values should be generated
+based on the project, zone\[s\], and PD name of the disk created. `storage` value
+should be generated based on the size of the underlying disk
+
+**Regional PD Note:** The volume handle format is different and the
+`nodeSelectorTerms` must contain both zones the PD is supported in.
+
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: my-pv
+  annotations:
+    pv.kubernetes.io/provisioned-by: pd.csi.storage.gke.io
+spec:
+  storageClassName: "csi-gce-pd"
+  capacity:
+    storage: 200Gi # MODIFY THIS LINE
+  accessModes:
+    - ReadWriteOnce
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: topology.gke.io/zone
+          operator: In
+          values:
+          - us-central1-c # MODIFY THIS LINE
+        # - us-central1-b <For Regional PD>
+  csi:
+    driver: "pd.csi.storage.gke.io"
+  # volumeHandle: "projects/${PROJECT}/regions/us-central1/disks/test-disk" <For Regional PD>
+    volumeHandle: "projects/${PROJECT}/zones/us-central1-c/disks/test-disk" # MODIFY THIS LINE
+```
+
+Then `kubectl apply` this YAML.
+
+## Use Persistent Volume In Pod
+
+1. Create example PVC and Pod
+
+```bash
+kubectl apply -f ./examples/kubernetes/demo-pod.yaml
+```
+
+2. Verify PV is created and bound to PVC
+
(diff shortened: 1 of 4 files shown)


docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Create Persistent Volume for Disk
+
+1. Create example Zonal Storage Class
+```
+$ kubectl apply -f ./examples/kubernetes/demo-zonal-sc.yaml
+```
+
+2. Change PROJECT variable to your project and create example Persistent Volume
```
msau42: Maybe emphasize here that you need to generate the volumeHandle based off the zone and PD name of the disk you created.

docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,63 @@
...
+```
+$ cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: PersistentVolume
+metadata:
```
msau42: There's actually one outstanding issue where you need to add the "provisioned-by" annotation to the PV in order for the CSI driver to handle deletion of the volume. https://github.com/kubernetes-csi/external-provisioner/issues/202

docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,63 @@
+# Kubernetes Pre-Provisioned Disks User Guide
```
msau42: Do we have some Table of Contents page that can link to all these user-guides?

docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,63 @@
...
+This guide gives a simple example on how to use this driver with disks that have been pre-provisioned by an administrator.
+
+## Install Driver
+
+See [instructions](driver-install.md)
+
+## Pre-Provision a Disk
```
msau42: Example for RePD?

docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+2. Change PROJECT variable to your project and create example Persistent Volume
+```
+$ cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: my-pv
+spec:
```
msau42: Do we also need to set PV.NodeAffinity manually too since we don't have a webhook to inject it?

msau42:
/assign @verult 
to also take a look

docs/kubernetes/user-guides/pre-provisioned.md
davidz627: There's a link to the "user guides" folder from the main README

docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,86 @@
...
+```bash
+kubectl apply -f ./examples/kubernetes/demo-zonal-sc.yaml
+```
+
+2. Create example Persistent Volume
+
+**Note:** The `volumeHandle` and `nodeSelectorTerms` values should be generated
+based on the project, zone, and PD name of the disk created.
+
+**RePD Note:** For RePD the volume handle is of the form:
```
msau42: Maybe have explicit examples for zonal vs repd?

davidz627: I would have to write a very similar note anyway about modifying these specific parts of the spec vs. right now which is modify + add one little thing. I think this is clear enough and reduces duplication

docs/kubernetes/user-guides/pre-provisioned.md
```diff
@@ -0,0 +1,90 @@
...
+
+2. Create example Persistent Volume
+
+**Note:** The `volumeHandle` and `nodeSelectorTerms` values should be generated
+based on the project, zone, and PD name of the disk created. `storage` value
+should be generated based on the size of the underlying disk
+
+**RePD Note:** For RePD the volume handle is of the form:
+`"projects/${PROJECT}/regions/us-central1/disks/test-disk"` and you must add
+another `nodeSelectorTerms` to match the zones of the RePD.
```
msau42: There's actually 2 ways to do it. IMO the simpler way is to add one more zone to the "values" array.  Would it be more clear if the example looked something like:
```
values:
- us-central1-c
- <2nd zone if Regional PD>
```

docs/kubernetes/user-guides/pre-provisioned.md
msau42: Would it be easier to spot the difference if you had the 2 formats side by side in the spec?

docs/kubernetes/user-guides/pre-provisioned.md
davidz627: ack - did not realize this was the case. I just modeled it after what it looked like after dynamic provisioning.

docs/kubernetes/user-guides/pre-provisioned.md
davidz627: not sure what you mean by "side by side" in this case. I do feel like it could be clearer but I'm not sure that duplicating everything for RePD is the solution. It feels a little clumsy (especially in terms of keeping in sync when things change).

msau42: Like maybe embed the RePD syntax as a comment in the spec above the zonal syntax line? Kind of like how the 2nd zone is also embedded into the spec too.

Also we should stick to the official "Regional PD" name.

Issue #213, opened 2018-12-20
Switch RePD to use v1 API


msau42:
It's still using the v1beta api

PR #404, opened 2019-10-14
Move all RePD compute API calls to v1 from v1beta

davidz627:
Could probably refactor out `CloudDisk` as well as that was an artifact of having different objects of disks from different APIs. We could probably also consolidate a lot of the zonal/regional codepaths now that we don't need to use entirely different compute APIs.

However, I've refrained from making large changes here in the interest of stability. Also I think theres a possibility we'd need to use different APIs again in the future so I've left the option for that open.

Fixes: #213 

/kind cleanup
/assign @msau42 
```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/cloud-disk.go
+++ b/pkg/gce-cloud-provider/compute/cloud-disk.go
@@ -15,13 +15,12 @@ limitations under the License.
 package gcecloudprovider
 
 import (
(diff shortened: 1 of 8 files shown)


msau42:
/lgtm

Now that we have gke regional cluster support, we could potentially add another storageclass test config to use Regional PDs in our external e2e tests.

PR #405, opened 2019-10-14
Prefix object names to disambiguate with other CSI drivers

### LINKED ISSUE #373: Unique object names in deployment
Some of the objects we create in the deployment manifests are cluster-wide.  We should probably have more unique names for them.  For example, our service account has a name like "csi-controller-sa" which when looking at it from the outside, you would have no idea it was for the pd csi driver (if you have multiple csi drivers on your system).

verult:
**What type of PR is this?**
/kind cleanup

Fixes #373

Prefix follows the convention established by controller and node deployment names.

```release-note
BREAKING CHANGE: All deployment objects in setup-cluster.yaml have been renamed. When deleting the deployment using ./delete-driver.sh, make sure to use specs from your previous deployment version to ensure the correct objects are cleaned up.
```

### DIFF
--- a/deploy/kubernetes/base/controller.yaml
+++ b/deploy/kubernetes/base/controller.yaml
@@ -13,8 +13,8 @@ spec:
       labels:
         app: gcp-compute-persistent-disk-csi-driver
     spec:
-      serviceAccountName: csi-controller-sa
-      priorityClassName: gce-pd-csi-driver-controller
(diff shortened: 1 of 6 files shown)


verult:
Stable deployment verified manually. Dev deployment verification in progress

davidz627:
/approve
did you resolve the resize test thing?

verult:
Yup I used a 1.16 cluster instead and expanding the PV size worked great

Issue #386, opened 2019-09-18
Add volume capacity metrics


davidz627:
like from `df` returned in `NodeGetVolumeStats`

PR #406, opened 2019-10-14
Add NodeGetVolumeStats Capability/Implementation/Tests

davidz627:
Fixes: #386 

/kind cleanup
/kind feature

/assign @msau42 

```release-note
Add GET_VOLUME_STATS Node Service Capability and implementation for getting stats for volume
```


pkg/gce-pd-csi-driver/node.go
```diff
@@ -364,7 +368,42 @@ func (ns *GCENodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRe
...
+
+	exists, err := ns.Mounter.Interface.ExistsPath(req.VolumePath)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "unknown error when stat on %s: %v", req.VolumePath, err)
+	}
+	if !exists {
+		return nil, status.Errorf(codes.NotFound, "path %s does not exist", req.VolumePath)
+	}
+
+	available, capacity, used, inodesFree, inodes, inodesUsed, err := ns.VolumeStatter.StatFS(req.VolumePath)
```
msau42: what happens if this is a raw block volume?

davidz627: glad you asked. I got some weird behavior in my test:
```
"got: available 1930592256, capacity 1930592256, used 0, inodesFree 471051, inodes 471336, inodesUsed 285 -- expected: available ~= 5368709120, capacity ~= 5368709120, used = 0, inodesFree != 0, inodes != 0 , inodesUsed != 0"
```

It seems the available and capacity bytes are much smaller than expected. Also this might be just my misunderstanding but shouldn't there be `0` inodes total for an unformatted block device?

Looking into it some more but posting here just in case you have thoughts.

davidz627: ok I think these numbers are just random garbage numbers since statfs specifically gets filesystem statistics. So for a block device we would probably just want to do a different call..

Though FS/Block isn't passed into this function... we could inspect it based on the filename of the `volumePath` but that seems hacky

msau42: cc @gnufied 

Seems like we may be missing block/fs capability in this CSI call too?  Or do we need to detect what type it is based on the path?

msau42: These methods may be useful for you to detect if a path is a block device, along with getting the capacity of a block device:
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/pkg/util/volume_util_linux.go

davidz627: updated to contain logic for block as well

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -539,4 +496,110 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			Expect(gce.IsGCEError(err, "notFound")).To(BeTrue(), "Expected disk to not be found")
+		}()
+
+		verifyVolumeStats := func(a verifyArgs) error {
+			available, capacity, used, inodesFree, inodes, inodesUsed, err := client.NodeGetVolumeStats(volID, a.publishDir)
+			if err != nil {
+				return fmt.Errorf("failed to get node volume stats: %v", err)
+			}
+			if !equalWithinEpsilon(available, common.GbToBytes(defaultSizeGb), defaultEpsiolon) || !equalWithinEpsilon(capacity, common.GbToBytes(defaultSizeGb), defaultEpsiolon) || !equalWithinEpsilon(used, 0, defaultEpsiolon) ||
+				inodesFree == 0 || inodes == 0 || inodesUsed == 0 {
```
msau42: Should we write some files in the test so that inodes and used are non zero? (could be another test case)

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -539,4 +496,110 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			_, err = computeService.Disks.Get(p, z, volName).Do()
+			Expect(gce.IsGCEError(err, "notFound")).To(BeTrue(), "Expected disk to not be found")
+		}()
+
+		verifyVolumeStats := func(a verifyArgs) error {
+			available, capacity, used, inodesFree, inodes, inodesUsed, err := client.NodeGetVolumeStats(volID, a.publishDir)
+			if err != nil {
+				return fmt.Errorf("failed to get node volume stats: %v", err)
+			}
+			if !equalWithinEpsilon(available, common.GbToBytes(defaultSizeGb), defaultEpsiolon) || !equalWithinEpsilon(capacity, common.GbToBytes(defaultSizeGb), defaultEpsiolon) || !equalWithinEpsilon(used, 0, defaultEpsiolon) ||
```
msau42: typo: epsilon?

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -539,4 +496,110 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			}
+			return nil
+		}
+
+		// Attach Disk
+		err := testLifecycleWithVerify(volID, volName, instance, client, false /* readOnly */, true /* block */, verifyVolumeStats, nil)
+		Expect(err).To(BeNil(), "Failed to go through volume lifecycle")
+	})
+
+	It("Should get correct VolumeStats", func() {
```
msau42: Another test case will be to expand the volume (block and fs) and check volumestats

test/e2e/tests/single_zone_e2e_test.go
msau42: Also inodesFree is 0?

davidz627: dumb question - are there inodes in a raw block volume..?

test/e2e/tests/single_zone_e2e_test.go
davidz627: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/410

davidz627: dont want to make this PR even bigger right now

test/e2e/tests/single_zone_e2e_test.go
msau42: Inodes are a fs concept. This is the fs test case right?

pkg/gce-pd-csi-driver/node.go
gnufied: We could add capability here too but for now we decided not to. Depending on use case we could do this as a future update.

Also btw - used and available should not set when device is of type raw block.

test/e2e/tests/single_zone_e2e_test.go
davidz627: ah yes, this test is checking `if inodesFree==0 then Error`

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -53,13 +53,76 @@ func getCustomTestGCEDriver(t *testing.T, mounter *mount.SafeFormatAndMount, dev
...
+	}
+
+	testCases := []struct {
+		name       string
+		volumeID   string
+		volumePath string
+		expectErr  bool
+	}{
+		{
+			name:       "normal",
```
msau42: do we also want a unit test for fs vs block?

davidz627: covered in e2e test

Issue #309, opened 2019-06-20
Make sure we log successful operations in all paths


davidz627:
Maybe we can do this at the layer above the actual RPC methods. We should always log on RPC exit

PR #409, opened 2019-10-15
Clean up all logging for the driver, log successes in all cases

davidz627:
/kind cleanup
/assign @msau42 

Fixes: #309 

```release-note
Improve driver logs to log success in all paths as well as logging additional useful information
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -76,7 +76,7 @@ func CreateCloudProvider(vendorVersion string, configPath string) (*CloudProvide
 	// At this point configFile could still be nil.
 	// Any following code that uses configFile should handle nil pointer gracefully.
 
-	klog.V(1).Infof("Using GCE provider config %+v", configFile)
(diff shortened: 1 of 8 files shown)


pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -122,11 +122,11 @@ func generateTokenSource(configFile *ConfigFile) (oauth2.TokenSource, error) {
 
 	// DefaultTokenSource relies on GOOGLE_APPLICATION_CREDENTIALS env var being set.
 	if gac, ok := os.LookupEnv("GOOGLE_APPLICATION_CREDENTIALS"); ok {
-		klog.V(4).Infof("GOOGLE_APPLICATION_CREDENTIALS env var set %v", gac)
+		klog.V(2).Infof("GOOGLE_APPLICATION_CREDENTIALS env var set %v", gac)
```
msau42: this doesn't actually log the token right?

davidz627: it logs the location of the token

PR #411, opened 2019-10-21
Make driver test capabilities configurable and remove expansion from GKE tests

davidz627:
/kind failing-test

/assign @msau42 @misterikkit

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/driver-config.go
+++ b/test/k8s-integration/driver-config.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"path/filepath"
 	"text/template"
@@ -9,6 +10,7 @@ import (
 
 type driverConfig struct {
 	StorageClassFile string
+	Capabilities     []string
 }
 
 const (
@@ -19,7 +21,7 @@ const (
 
 // generateDriverConfigFile loads a testdriver config template and creates a file
 // with the test-specific configuration
-func generateDriverConfigFile(pkgDir, storageClassFile string) (string, error) {
+func generateDriverConfigFile(pkgDir, storageClassFile, deploymentStrat string) (string, error) {
 	// Load template
 	t, err := template.ParseFiles(filepath.Join(pkgDir, testConfigDir, configTemplateFile))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


test/k8s-integration/driver-config.go
```diff
@@ -38,10 +40,41 @@ func generateDriverConfigFile(pkgDir, storageClassFile string) (string, error) {
...
 
 	// Fill in template parameters
+	caps := []string{
+		"persistence",
+		"block",
+		"fsGroup",
+		"exec",
+		"multipods",
+		"topology",
+	}
```
misterikkit: Is there a well known source of truth for all valid capability values? Is it part of the CSI capability spec?

test/k8s-integration/driver-config.go
```diff
@@ -38,10 +40,41 @@ func generateDriverConfigFile(pkgDir, storageClassFile string) (string, error) {
...
 
+	/* Unsupported Capabilities:
+	   snapshotDataSource
+	   pvcDataSource
+	   RWX
+	   volumeLimits # PD Supports volume limits but test is very slow
+	   singleNodeVolume
+	   dataSource
+	*/
+
```
misterikkit: :+1:

test/k8s-integration/driver-config_test.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import "testing"
+
+// WARNING: Not an official test - run this as a quick n dirty way to test your
+// template generation code :)
+func TestGenerateDriverConfigFile(t *testing.T) {
+	_, err := generateDriverConfigFile("/usr/local/google/home/dyzz/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver", "sc-standard.yaml", "gke")
```
misterikkit: Did you want to commit this file as-is?

I am in favor of a test that renders the template and then verifies it parses as valid yaml...

test/k8s-integration/driver-config.go
davidz627: They're here - definitely not easy to find.
https://github.com/kubernetes/kubernetes/blob/23320c958edc85feb541db85edc43ec629288a7d/test/e2e/storage/testsuites/testdriver.go#L136

test/k8s-integration/driver-config_test.go
davidz627: I'm not convinced we need a test for the test. This was just for my personal iteration speed. I would rather remove this file than spend more time on it

test/k8s-integration/driver-config.go
msau42: Add a comment where to find the capabilities from?

test/k8s-integration/driver-config.go
```diff
@@ -38,10 +40,41 @@ func generateDriverConfigFile(pkgDir, storageClassFile string) (string, error) {
...
+		"fsGroup",
+		"exec",
+		"multipods",
+		"topology",
+	}
+
+	switch deploymentStrat {
+	case "gke":
+	case "gce":
+		// TODO: OSS K8S supports volume expansion for CSI by default in 1.16+;
```
msau42: Should we consider K8s version specific capability setting? ie if we're running on a 1.16+ cluster, then enable resizing tests.

test/k8s-integration/driver-config_test.go
misterikkit: :+1:

test/k8s-integration/driver-config.go
misterikkit: Do we need it today? Should we try adding it later?

davidz627:
@msau42 @misterikkit updated

test/k8s-integration/driver-config.go
msau42: We can do it later but it means that we will be lacking test coverage of these new features we're releasing until GKE can catch up.

misterikkit: Is there a subset of coverage we can get on GKE before k8s 1.16 is available?

PR #415, opened 2019-10-23
Add support for multi-writer PD

sschmitt:
/kind feature

/cc @msau42

Adds multi writer support (currently alpha in GCP).

There's a bit of code duplication due to alpha copies of certain methods. I found it challenging to reduce the duplication but I'd be open to suggestions on how to refactor. On the other hand it might just be a temporary situation until the API's move to beta / GA.

```release-note
Add support for multi-writer raw block devices.
```


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -288,6 +297,66 @@ func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta
 	return nil
 }
 
+func (cloud *CloudProvider) insertRegionalAlphaDisk(ctx context.Context, volKey *meta.Key, diskType string, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID, diskEncryptionKmsKey string, multiWriter bool) error {
+	diskToCreateAlpha := &computealpha.Disk{
```
davidz627: Thanks for doing this, It looks pretty good as-is but I think there's also a way to de-dupe some of this copy-pasted code.

Assuming zonal here:
1) Have one `insertDisk` that takes in all the parameters
2) Create a v1 disk with logic up to line `318` here.
3) Make `insertOp` a `interface{}` type (this part is kind of nasty, suggestions welcome)
3) ```
if multiWriter { 
  convertV1DiskToV1AlphaDisk(disk) // you have to write this
  insertOp = cloud.alphaService.Insert(disk)
} else{
  insertOp = cloud.service.Insert(dis)
}
```
Then the rest of the error handling logic is shared
4) typeAssert on `insertOp` and call the respective `waitForOp` on it
5) Share the other logic

This would dedupe most of the code and only have branching for the actual GCE API Insert call and the WaitForOp.
Let me know what you think. Happy to discuss more

File: pkg/gce-cloud-provider/compute/gce-compute.go [hunk 2]
```diff
@@ -288,6 +297,66 @@ func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta
...
+	if diskEncryptionKmsKey != "" {
+		diskToCreateAlpha.DiskEncryptionKey = &computealpha.CustomerEncryptionKey{
+			KmsKeyName: diskEncryptionKmsKey,
+		}
+	}
+
+	insertOp, err := cloud.alphaService.RegionDisks.Insert(cloud.project, volKey.Region, diskToCreateAlpha).Context(ctx).Do()
+	if err != nil {
+		if IsGCEError(err, "alreadyExists") {
+			disk, err := cloud.GetDisk(ctx, volKey)
```
davidz627: does `GetDisk` need to be changed as well? If we are to get a disk that is multi-writer what happens (since it is only on the alpha struct)?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -288,6 +297,66 @@ func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta
...
+			if err != nil {
+				return err
+			}
+			err = cloud.ValidateExistingDisk(ctx, disk, diskType,
```
davidz627: validation might also need to be changed to check if `multiWriter` field is equal

pkg/gce-cloud-provider/compute/gce-compute.go
davidz627: you may be able to leverage `CloudDisk` here (originally created for this different api problem for RePD)

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -346,14 +346,19 @@ func TestCreateVolumeArguments(t *testing.T) {
 			},
 		},
 		{
-			name: "fail with MULTI_NODE_MULTI_WRITER capability",
+			name: "success with block/MULTI_NODE_MULTI_WRITER capabilities",
```
davidz627: what about `fs/MULTI_NODE_MULTI_WRITER` should we fail?

msau42:
@davidz627 I think the e2e is going to require the boskos projects to be whitelisted before we can merge it

pkg/gce-cloud-provider/compute/gce-compute.go
sschmitt: Thanks for the review. I notice the `waitForOp` methods only refer to `op.Name`. I wonder if it's okay to call the non-alpha operations API to check on the status of an alpha operation. My thinking there is that we're not leveraging alpha features of the operations API.
I'm not sure so I'll err on the side of caution here. The solution you recommended works just fine.

davidz627: ack. I didn't notice that - we could just grab the op name then and avoid the whole type assertion thing

davidz627: wait... looked at it again and we still might need to since we're doing the op get on `svc.ZoneOperations.Get(project, zone, op.Name).Context(ctx).Do()` - thats the v1 service. If we have an alpha service op maybe we can't "find it" unless we use a `v1alpha.ZoneOperations` call? Could you verify

sschmitt: David, I ran some tests and found that the v1 operations API is able to get the status of beta and alpha operations.

This can easily be tested using the in-browser API Explorer:
https://cloud.google.com/compute/docs/reference/rest/beta/disks/insert
https://cloud.google.com/compute/docs/reference/rest/v1/zoneOperations/get

I think there also might be some opportunity to reduce the duplication between Zonal and Regional methods but I'll skip that for now. You can let me know your thoughts there. I'm happy to make additional changes.

In the meantime I'll address the rest of your comments and add another commit.

pkg/gce-cloud-provider/compute/gce-compute.go
sschmitt: Yes, GetDisk has to be updated. CloudDisk looks to be the way to go.

pkg/gce-cloud-provider/compute/gce-compute.go
davidz627: lets keep this change focused on adding the multiwriter stuff - feel free to open up a fix to reduce zonal+regional duplication afterwards if you're interested. That would be really cool :+1: 

Ack on the `ops` - lets dedupe them and make sure to add a comment in the code with that exact finding (so someone doesn't come in and waste cycles trying to "fix" it later)

sschmitt:
> Mostly looks good, just some comments about code de-dupe and maybe some missed pieces.
> 
> Could you please write some E2E tests for this too? You can see examples in:
> `test/e2e/tests/single_zone_e2e_test.go`
> and run them locally with:
> `test/run-e2e-local.sh`

@davidz627 E2E is a little tricky because this feature is still in Alpha. I also notice that the E2E tests are hardcoded for us-central1. Multi-writer PD is currently only available in us-east1-a in whitelisted projects.

Should I write E2E tests and leave them commented out or skip them?

davidz627:
> I also notice that the E2E tests are hardcoded for us-central1

This could probably be changed pretty easily - please do so unless it seems like it would be significant additional investment.

> in whitelisted projects.

I will take a look at how to resolve this and try get the projects all whitelisted - I will update this PR when I know more.

> Should I write E2E tests and leave them commented out or skip them?

Yes, for now. You should be able to run the tests in your own whitelisted project[s]

sschmitt:
@davidz627 Perfect. Thanks for the guidance.

pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -182,15 +182,22 @@ func (cloud *FakeCloudProvider) ValidateExistingDisk(ctx context.Context, resp *
 		return fmt.Errorf("disk already exists with incompatible type. Need %v. Got %v",
 			diskType, respType[len(respType)-1])
 	}
+
+	// We are assuming here that a multiWriter disk could be used as non-multiWriter
+	if multiWriter && !resp.GetMultiWriter() {
```
davidz627: how about the other way around

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -298,7 +305,7 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 		PublishContext: nil,
 	}
 
-	_, err = gceCS.CloudProvider.GetDisk(ctx, volKey)
+	_, err = gceCS.CloudProvider.GetDisk(ctx, volKey, gce.V1)
```
davidz627: why is this definitely `V1`? Couldn't it be a alpha multiwriter disk that we want to get here

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -248,16 +330,31 @@ func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta
 		}
 	}
 
-	insertOp, err := cloud.service.RegionDisks.Insert(cloud.project, volKey.Region, diskToCreate).Context(ctx).Do()
+	if apiVersion == Alpha {
+		var insertOp *computealpha.Operation
+		alphaDiskToCreate := convertV1DiskToAlphaDisk(diskToCreate)
+		alphaDiskToCreate.MultiWriter = multiWriter
+		insertOp, err = cloud.alphaService.RegionDisks.Insert(cloud.project, volKey.Region, alphaDiskToCreate).Context(ctx).Do()
+		if err == nil {
```
davidz627: nit: `if insertOp != nil` instead?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -214,22 +252,66 @@ func (cloud *CloudProvider) ValidateExistingDisk(ctx context.Context, resp *Clou
...
+		ReplicaZones:      v1Disk.ReplicaZones,
+		DiskEncryptionKey: dek,
+	}
+}
+
+func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta.Key, diskType string, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID, diskEncryptionKmsKey string, multiWriter bool) error {
+	var (
+		err        error
+		opName     string
+		apiVersion = V1
```
davidz627: Lets call this `gceAPIVersion` to disambiguate.

pkg/gce-cloud-provider/compute/fake-gce.go
sschmitt: The other way around would be when the existing disk is enabled for multi-writer but the user didn't ask for that capability. I assume here that the user would be okay with that.

It's actually quite challenging to check the opposite because the user might not have Alpha API access. I suppose we could first try alpha and if that fails fall back to v1.

pkg/gce-pd-csi-driver/controller.go
sschmitt: The v1 API returns disks that use alpha features, it just doesn't have any information with respect to those features. Here the API call is used to detect the existence of the disk. The disk itself is thrown away and only the error code is parsed. I didn't see a need to use anything other than v1.

sschmitt:
@davidz627 Are there any further concerns? You had a question [here](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/415#discussion_r349385512), I wasn't sure if you saw my response.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -248,16 +330,31 @@ func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta
...
+		var insertOp *computev1.Operation
+		insertOp, err = cloud.service.RegionDisks.Insert(cloud.project, volKey.Region, diskToCreate).Context(ctx).Do()
+		if insertOp != nil {
+			opName = insertOp.Name
+		}
+	}
 	if err != nil {
 		if IsGCEError(err, "alreadyExists") {
-			disk, err := cloud.GetDisk(ctx, volKey)
+			disk, err := cloud.GetDisk(ctx, volKey, GCEAPIVersionV1)
```
davidz627: should this be `gceAPIVersion`?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -267,16 +364,17 @@ func (cloud *CloudProvider) insertRegionalDisk(ctx context.Context, volKey *meta
 		return status.Error(codes.Internal, fmt.Sprintf("unkown Insert disk error: %v", err))
 	}
 
-	err = cloud.waitForRegionalOp(ctx, insertOp, volKey.Region)
+	err = cloud.waitForRegionalOp(ctx, opName, volKey.Region)
 	if err != nil {
 		if IsGCEError(err, "alreadyExists") {
-			disk, err := cloud.GetDisk(ctx, volKey)
+			disk, err := cloud.GetDisk(ctx, volKey, GCEAPIVersionV1)
```
davidz627: shoudl this be possibly alpha if multiwriter is set?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -551,6 +677,16 @@ func opIsDone(op *computev1.Operation) (bool, error) {
 	return true, nil
 }
 
+func alphaOpIsDone(op *computealpha.Operation) (bool, error) {
```
davidz627: is this used anywhere?

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -571,6 +573,74 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			Expect(err).To(BeNil(), "DeleteVolume failed")
+
+			// Validate Disk Deleted
+			_, err = computeAlphaService.Disks.Get(p, zone, volName).Do()
+			Expect(gce.IsGCEError(err, "notFound")).To(BeTrue(), "Expected disk to not be found")
+		}()
+	})
+
+	// Pending while multi-writer feature is in Alpha
+	PIt("Should complete entire disk lifecycle with multi-writer disk", func() {
```
davidz627: did you run these tests in your project?

pkg/gce-pd-csi-driver/controller_test.go
mattcary: @nikhilkathare Any details here?

test/e2e/tests/multi_zone_e2e_test.go
```diff
@@ -257,7 +257,11 @@ func testLifecycleWithVerify(volID string, volName string, instance *remote.Inst
 	if secondMountVerify != nil {
 		// Mount disk somewhere else
 		secondPublishDir := filepath.Join("/tmp/", volName, "secondmount")
```
mattcary, line 259: Sorry if I'm missing something from the test setup, but will this test multizonal shared PD?

pkg/gce-pd-csi-driver/controller_test.go
nikhilkathare: @mattcary  : Hi matt, fs/MULTI_NODE_MULTI_WRITER has handled above as mount/MULTI_NODE_MULTI_WRITER capability.

test/e2e/tests/multi_zone_e2e_test.go
nikhilkathare: Shared PD is yet not supported on multizone. This is the fix added to handle function testLifecycleWithVerify with useBlock=true.

pkg/gce-pd-csi-driver/controller_test.go
nikhilkathare: > @nikhilkathare Any details here?

Hi matt, fs/MULTI_NODE_MULTI_WRITER has handled above as mount/MULTI_NODE_MULTI_WRITER capability.

test/e2e/tests/multi_zone_e2e_test.go
mattcary: Thanks for clarifying.

mattcary:
Thanks for resurrecting this PR, Nikhil. It looks good to me, just a couple of general questions:

1) the multiwriter PD API is beta now. But I think continuing to work here through the alpha API makes sense in case there are new alpha-only features that we end up having to use. But maybe not, WDYT?

2) I believe that in beta only scsi is supported by PD, but by GA it will be NVMe only. Do you see that causing any problems?

nikhilkathare:
Hi Matt, Thanks for your review.
Response for your questions.
1. Agreed. Let us go ahead and use Alpha APIs for now. Beta APIs changes can be done later as required.
2. We don’t see any changes required for NVMe but if required we can make the necessary changes after testing on NVMe PDs whenever it becomes available.

mattcary:
> Hi Matt, Thanks for your review.
> Response for your questions.
> 
> 1. Agreed. Let us go ahead and use Alpha APIs for now. Beta APIs changes can be done later as required.
> 2. We don’t see any changes required for NVMe but if required we can make the necessary changes after testing on NVMe PDs whenever it becomes available.

Cool, thanks for the info.

msau42:
@nikhilkathare this lgtm! Thanks for picking this up! One last thing, would you be able to squash the commits to reduce some of the "merge branch" and "address comments" commits? If you haven't tried it before, I would suggest practicing in another branch before doing it here. Take a look [here](https://github.com/kubernetes/community/blob/d337c9d8aed36b8a3c38db6b0e73a26ccb6f0d8e/contributors/guide/github-workflow.md#squash-commits) for pointers on how to do the squshing.

nikhilkathare:
/cc @msau42 @mattcary Thanks for reviewing the code. Tried to squash the commits using rebase but I could squash only final commits(last 5 commits) which were done by me. When tried to squash further, rebase is erroring out with conflicts while applying commits and exiting to resolve errors manually in other commits. As commits in this PR are done with lot of gap there are many commits that have got in between which squash is not able to resolve. Should we create new branch and take change from this branch to new branch or any other better approach. Let us know what approach you would like us to take ?

msau42:
Hi @nikhilkathare can you try just squashing your own commits then? It's not a big deal if we can't get it to work, it's mainly to just try to cleanup the commit listing a little bit.

nikhilkathare:
@msau42 Thanks for quick response. Yes I have squashed my five changes to one and pushed, this led to removal of tag lgtm from PR. Now no further squash could be done as there is merge(which has other commits) between our PR commits.

msau42:
/lgtm
/approve

Thank you so much!

PR #417, opened 2019-10-24
Add blanket tolerations to node daemonsets

### LINKED ISSUE #416: System critical daemon-set should include blanket toleration
```
tolerations:
      - effect: NoExecute
        operator: Exists
      - effect: NoSchedule
        operator: Exists
```

This makes sure the pod will never be evicted by a taint
/assign

davidz627:
/kind feature

/assign @msau42 @verult 

Fixes #416 

```release-note
Add a blanket toleration to the Node Daemonset of the driver deployment so that it can be deployed on all nodes
```

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -52,7 +52,8 @@ spec:
               mountPath: /csi
             - name: device-dir
               mountPath: /dev
-            # The following mounts are required to trigger host udevadm from container
+            # The following mounts are required to trigger host udevadm from
+            # container
             - name: udev-rules-etc
               mountPath: /etc/udev
             - name: udev-rules-lib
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/base/node.yaml
```diff
@@ -95,3 +97,12 @@ spec:
...
             path: /sys
             type: Directory
+      tolerations:
+      # Blanket tolerations to make sure the node component is scheduled on all
+      # nodes.
+      - effect: NoSchedule
+        operator: Exists
+      - effect: NoExecute
+        operator: Exists
+      - key: CriticalAddonsOnly
```
msau42: double check we can actually set things like this without being in the kube-system namespace.

Also, is system priorityclass useful?

misterikkit: I wonder why we don't just to a blanket toleration like,
```
- operator: Exists
```

which matches all keys and effects?

davidz627: @misterikkit do you mean if we don't specify a `Key` the behavior will be to match *all* keys not *none* keys? (I haven't looked at docs but that feels backwards to me).

IIUC priority class will cause the driver to be ranked higher in scheduling order and reduce possibility of pre-emption (only by higher priority pods)

davidz627: found the doc for toleration. Updated

PR #427, opened 2019-11-06
Set CPU and MEM requests for Node component at fairly conservative levels.

davidz627:
They are set based on utilization of concurrently mounting 100 volumes on the node at once of 5Gi size. Larger disk sizes cause larger spikes in usage but for very short periods of time.

/kind feature

```release-note
Set Node Daemonset CPU and MEM requests to 45m and 55MiB respectively
```

### DIFF
--- a/deploy/kubernetes/base/node.yaml
+++ b/deploy/kubernetes/base/node.yaml
@@ -30,6 +30,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+          resources:
+            requests:
+              cpu: "5m"
+              memory: "5Mi"
           volumeMounts:
             - name: plugin-dir
               mountPath: /csi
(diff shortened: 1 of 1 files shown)


jingxu97:
Once request is set, it will prevent pod being scheduled on the node if requested resources is not sufficient. Controller does not monitor the actual resource usage, so even if node has sufficient resources, it will fail to start pods. A very common situation I have experienced is that even node has sufficient CPU, but fail to start any new pods because existing pods has some default request value even though it does not actually consumes that much.
Basically I am trying to understand what is the benefit of setting the request value?

msau42:
> concurrently mounting 100 volumes on the node at once

I don't think this is a common usage pattern, so I'm not sure we should set requests this high.  By reserving this much CPU, that prevents normal pods from getting scheduled to the node.

jingxu97:
> > concurrently mounting 100 volumes on the node at once
> 
> I don't think this is a common usage pattern, so I'm not sure we should set requests this high. By reserving this much CPU, that prevents normal pods from getting scheduled to the node.

On the other hand, it might also prevent the daemonSet pod getting scheduled?

PR #444, opened 2019-12-11
Use canary quay images to test latest sidecars

msau42:
**What type of PR is this?**
/kind failing-test

Use the public canary quay images until we redo our CI pipeline for latest testing

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/overlays/prow-gke-release-staging-head/kustomization.yaml
+++ b/deploy/kubernetes/overlays/prow-gke-release-staging-head/kustomization.yaml
@@ -7,14 +7,14 @@ images:
   newName: gcr.io/gke-release-staging/gcp-compute-persistent-disk-csi-driver
   newTag: "latest"
 - name: gke.gcr.io/csi-provisioner
-  newName: gcr.io/gke-release-staging/csi-provisioner
-  newTag: "latest"
+  newName: quay.io/k8scsi/csi-provisioner
+  newTag: "canary"
 - name: gke.gcr.io/csi-attacher
-  newName: gcr.io/gke-release-staging/csi-attacher
-  newTag: "latest"
+  newName: quay.io/k8scsi/csi-attacher
+  newTag: "canary"
 - name: gke.gcr.io/csi-node-driver-registrar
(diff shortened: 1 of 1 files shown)


deploy/kubernetes/overlays/prow-gke-release-staging-head/kustomization.yaml
```diff
@@ -7,14 +7,14 @@ images:
   newName: gcr.io/gke-release-staging/gcp-compute-persistent-disk-csi-driver
   newTag: "latest"
```
msau42: hmmm this is a problem too isn't it?

msau42: maybe we need to enable GCE_PD_DO_DRIVER_BUILD?

davidz627: yeah this tag is really old now - from sep 9.

Enabling the build flag isn't so easy - then we also need to add logic to go and parse this kustomization file and swap out "latest" for whatever the new build tag is... Though that logic already exists for the pre-submits so maybe its not so hard to port over to this file too. Want to do that or follow up later with it?

We should really figure out what to do about the test tag versions

msau42: From a quick glance, I think our k8s-integration.go should be able to handle substituting the image properly when we set the flag. It takes [both overlay and registry](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/driver.go#L15) as an argument. So no extra changes needed here.

davidz627: ack. are you adding that change to this PR?

msau42: no, setting DO_DRIVER_BUILD is not in this repo. I'll send out a separate pr for that

PR #448, opened 2020-01-08
Add psp to allow deploying csi-gce-pd-node

ffilippopoulos:
**What type of PR is this?**
/kind feature

In case of a cluster with `PodSecurityPolicies` enabled using kustomization base will not be enough to deploy `csi-gce-pd-node` and we have to also create the psp resource with local manifests. This PR is adding the missing psp that we used to deploy `csi-gce-pd-node`.
Since `podsecuritypolicy` is a kube resource available to all clusters regardless if the feature is enabled or not this will not impact users that do not use policies.

Users that patch psps locally will have to drop their patches and use the base, others shall not be affected.

```release-note
remove existing local podsecuritypolicies manifests for `csi-gce-pd-node
```

### DIFF
--- a/deploy/kubernetes/base/setup-cluster.yaml
+++ b/deploy/kubernetes/base/setup-cluster.yaml
@@ -138,3 +138,56 @@ roleRef:
   kind: ClusterRole
   name: csi-gce-pd-resizer-role
   apiGroup: rbac.authorization.k8s.io
+
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: csi-gce-pd-node-psp
+spec:
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  privileged: true
+  volumes:
+  - '*'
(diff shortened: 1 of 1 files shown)


ffilippopoulos:
cncf cla signed

davidz627:
/lgtm
/approve
Thanks @ffilippopoulos!

davidz627:
oops. Jumped the gun there.
@ffilippopoulos the only thing I wanted to ask of you is could you improve the release note to have a description of what this change is adding. If user action is required please note that in a separate line in all caps along with the precondition and action required

ffilippopoulos:
@davidz627 thank you very much for looking into this. I updated the release note in the PR description.

verult:
/cc

@ffilippopoulos does the gce.privileged PSP satisfy your requirements? https://github.com/kubernetes/kubernetes/blob/3491129d348eb6fb5e4274112df5a946ded740af/cluster/gce/addons/podsecuritypolicies/privileged.yaml#L4

AFAIU it should be installed by default if PSP is enabled.

ffilippopoulos:
@verult as far as I can see that policy doesn't specify any host paths that are allowed to mount and is a bit wide regarding the allowed host ports to bind. We prefer to use psps that fit the deployment precisely as best practice and not pass out more privileges than we should.

PR #449, opened 2020-01-10
Allow disabling particular driver services

rfranzke:
**What type of PR is this?**
/kind feature

This PR adds the new `zone` field to the GCE cloud config and makes the controller server independent of the GCE metadata service if both `project-id` and `zone` are present in the config.
Also, it adds two new command line flags `run-controller-service` and `run-node-service` (both default to `true`) which allow to particularly disable individual services if not required.

This enables the use-case of running the CSI controllers (csi-provisioner, csi-attacher, etc., + the driver controller) separately outside of the cluster, and not necessarily on an GCE compute instance.

```release-note
It is now possible to disable the controller service by setting `--run-controller-service=false`. Similarly, it is possible to disable the node service by setting `--run-node-service=false`. The latter enables running the controller server of the GCE PD driver separately/outside of the cluster it is serving. Also, if both `project-id` and `zone` are specified in the GCE cloud config then the controller server does no longer try to contact the GCE metadata service.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -30,9 +30,11 @@ import (
 )
 
 var (
-	endpoint          = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
-	gceConfigFilePath = flag.String("cloud-config", "", "Path to GCE cloud provider config")
-	vendorVersion     string
+	cloudConfigFilePath  = flag.String("cloud-config", "", "Path to GCE cloud provider config")
+	endpoint             = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	runControllerService = flag.Bool("run-controller-service", true, "If set to false then the CSI driver does not activate its controller service (default: true)")
+	runNodeService       = flag.Bool("run-node-service", true, "If set to false then the CSI driver does not activate its node service (default: true)")
+	vendorVersion        string
 )
 
 const (
@@ -57,6 +59,8 @@ func main() {
 }
 
 func handle() {
+	var err error
+
 	if vendorVersion == "" {
 		klog.Fatalf("vendorVersion must be set at compile time")
 	}
@@ -68,20 +72,35 @@ func handle() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	cloudProvider, err := gce.CreateCloudProvider(ctx, vendorVersion, *gceConfigFilePath)
-	if err != nil {
-		klog.Fatalf("Failed to get cloud provider: %v", err)
+	//Initialize identity server
+	identityServer := driver.NewIdentityServer(gceDriver)
+
+	//Initialize requirements for the controller service
+	var controllerServer *driver.GCEControllerServer
+	if *runControllerService {
+		cloudProvider, err := gce.CreateCloudProvider(ctx, vendorVersion, *cloudConfigFilePath)
+		if err != nil {
+			klog.Fatalf("Failed to get cloud provider: %v", err)
+		}
+		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider)
+	} else if *cloudConfigFilePath != "" {
+		klog.Warningf("controller service is disabled but cloud config given - it has no effect")
 	}
 
-	mounter := mountmanager.NewSafeMounter()
-	deviceUtils := mountmanager.NewDeviceUtils()
-	statter := mountmanager.NewStatter()
-	ms, err := metadataservice.NewMetadataService()
-	if err != nil {
-		klog.Fatalf("Failed to set up metadata service: %v", err)
+	//Initialize requirements for the node service
+	var nodeServer *driver.GCENodeServer
+	if *runNodeService {
+		mounter := mountmanager.NewSafeMounter()
+		deviceUtils := mountmanager.NewDeviceUtils()
+		statter := mountmanager.NewStatter()
+		meta, err := metadataservice.NewMetadataService()
+		if err != nil {
+			klog.Fatalf("Failed to set up metadata service: %v", err)
+		}
+		nodeServer = driver.NewNodeServer(gceDriver, mounter, deviceUtils, meta, statter)
 	}
 
-	err = gceDriver.SetupGCEDriver(cloudProvider, mounter, deviceUtils, ms, statter, driverName, vendorVersion)
+	err = gceDriver.SetupGCEDriver(driverName, vendorVersion, identityServer, controllerServer, nodeServer)
 	if err != nil {
 		klog.Fatalf("Failed to initialize GCE CSI Driver: %v", err)
 	}
--- a/docs/kubernetes/user-guides/driver-install.md
+++ b/docs/kubernetes/user-guides/driver-install.md
@@ -18,7 +18,7 @@ compute.instances.get
 compute.instances.attachDisk
 compute.instances.detachDisk
 roles/compute.storageAdmin
-roles/iam.serviceAccountUser 
+roles/iam.serviceAccountUser
 ```
 
 If there is a pre-existing service account with these roles for use then the
@@ -80,3 +80,21 @@ iam.serviceAccounts.delete
 
 These permissions are not required if you already have a service account ready
 for use by the PD Driver.
+
+## Disabling particular CSI driver services
+
+Traditionally, you run the CSI controllers with the GCE PD driver in the same Kubernetes cluster.
+Though, there may be cases where you will only want to run a subset of the available driver services (for example, one scenario is running the controllers outside of the cluster they are serving (while the GCE PD driver still runs inside the served cluster), but there might be others scenarios).
+The CSI driver consists out of these services:
+
+* The **controller** service starts the GRPC server that serves `CreateVolume`, `DeleteVolume`, etc. It is depending on the GCP service account credentials and talks with the GCP API.
+* The **identity** service is responsible to provide identity services like capability information of the CSI plugin.
+* The **node** service implements the various operations for volumes that are run locally from the node, for example `NodePublishVolume`, `NodeStageVolume`, etc. It does not do operations like `CreateVolume` or `ControllerPublish`. Also, as it runs directly on the GCE instances, it is depending on the GCE metadata service.
+
+The CSI driver has two command line flags, `--run-controller-service` and `--run-node-service` which both default to `true`.
+You can disable the individual services by setting the respective flags to `false`.
+
+Note: If you want to run the CSI controllers outside of the cluster you have to specify both the `zone` and `projectId` parameters in the GCE cloud provider config.
+The `zone` is the name of one of the availability zones the served Kubernetes cluster is deployed to.
+It is used to derive the GCP region and to discover the other availability zones in this region.
+The `project-id` is the GCP project ID in which the controller is operating.
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -38,6 +38,9 @@ const (
 )
 
 type GCECompute interface {
+	// Metadata information
+	GetDefaultProject() string
+	GetDefaultZone() string
 	// Disk Methods
 	GetDisk(ctx context.Context, volumeKey *meta.Key) (*CloudDisk, error)
 	RepairUnderspecifiedVolumeKey(ctx context.Context, volumeKey *meta.Key) (*meta.Key, error)
@@ -61,6 +64,16 @@ type GCECompute interface {
 	GetSnapshot(ctx context.Context, snapshotName string) (*computev1.Snapshot, error)
 	CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*computev1.Snapshot, error)
 	DeleteSnapshot(ctx context.Context, snapshotName string) error
+}
+
+// GetDefaultProject returns the project that was used to instantiate this GCE client.
+func (cloud *CloudProvider) GetDefaultProject() string {
+	return cloud.project
+}
+
+// GetDefaultZone returns the zone that was used to instantiate this GCE client.
+func (cloud *CloudProvider) GetDefaultZone() string {
+	return cloud.zone
 }
 
 // ListDisks lists disks based on maxEntries and pageToken only in the project
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -66,6 +66,7 @@ type ConfigGlobal struct {
 	TokenURL  string `gcfg:"token-url"`
 	TokenBody string `gcfg:"token-body"`
 	ProjectId string `gcfg:"project-id"`
+	Zone      string `gcfg:"zone"`
 }
 
 func CreateCloudProvider(ctx context.Context, vendorVersion string, configPath string) (*CloudProvider, error) {
@@ -103,7 +104,6 @@ func CreateCloudProvider(ctx context.Context, vendorVersion string, configPath s
 }
 
 func generateTokenSource(ctx context.Context, configFile *ConfigFile) (oauth2.TokenSource, error) {
-
 	if configFile != nil && configFile.Global.TokenURL != "" && configFile.Global.TokenURL != "nil" {
 		// configFile.Global.TokenURL is defined
 		// Use AltTokenSource
@@ -184,9 +184,16 @@ func newOauthClient(ctx context.Context, tokenSource oauth2.TokenSource) (*http.
 func getProjectAndZone(config *ConfigFile) (string, string, error) {
 	var err error
 
-	zone, err := metadata.Zone()
-	if err != nil {
-		return "", "", err
+	var zone string
+	if config == nil || config.Global.Zone == "" {
+		zone, err = metadata.Zone()
+		if err != nil {
+			return "", "", err
+		}
+		klog.V(2).Infof("Using GCP zone from the Metadata server: %q", zone)
+	} else {
+		zone = config.Global.Zone
+		klog.V(2).Infof("Using GCP zone from the local GCE cloud provider config file: %q", zone)
 	}
 
 	var projectID string
--- a/pkg/gce-pd-csi-driver/gce-pd-driver.go
+++ b/pkg/gce-pd-csi-driver/gce-pd-driver.go
@@ -45,14 +45,10 @@ func GetGCEDriver() *GCEDriver {
 	return &GCEDriver{}
 }
 
-func (gceDriver *GCEDriver) SetupGCEDriver(cloudProvider gce.GCECompute, mounter *mount.SafeFormatAndMount,
-	deviceUtils mountmanager.DeviceUtils, meta metadataservice.MetadataService, statter mountmanager.Statter, name, vendorVersion string) error {
+func (gceDriver *GCEDriver) SetupGCEDriver(name, vendorVersion string, identityServer *GCEIdentityServer, controllerServer *GCEControllerServer, nodeServer *GCENodeServer) error {
 	if name == "" {
 		return fmt.Errorf("Driver name missing")
 	}
-
-	gceDriver.name = name
-	gceDriver.vendorVersion = vendorVersion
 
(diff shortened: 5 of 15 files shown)


rfranzke:
@davidz627 @msau42 @jingxu97 any feedback for this PR?

hantaowang:
why have `SetupGCEDriver1 take `runAsController` as a *bool instead of just the bool? Seems weird because then in all tests you declare a bool and then take the address of it, despite the bool being a built in type. 

Another nit is that (correct me if im wrong) but I don't think runAsController can every be a nil pointer since it defaults to false. 

Its a while since I've looked at this code but it looks like your changes to pkg/gce-cloud-provider/compute/gce.go and renaming gceConfigFilePath are separate from the goal of this PR.

rfranzke:
> why have SetupGCEDriver1 take runAsController as a *bool instead of just the bool? Seems weird because then in all tests you declare a bool and then take the address of it, despite the bool being a built in type.
Another nit is that (correct me if im wrong) but I don't think runAsController can ever be a nil pointer since it defaults to false.

Thanks, valid feedback, I improved it, PTAL.

> Its been a while since I've looked at this code but it looks like your changes to pkg/gce-cloud-provider/compute/gce.go 

Yes, they are related to the goal. Without this change you can't run the driver in the controller mode on a non-GCE instance.

> and renaming gceConfigFilePath are separate from the goal of this PR.

This is not related to the goal. If it matters that much I can also keep it as it was before.

cmd/main.go
```diff
@@ -33,9 +33,10 @@ func init() {
 }
 
 var (
-	endpoint          = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
-	gceConfigFilePath = flag.String("cloud-config", "", "Path to GCE cloud provider config")
-	vendorVersion     string
+	endpoint            = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	cloudConfigFilePath = flag.String("cloud-config", "", "Path to GCE cloud provider config")
+	runAsController     = flag.Bool("run-as-controller", false, "If set to true then the CSI driver runs the controller only and does not start the node service.")
```
davidz627: don't want to expand the scope of this PR too much but thinking about the "end-state" of splitting this out is useful to inform the design.

IMO the driver should have 3 distinct booleans:
`run-controller-service`, `run-node-service`, (maybe) `run-identity-service`.

These should all default to true. You can turn each of them off piecemeal if required. WDYT?

docs/kubernetes/user-guides/driver-install.md
```diff
@@ -20,4 +20,13 @@ $ ./deploy/setup-project.sh
...
 $ GCE_PD_DRIVER_VERSION=stable                    # Driver version to deploy
 $ ./deploy/kubernetes/deploy-driver.sh
-```
\ No newline at end of file
+```
+
+## Running the controller separately
+
+Traditionally, you run the CSI controllers with the GCE PD driver in the same Kubernetes cluster.
+Though, in some scenarios you might want to run the CSI controllers (csi-provisioner, csi-attacher, etc.) together with the GCE PD controller service of this driver separately from the Kubernetes cluster it serves (while the GCE PD driver with an activated node service still runs inside the cluster).
```
davidz627: Though this is one use case I don't think it's the only one. I would make the description more general: "there may be cases where you will only want to run a subset of the available driver services:
- brief description of controller srevice and its dependencies (for example GCP Service Account Creds)
- brief description of node service and dependencies (for example metadata server)"

You could take the scenario written here and make it an example following the description.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -63,9 +63,10 @@ type ConfigFile struct {
 }
 
 type ConfigGlobal struct {
+	LocalZone string `gcfg:"local-zone"`
```
davidz627: can you document what the `LocalZone` is and when you would want to use it?

rfranzke:
Thanks for your feedback @davidz627, can you now check again? I updated the PR based on your suggestions. Please double-check the docs section.

cmd/main.go
```diff
@@ -33,9 +33,12 @@ func init() {
 }
 
 var (
-	endpoint          = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
-	gceConfigFilePath = flag.String("cloud-config", "", "Path to GCE cloud provider config")
-	vendorVersion     string
+	endpoint             = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	cloudConfigFilePath  = flag.String("cloud-config", "", "Path to GCE cloud provider config")
+	runControllerService = flag.Bool("run-controller-service", true, "If set to false then the CSI driver does not activate its controller service (default: true)")
+	runIdentityService   = flag.Bool("run-identity-service", true, "If set to false then the CSI driver does not activate its identity service (default: true)")
```
davidz627: sorry for the churn - on second though, i don't think we need the ability to turn off the identity server since anything using the driver is most likely going to have to be able to interrogate the driver for at least its name.

cmd/main.go
```diff
@@ -66,22 +69,25 @@ func handle() {
...
 
 	mounter := mountmanager.NewSafeMounter()
 	deviceUtils := mountmanager.NewDeviceUtils()
 	statter := mountmanager.NewStatter()
-	ms, err := metadataservice.NewMetadataService()
-	if err != nil {
-		klog.Fatalf("Failed to set up metadata service: %v", err)
+
+	var ms metadataservice.MetadataService
+	if !*runNodeService {
```
davidz627: i thought we *do* set up the metadata service if we're running nodeService?

cmd/main.go
```diff
@@ -66,22 +69,25 @@ func handle() {
 
 	gceDriver := driver.GetGCEDriver()
 
-	//Initialize GCE Driver (Move setup to main?)
-	cloudProvider, err := gce.CreateCloudProvider(vendorVersion, *gceConfigFilePath)
+	// Initialize GCE Driver (Move setup to main?)
+	cloudProvider, err := gce.CreateCloudProvider(vendorVersion, *cloudConfigFilePath)
```
davidz627: do we need the cloud provider if we're not running the controller service?

cmd/main.go
```diff
@@ -66,22 +69,25 @@ func handle() {
...
 	if err != nil {
 		klog.Fatalf("Failed to get cloud provider: %v", err)
 	}
 
 	mounter := mountmanager.NewSafeMounter()
```
davidz627: i dont think we need things like the mounter, deviceUtils, or statter if we're not running the node service

docs/kubernetes/user-guides/driver-install.md
```diff
@@ -20,4 +20,21 @@ $ ./deploy/setup-project.sh
...
+
+## Disabling particular CSI driver services
+
+Traditionally, you run the CSI controllers with the GCE PD driver in the same Kubernetes cluster.
+Though, there may be cases where you will only want to run a subset of the available driver services (for example, one scenario is running the controllers outside of the cluster they are serving (while the GCE PD driver still runs inside the served cluster), but there might be others scenarios).
+The CSI driver consists out of these services:
+
+* The **controller** service starts the GRPC server that is used for communication between the CSI components. It is depending on the GCP service account credentials.
+* The **identity** service is responsible to provide identity services like capability information of the CSI plugin.
+* The **node** service implements the various operations for volumes like creation, deletion, etc. It is depending on the GCE metadata service.
```
davidz627: node implements operations that are run locally from the node. For example `NodePublishVolume`, `NodeStageVolume`. It does *not* do operations like `Create` or `ControllerPublish`

docs/kubernetes/user-guides/driver-install.md
```diff
@@ -20,4 +20,21 @@ $ ./deploy/setup-project.sh
...
\ No newline at end of file
+```
+
+## Disabling particular CSI driver services
+
+Traditionally, you run the CSI controllers with the GCE PD driver in the same Kubernetes cluster.
+Though, there may be cases where you will only want to run a subset of the available driver services (for example, one scenario is running the controllers outside of the cluster they are serving (while the GCE PD driver still runs inside the served cluster), but there might be others scenarios).
+The CSI driver consists out of these services:
+
+* The **controller** service starts the GRPC server that is used for communication between the CSI components. It is depending on the GCP service account credentials.
```
davidz627: This isn't entirely accurate - the controller is the component that serves `CreateVolume`, `DeleteVolume` etc. Although from a cursory glance at the code it does seem like the controller component also depends on the metadataservice for getting the `project` and `zones` for particular operations 

I'm starting to think that maybe your goal is to not actually be to turn off particular services - instead is it to specify a "project" and "zone" external to the cluster that the driver is running on? In that case there is no need to turn off particular services, you can add some options to provide project and zone as flags and use those values if they are provided

docs/kubernetes/user-guides/driver-install.md
```diff
@@ -20,4 +20,21 @@ $ ./deploy/setup-project.sh
...
+* The **identity** service is responsible to provide identity services like capability information of the CSI plugin.
+* The **node** service implements the various operations for volumes like creation, deletion, etc. It is depending on the GCE metadata service.
+
+The CSI driver has three command line flags, `--run-controller-service`, `--run-identity-service`, and `--run-node-service` which all default to `true`.
+You can disable individually services by setting the respective flags to `false`.
+
+Note: If you want to run the CSI controllers outside of the cluster you might need to specify the `local-zone` parameter in the GCE cloud provider config.
```
davidz627: I now see why this was included in the PR. I think this is actually the main part that aligns with your goal right? Don't we need the `project` as well then? I don't think `local-zone` and `local-project` are the right names for these. They're actually `client-project`, `client-zone`?

davidz627: it seems to me what you're actually looking for is to do a "`MetadataServer` override" in some sense. You could then have a flag that is `use-local-metadata`. When this value is `true (default)` the components will use the metadata service to get local zone and project. When false, a project and zone must be specified through the cloud config and the driver can be deployed outside of the client project.

cmd/main.go
davidz627: we actually need this in both node and controller service.

docs/kubernetes/user-guides/driver-install.md
msau42: I'm actually not sure why you need a "local-zone" parameter. With the topology feature, the controller can already provision in zones that it's not running in. Is this for the "fallback" behavior when you're not using topology?

msau42: and the downside of supporting only local-zone is you won't be able to support multi-zone clusters in the future, and potential Regional PD as well.

davidz627: right this would only be for the "default" when no topology is specified. For regional PD if no topology is specified we actually take a specified zone (from metadata service rn) and enumerate all zones in that region to pick a second zone from (of course this doesn't actually work well in practice as you would probably care what other zone you pick). 

@msau42 is actually right, using cloud-config and providing the project + topology should already be enough for the driver to provision and attach volumes in projects/zones it's not running in.

cmd/main.go
rfranzke: Do you want me to add another flag indicating the the driver is not running on a GCE instance? The metadata service is only available there.

docs/kubernetes/user-guides/driver-install.md
rfranzke: Thanks for your comments, though, I'm a little bit confused now and need more help/guidance.

@davidz627 you are right, what I effectively want to prevent is that the driver is talking to the metadata service. For the project name it was already possible to specify it via the cloud config, and the metadata svc was not contacted: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce.go#L192-L205
I just "replicated" the same approach for the zone (FYI: the name `local-zone` is inspired from https://github.com/kubernetes/legacy-cloud-providers/blob/master/gce/gce.go#L182-L185, but we could call it differently, sure).

> and the downside of supporting only local-zone is you won't be able to support multi-zone clusters in the future, and potential Regional PD as well.

This comment I don't understand - why do you [query the zone from the metadata service](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce.go#L187-L190) then in the first place (if it's not needed/prevents multi-zone clusters and potential regional PDs)?

From what I see in the code, the zone information is only used to [derive the region name](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L87) and to [discover the other zones in this region](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L135-L151). There is another usage when [listing disks in a zone](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L66-L81) - however, how does all of this work for multi-zone clusters already with today's `master` branch (where the zone is coming from the metadata svc, i.e., the controller can only handle the same zone where it is running in)?

> right this would only be for the "default" when no topology is specified. For regional PD if no topology is specified we actually take a specified zone (from metadata service rn) and enumerate all zones in that region to pick a second zone from (of course this doesn't actually work well in practice as you would probably care what other zone you pick).

In my case, I don't want to assume the controller is running on a GCE instance and has access to the metadata service.

> @msau42 is actually right, using cloud-config and providing the project + topology should already be enough for the driver to provision and attach volumes in projects/zones it's not running in.

What do you more concretely suggest to prevent this unconditional metadata svc call: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce.go#L187-L190?

davidz627: I think my concrete suggestions going forward would be:

1) Create a flag for `zone` that mirrors the one for `project` right now in the cloud config that will allow you to bypass the metadata config
    - If both project/zone are specified we should be able to turn off the metadata interrogation completely (through some reasonable combination of flags with validation)
2) Don't do any "controller" or "node" disabling in this PR - if there is a follow-on need for that please submit a seperate PR it is fundamentally a different issue.
3) Once you have the ability to turn off the metadata service - you should provide project/zone in the cloud config - however, I would still recommend using the Topology feature to influence volume topology instead of relying on the defaulting behavior (though the defaults may work in your case with single zonal volumes).

Please let me know if any of that doesn't make sense. Hoping it helps

rfranzke: Thank you, I'll try going in this direction. Though, I'm not fully aware of the volume topology that you are referring to (as I'm new to CSI). Can you point me in the right direction here, e.g., some place in the code where I can see it in action?

davidz627: Sure, thanks for being so flexible! For topology you can try reading the docs here: https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode. It's essentially the way you can influence where PD's are created and therefore scheduled (either dynamically based on Pod using the volume with `WaitForFirstConsumer` or by specifying Topology directly).

cmd/main.go
davidz627: lets call it `use-metadata-server` and do some validation that `if !use-metadataserver then project & zone must be set on cloud config`. Basically the metadataserver becomes the "backup" method to get zone and project on a GCE instance if the project and zone are not provided directly

rfranzke:
Hi again, meanwhile I had time to update the PR.

I was first trying to implement @davidz627's suggestion (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/449#discussion_r372119652), however, I reached a point where it became clear that this isn't enough for my use-case.

Just to reiterate: I want to allow running the controller server of the GCE PD CSI driver separately/outside of the cluster that it is serving - potentially not even on a GCE instance.
With only adding `zone` to the GCE cloud config and making the controller server independent of the GCE metadata service this would not be possible because the node server would still be activated - and it is relying on the metadata service.

I reached out to @davidz627 via Slack and we agreed on introducing the earlier discussed `run-controller-service` and `run-node-service` flag that allow to disable the particular services individually. This enables use-cases like mine, but probably even more.

TL;DR - The PR now comprises two commits:
1. Making the controller service independent of the GCE metadata service if both `project-id` and `zone` are specified in its config
2. Allow disabling the controller and/or node service via command line flags

The tests are now clearly instantiating only the objects that are really needed for the various servers (e.g., the identity server test does no longer create a fake metadata service).

Please take a look, I'm happy to get your feedback on this.

pkg/gce-pd-csi-driver/gce-pd-driver.go
```diff
@@ -80,8 +80,14 @@ func (gceDriver *GCEDriver) SetupGCEDriver(cloudProvider gce.GCECompute, mounter
 
 	// Set up RPC Servers
 	gceDriver.ids = NewIdentityServer(gceDriver)
-	gceDriver.ns = NewNodeServer(gceDriver, mounter, deviceUtils, meta, statter)
-	gceDriver.cs = NewControllerServer(gceDriver, cloudProvider)
+
+	if mounter != nil && deviceUtils != nil && meta != nil && statter != nil {
+		gceDriver.ns = NewNodeServer(gceDriver, mounter, deviceUtils, meta, statter)
```
davidz627: I would make a "small" refactor here so that we're creating the actual `NewNodeServer` and `NewControllerServer` inside `handle()` where you have the "Initializer requirements for the controller service". Then have `SetupGCEDriver` take in node server and controller server as inputs if possible. That cleans up the logic significantly and puts all of the "run-{service}" logic into the same spot so it's easier to understand.

Right now my concern is if you accidentally forget to send in the `meta` for example, then your `NodeServer` might not start despite that not being your intention since you passed in `mounter`, `deviceUtils`...etc.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -63,6 +66,16 @@ type GCECompute interface {
 	DeleteSnapshot(ctx context.Context, snapshotName string) error
 }
 
+// GetProject returns the project that was used to instantiate this GCE client.
+func (cloud *CloudProvider) GetProject() string {
+	return cloud.project
+}
+
+// GetZone returns the zone that was used to instantiate this GCE client.
+func (cloud *CloudProvider) GetZone() string {
```
davidz627: would it make sense to call these `GetDefaultZone` `GetDefaultProject`? I'm not sure if default is the right word but I feel like it needs some sort of qualifier since it's not clear "what zone" are we getting.

rfranzke: Sure, done. I called it as you suggested (with "default" now, although, I'm also not sure if that's the best name). If a better name comes into your mind shout it out :)

pkg/gce-pd-csi-driver/gce-pd-driver.go
rfranzke: Thanks, very nice suggestion, indeed, this is way better!

rfranzke: FYI: Due to consistency, I also moved the `NewIdentityServer` function out and made it a parameter in the `SetupGCEDriver` function.

cmd/main.go
```diff
@@ -68,20 +72,35 @@ func handle() {
...
+	//Initialize requirements for the controller service
+	var controllerServer *driver.GCEControllerServer
+	if *runControllerService {
+		cloudProvider, err := gce.CreateCloudProvider(ctx, vendorVersion, *cloudConfigFilePath)
+		if err != nil {
+			klog.Fatalf("Failed to get cloud provider: %v", err)
+		}
+		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider)
+	} else if *cloudConfigFilePath != "" {
+		klog.Warningf("controller service is disabled but cloud config given - it has no effect")
```
davidz627: nice

Issue #419, opened 2019-10-25
Migrate to go mod


davidz627:
Instructions can be found here: https://github.com/kubernetes-csi/external-attacher/tree/master/release-tools#dependencies-and-vendoring

PR #451, opened 2020-01-15
Switch over to go mod 

davidz627:
/kind cleanup
/assign @msau42 @misterikkit 

Fixes: #419 

```release-note
NONE
```


davidz627:
I'm considering removing the vendor directly entirely. Please comment thoughts

davidz627:
lots of work required to bring csi-sanity up (need to update mount packages) - turning off sanity tests for now

davidz627:
Removed vendor directory entirely - AFAIK the main arguments for keeping it and their refutations are:
1) Compatibility with older versions of tools for building -> Those are no longer supported and the kubernetes ecosystem requires go 1.13+ now
2) In case some of the repos disappear off the earth -> partially solved by the go mod proxy IIUC and we only depend on other large repos in the kubernetes/Google/Golang ecosystem which we're completely tied to anyway. If Google goes away we don't really need this driver do we.

Also did some sketchy things so I had to disable csi sanity for now: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/452

davidz627:
Another reason to keep vendor is review of dependencies - updates and new dependencies will have clear vendor/ code updates on PRs. However, people don't look at those anyway. Also, our build pipeline does automated scanning of vendored dependencies and handles the licenses as well

PR #454, opened 2020-01-16
Break k/k dependency by migrating util/mount and copying resizefs. Re-enable sanity testing. Fix deps.

### LINKED ISSUE #452: Re-enable sanity tests and NodeGetVolumeStats Unit test
Disabled as part of the migration to `go mod` for development velocity. I mainly shot myself in the foot by generating some hacks in the long past so those need to get fixed before re-enabling the tests.

Basically NodeGetVolumeStats uses a fake mounter for unit testing and sanity testing but the fake mounter had a bug. I fixed the bug directly in the `vendor/` directory but did not upstream it because a TON of changes were being made to the mounter at the same time (including moving the folder out of k/k entirely). Now that's come to bite me in the butt and we need to change the whole repo's dependency over to the new mounter library in order to get the fix with `go mode` without using the vendor dir.

davidz627:
Explicitly bump versions of k8s.io/apimachinery, k8s.io/utils, csi.k8s.io/csi-test. Migrate from k/k/util/mnt to k/utils/mnt and copy resizefs impl breaking k/k dependency :) Fix sanity testing and node unit tests.

/kind cleanup
/assign @misterikkit @msau42 

Fixes #452 

```release-note
NONE
```


davidz627:
Some of the resize e2e tests are failing because of permissions things. I will look into it tomorrow

davidz627:
@misterikkit @msau42 should be ready for review now :)

PR #455, opened 2020-01-18
Update volume limits for new machine types: https://cloud.google.com/compute/docs/machine-types

### LINKED ISSUE #445: Support attach limits for new machine types
There are a ton of new machines types and most should have 127 as the attach limit 

https://cloud.google.com/compute/docs/machine-types

/help

davidz627:
/kind bug
/kind cleanup

Fixes #445 

/assign @msau42 

```release-note
Set volume limits to 15 only for machine-types: "f1-micro", "g1-small", "e2-micro", "e2-small", "e2-medium". Limit is 127 for all others
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -505,16 +505,16 @@ func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpa
 }
 
 func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
-	var volumeLimits int64
-
 	// Machine-type format: n1-type-CPUS or custom-CPUS-RAM or f1/g1-type
 	machineType := ns.MetadataService.GetMachineType()
-	if strings.HasPrefix(machineType, "n1-") || strings.HasPrefix(machineType, "custom-") {
-		volumeLimits = volumeLimitBig
-	} else {
-		volumeLimits = volumeLimitSmall
+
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -505,16 +505,16 @@ func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpa
...
-	var volumeLimits int64
-
 	// Machine-type format: n1-type-CPUS or custom-CPUS-RAM or f1/g1-type
 	machineType := ns.MetadataService.GetMachineType()
-	if strings.HasPrefix(machineType, "n1-") || strings.HasPrefix(machineType, "custom-") {
-		volumeLimits = volumeLimitBig
-	} else {
-		volumeLimits = volumeLimitSmall
+
+	smallMachineTypes := []string{"f1-micro", "g1-small", "e2-micro", "e2-small", "e2-medium"}
```
yoesmat: What about using a map of machine types to volume size?

davidz627: there are only two volume size limits and only a very small subset are "small" so that is probably not necessary, see here for an interesting comparison of lookup times between slices and maps: https://www.darkcoding.net/software/go-slice-search-vs-map-lookup/ (breakeven point around 5 items).

The thing I want to avoid is a map for small things with all the value being `smallValue` and then "everything not in the map is big". Or else I would have to be exhaustive with the map which is error prone and repetitive

PR #459, opened 2020-01-28
Update VerifyDevicePath logic to run udevadm --trigger to fix scenarios where disk is not found or is wrong device

### LINKED ISSUE #376: udevadm may be never triggered right now
In the function `VerifyDevicePath`(https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/mount-manager/device-utils.go#L89), the value of `sdBefore` is the same with `sdAfter`. 

Then the function `udevadmChangeToDrive` will never be called. It seems like the code doesn't actually do anything. 

/ping @davidz627

davidz627:
TODO:
- [x] Write an E2E test that tests this functionality: attaches disk -> purposely links the `/dev/by-id/{diskName}` to the wrong device or deletes the `/dev/by-id/{diskName}` entirely then make sure `NodeStage` is able to self-heal that using `udevadm --trigger`

/kind bug
/assign @msau42 
Fixes #376 

`/lib/udev/scsi_id` was pointing to the binary on the host due to the `/lib/udev` hostPath. This was causing weird package dynamically linked library finding issues (couldn't find the right version of dynamic libs). The solution is to use the `/lib/udev/scsi_id` from the container by copying it into a special directory. Thanks @msau42 for pointing out that `/lib/udev` was coming from the host

ig9OoyenpxqdCQyABmOQBZDI0duHk2QZZmWg2Hxd4ro

```release-note
During NodeStageVolume run udevadm --trigger to fix device symlinks if device path is not found or device path points to the wrong device
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -28,10 +28,6 @@ import (
 	mountmanager "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/mount-manager"
 )
 
-func init() {
-	flag.Set("logtostderr", "true")
-}
-
 var (
 	endpoint          = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
 	gceConfigFilePath = flag.String("cloud-config", "", "Path to GCE cloud provider config")
@@ -49,6 +45,7 @@ func init() {
 	// Use V(5) for GCE Cloud Provider Call informational logging
 	// Use V(6) for extra repeated/polling information
 	klog.InitFlags(flag.CommandLine)
+	flag.Set("logtostderr", "true")
 }
 
 func main() {
--- a/pkg/mount-manager/device-utils.go
+++ b/pkg/mount-manager/device-utils.go
@@ -20,10 +20,13 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
+	"time"
 
-	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
+	pathutils "k8s.io/utils/path"
 )
 
 const (
@@ -46,6 +49,14 @@ const (
 	// 'fsck' found errors but exited without correcting them
 	fsckErrorsUncorrected = 4
 	defaultMountCommand   = "mount"
+	// scsi_id output should be in the form of:
+	// 0Google PersistentDisk <disk name>
+	scsiPattern = `^0Google\s+PersistentDisk\s+([\S]+)\s*$`
+)
+
+var (
+	// regex to parse scsi_id output and extract the serial
+	scsiRegex = regexp.MustCompile(scsiPattern)
 )
 
 // DeviceUtils are a collection of methods that act on the devices attached
@@ -57,7 +68,7 @@ type DeviceUtils interface {
 
 	// VerifyDevicePath returns the first of the list of device paths that
 	// exists on the machine, or an empty string if none exists
-	VerifyDevicePath(devicePaths []string) (string, error)
+	VerifyDevicePath(devicePaths []string, diskName string) (string, error)
 }
 
 type deviceUtils struct {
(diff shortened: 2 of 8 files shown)


davidz627:
Wrote the e2e test and confirmed its failing before the `udevadm` logic fix, and passes afterwards :)

test/e2e/utils/utils.go
```diff
@@ -49,8 +49,8 @@ func GCEClientAndDriverSetup(instance *remote.InstanceInfo) (*remote.TestContext
 	endpoint := fmt.Sprintf("tcp://localhost:%s", port)
 
 	workspace := remote.NewWorkspaceDir("gce-pd-e2e-")
-	driverRunCmd := fmt.Sprintf("sh -c '/usr/bin/nohup %s/gce-pd-csi-driver --endpoint=%s> %s/prog.out 2> %s/prog.err < /dev/null &'",
-		workspace, endpoint, workspace, workspace)
+	driverRunCmd := fmt.Sprintf("sh -c '/usr/bin/nohup %s/gce-pd-csi-driver -v=4 --endpoint=%s 2> %s/prog.out < /dev/null > /dev/null &'",
```
davidz627: small fix to improve debugging experience. This is on purpose

cmd/main.go
```diff
@@ -28,10 +28,6 @@ import (
 	mountmanager "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/mount-manager"
 )
 
-func init() {
```
davidz627: we had two init functions before, I don't know if that was causing issues but it doesn't hurt to fix

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -81,7 +84,69 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			err = client.ControllerUnpublishVolume(volID, instance.GetNodeID())
+			if err != nil {
+				klog.Errorf("Failed to detach disk: %v", err)
+			}
+
+		}()
 
+		// MESS UP THE symlink
+
+		err = testutils.RmAll(instance, "/dev/disk/by-id")
```
davidz627: this only tests the "symlink not found" path, need to augment test to also switch symlinks around and test that as well.

pkg/mount-manager/device-utils.go
```diff
@@ -85,75 +95,146 @@ func (m *deviceUtils) GetDiskByIdPaths(deviceName string, partition string) []st
...
-			return udevadmChangeToDrive(sd)
+	scsiIDPath := "/lib/udev_containerized/scsi_id"
+	exists, err := pathutils.Exists(pathutils.CheckFollowSymlink, scsiIDPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to check scsi_id existence: %v", err)
+	}
+	if !exists {
+		// No SCSI ID tool, no udevadm fixing is possible so just return best
+		// effort devicePath.
+		klog.Warningf("Could not find scsi_id tool at %s, no udevadm fixing is possible so we just return the best effort devicePath", scsiIDPath)
```
msau42: Should we make it required? I'm worried that if there's an issue in our container build process that doesn't install it properly, then this problem will go undetected.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -81,7 +84,68 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			// Detach Disk
+			err = client.ControllerUnpublishVolume(volID, instance.GetNodeID())
+			if err != nil {
+				klog.Errorf("Failed to detach disk: %v", err)
+			}
+
+		}()
+
+		// MESS UP THE symlink
+		err = testutils.RmAll(instance, "/dev/disk/by-id")
```
msau42: This has the potential to be incredibly destructive if you run the test locally or not on some throwaway machine. Prefer only touching the specific disk that the test is using.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -81,7 +84,68 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+		Expect(err).To(BeNil(), "failed to remove /dev/by-id folder")
+
+		// Stage Disk
+		stageDir := filepath.Join("/tmp/", volName, "stage")
+		err = client.NodeStageExt4Volume(volID, stageDir)
+		Expect(err).To(BeNil(), "failed to repair /dev/by-id symlink and stage volume")
```
msau42: Can we also validate the link is correct?

msau42: Or do something like write data to it, and then validate the data is correct

test/e2e/tests/single_zone_e2e_test.go
davidz627: haha oops you're right. that's still pretty risky to be honest but far less so

pkg/mount-manager/device-utils.go
```diff
@@ -85,75 +95,146 @@ func (m *deviceUtils) GetDiskByIdPaths(deviceName string, partition string) []st
...
-// --action=change" for newly created "/dev/sd*" drives (exist only in
-// after set). This is workaround for Issue #7972. Once the underlying
-// issue has been resolved, this may be removed.
-func udevadmChangeToNewDrives(sdBeforeSet sets.String) error {
-	sdAfter, err := filepath.Glob(diskSDPattern)
+// VerifyDevicePath returns the first devicePath that maps to a real disk in the
+// candidate devicePaths or an empty string if none is found. If
+// /lib/udev/scsi_id exists it will attempt to fix any issues caused by missing
+// paths or mismatched devices by running a udevadm --trigger.
+func (m *deviceUtils) VerifyDevicePath(devicePaths []string, diskName string) (string, error) {
```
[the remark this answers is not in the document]
davidz627: I guess we don't really have to, but then we would need to add partition as an argument. (I can follow up but probably out of scope for this PR)

pkg/mount-manager/device-utils.go
davidz627: yeah, you're right. Since this is containerized it does make sense that it should always exist. If not it's definitely a problem.

test/e2e/tests/single_zone_e2e_test.go
davidz627: added some validation logic that goes and checks which disk is actually backing a link

pkg/mount-manager/device-utils.go
```diff
@@ -85,75 +96,166 @@ func (m *deviceUtils) GetDiskByIdPaths(deviceName string, partition string) []st
...
+	scsiIDPath := "/lib/udev_containerized/scsi_id"
+	exists, err := pathutils.Exists(pathutils.CheckFollowSymlink, scsiIDPath)
 	if err != nil {
-		return fmt.Errorf("Error filepath.Glob(\"%s\"): %v\r\n", diskSDPattern, err)
+		return "", fmt.Errorf("failed to check scsi_id existence: %v", err)
+	}
+	if !exists {
+		// No SCSI ID tool, the driver should be containerized with the tool so
+		// maybe something is wrong with the build process
+		return "", fmt.Errorf("could not find scsi_id tool at %s, no udevadm fixing is possible so we just return the best effort devicePath", scsiIDPath)
```
msau42: error message is out of date?

pkg/mount-manager/device-utils.go
```diff
@@ -85,75 +96,166 @@ func (m *deviceUtils) GetDiskByIdPaths(deviceName string, partition string) []st
...
 	}
 
-	for _, sd := range sdAfter {
-		if !sdBeforeSet.Has(sd) {
-			return udevadmChangeToDrive(sd)
+	err = wait.Poll(pollInterval, pollTimeout, func() (bool, error) {
+		var innerErr error
+
+		devicePath, innerErr = existingDevicePath(devicePaths)
```
msau42: We don't regenerate `devicePaths` on the next loop, so would this still fail?

pkg/mount-manager/device-utils.go
```diff
@@ -85,75 +96,166 @@ func (m *deviceUtils) GetDiskByIdPaths(deviceName string, partition string) []st
...
 		"trigger",
 		"--action=change",
-		fmt.Sprintf("--property-match=DEVNAME=%s", drive)).CombinedOutput()
+		fmt.Sprintf("--property-match=DEVNAME=%s", devSDX)).CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("udevadmChangeToDrive: udevadm trigger failed for drive %q with %v.", drive, err)
+		return fmt.Errorf("udevadmChangeToDrive: udevadm trigger failed for drive %q with output %s: %v.", devSDX, string(out), err)
 	}
+
+	time.Sleep(time.Second)
```
msau42: Since this is inside a Poll loop now, do we still need this sleep?

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -81,7 +85,78 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			if err != nil {
+				klog.Errorf("Failed to detach disk: %v", err)
+			}
+
+		}()
+
+		// MESS UP THE symlink
+		devicePaths := mountmanager.NewDeviceUtils().GetDiskByIdPaths(volName, "")
+		for _, devicePath := range devicePaths {
+			err = testutils.RmAll(instance, devicePath)
```
msau42: Is this still going to be potentially destructive if volume name is a prefix of an existing disk? ie, volume name is "foo" but a real disk is called "foobar"

test/e2e/utils/utils.go
```diff
@@ -195,10 +196,69 @@ func GetBlockSizeInGb(instance *remote.InstanceInfo, devicePath string) (int64,
...
+func CopyFile(instance *remote.InstanceInfo, src, dest string) error {
+	output, err := instance.SSH("cp", src, dest)
+	if err != nil {
+		return fmt.Errorf("failed to copy %s to %s. Output: %v, errror: %v", src, dest, output, err)
+	}
+	return nil
+}
+
+// ValidateLogicalLinkIsDisk takes a symlink location at "link" and finds the
+// link location - it then finds the backing PD using scsi_id and validates that
```
msau42: This is fine for now, although when NVMe comes up, this test will get outdated. I think writing data and verifying it would be more portable.

davidz627: The thing is writing data would work if the symlink was made incorrectly, so I'm not sure it validates what we care about. (Which is that the volume matches the device). Writing this check was much harder than writing data :)

I don't think network attached PD supports NVME. Only available for local ssd

pkg/mount-manager/device-utils.go
davidz627: Not sure why this is still here. Thought I removed it, fixing

pkg/mount-manager/device-utils.go
davidz627: DevicePaths are a list of all possible device paths generated from a pattern, the existingdevicepath one is checking which of those are present on disk. So this essentially "regenerate" the new device pah if it exists now

test/e2e/tests/single_zone_e2e_test.go
davidz627: I don't think so. The device paths are full absolute paths.

We're doing a:
`rm -rf /dev/disk/by-id/foo`
which wont affect:
`/dev/disk/by-id/foobar`

test/e2e/utils/utils.go
msau42: Yes, there's no NVME for PD now, I'm just saying for future proofing.

pkg/mount-manager/device-utils.go
msau42: Thanks, I understand now

PR #460, opened 2020-02-04
Add required IAM permissions to driver installation guide

### LINKED ISSUE #389: Driver bringup documentation improvements
- [x] A little guide on setting up the go repo into the correct format 
```
mkdir -p $HOME/go/src/sigs.k8s.io && cd go/src/sigs.k8s.io
```
then use git clone to clone both repos.
- [ ] A list of the permissions/roles required to run the installation scripts

davidz627:
/kind documentation

Fixes #389

Improve docs a little at a time
/assign @msau42 

```release-note
NONE
```

### DIFF
--- a/docs/kubernetes/user-guides/driver-install.md
+++ b/docs/kubernetes/user-guides/driver-install.md
@@ -3,21 +3,80 @@
 ## Install Driver
 
 1. Clone the driver to your local machine
-```
+
+```console
 $ git clone https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver $GOPATH/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 ```
 
-2. [One-time per project] Create GCP service account for the CSI driver and set required roles
+2. [One-time per project] Set up or use an existing service account:
+
+The driver requires a service account that has the following permissions and
+roles to function properly:
+
 ```
+compute.instances.get
+compute.instances.attachDisk
+compute.instances.detachDisk
+roles/compute.storageAdmin
+roles/iam.serviceAccountUser 
+```
+
+If there is a pre-existing service account with these roles for use then the
+service account key must be downloaded and made discoverable through environment
+variable
+
+```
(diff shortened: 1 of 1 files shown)


docs/kubernetes/user-guides/driver-install.md
```diff
@@ -1,22 +1,45 @@
 # Kubernetes Driver Installation Guide
 
+## GCP Permissions Required
+
+The service account performing driver installation must have at least the
+following GCP IAM permissions:
+
+```
+resourcemanager.projects.getIamPolicy
```
msau42: This is to run the initial setup script that creates a new service account right?

If a user wants to import/use an existing service account, then the set of permissions they need would be different.

docs/kubernetes/user-guides/driver-install.md
```diff
@@ -1,22 +1,53 @@
...
+exist the additional permissions are required:
+
+```
+resourcemanager.projects.getIamPolicy
+resourcemanager.projects.setIamPolicy
+iam.serviceAccounts.create
+iam.serviceAccounts.delete
+```
+
+These permissions are not required if you already have a service account ready
```
msau42: Do we have instructions on how to use an existing service account and what permissions those need?

docs/kubernetes/user-guides/driver-install.md
```diff
@@ -1,22 +1,53 @@
 # Kubernetes Driver Installation Guide
 
+## GCP Permissions Required
+
+The service account running `setup-project.sh` in order to set up a service
```
msau42: The meta wording is a little confusing. How about something like "The setup-project.sh script only needs to be run once per project to generate a service account for the driver. The user or service account running this script needs the following permissions"

docs/kubernetes/user-guides/driver-install.md
davidz627: added, PTAL

docs/kubernetes/user-guides/driver-install.md
davidz627: yeah this is really hard to work. I've updated it all, PTAL

PR #462, opened 2020-02-04
Fix Resize call to return size in Gi as expected when disk is already the request size or larger

### LINKED ISSUE #433: Bug in NodeExpandVolume
It looks like we may have a bug in CSI `NodeExpandVolume` implementation:

```
FailedMount: MountVolume.MountDevice failed while expanding volume for volume "pvc-229df8db-a895-43b9-8e0b-dafa5bdf19b2" : MountVolume.NodeExpandVolume failed : Expander.NodeExpand failed to expand the volume : rpc error: code = Internal desc = ControllerExpandVolume resize requested for 6917529027641081856 but after resize volume was size 6442450944
```

Failure logs -  https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-cos-k8sbeta-serial/1198911611914424322/build-log.txt

cc @msau42

davidz627:
/kind bug
/kind failing-test
/assign @msau42 @gnufied

Fixes #433 

```release-note
Fixed bug where ControllerExpandVolume was returning incorrect size when disk was already the requested size or larger
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -620,6 +620,10 @@ func (cloud *CloudProvider) CreateSnapshot(ctx context.Context, volKey *meta.Key
 	}
 }
 
+// ResizeDisk takes in the requested disk size in bytes and returns the resized
+// size in Gi
+// TODO(#461) The whole driver could benefit from standardized usage of the
+// k8s.io/apimachinery/quantity package for better size handling
 func (cloud *CloudProvider) ResizeDisk(ctx context.Context, volKey *meta.Key, requestBytes int64) (int64, error) {
 	klog.V(5).Infof("Resizing disk %v to size %v", volKey, requestBytes)
 	cloudDisk, err := cloud.GetDisk(ctx, volKey)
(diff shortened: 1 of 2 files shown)


davidz627:
Also the error logs were totally incorrect so I just fixed that as I was going

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -630,9 +634,10 @@ func (cloud *CloudProvider) ResizeDisk(ctx context.Context, volKey *meta.Key, re
 	sizeGb := cloudDisk.GetSizeGb()
 	requestGb := common.BytesToGb(requestBytes)
 
-	// If disk is already of size equal or greater than requested size, we simply return
+	// If disk is already of size equal or greater than requested size, we
+	// simply return the found size
 	if sizeGb >= requestGb {
-		return requestBytes, nil
+		return sizeGb, nil
```
msau42: @gnufied also identified this as one of the problematic driver responses to resize. We should be returning the actual size, not the requested size.

msau42: Nm it's too early in the morning to read properly :-)

msau42: Can we add a unit test for this case that verifies we return actual size instead of request?

davidz627: it's really hard to unit test this package right now as there was no easy way to fake out the actual cloud provider layer. So this "is" the cloud provider layer that I hacked together that we use as an interface to fake out. Suggestions welcome. It's in the backlog to switch over to the autogenerated cloud provider client layer that we switched over to in Kubernetes a while back - that provided the ability to fake out calls and unit test these types of things effectively.

PR #465, opened 2020-02-05
Thread contexts into all appropriate functions in non-test code

### LINKED ISSUE #144: CreateVolume call failing on trying to find the disk
First pvc got provisioned alright but second one is stuck on:

```
{"log":"E1030 16:47:30.481739       1 utils.go:55] GRPC error: rpc error: code = Internal desc = CreateVolume unknown get disk error when validating: context canceled\n","stream":"stderr","time":"2018-10-30T16:47:30.482182878Z"}
```

This is when code is trying to fetch the disk to verify if the disk already exists.   I am not sure what is causing the context to be cancelled. The overall pod is failing because of snapshot controller not able to start, but I am not sure if this is related to that.

cc @msau42

davidz627:
/assign @msau42 
/kind cleanup

We thread contexts through so that cancelled contexts will cancel cloud provider calls in the driver.

```release-note
NONE
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -15,6 +15,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"math/rand"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -63,8 +64,11 @@ func handle() {
 
 	gceDriver := driver.GetGCEDriver()
 
-	//Initialize GCE Driver (Move setup to main?)
-	cloudProvider, err := gce.CreateCloudProvider(vendorVersion, *gceConfigFilePath)
+	//Initialize GCE Driver
+	ctx, cancel := context.WithCancel(context.Background())
```
msau42: I'm not sure if this will solve the issue? With this change, will we end up canceling all outstanding operations whenever the driver restarts?

Have you tried to repro the issue by setting a short timeout?

davidz627: unliked the issue, this PR doesn't have anything to do with it other than the fact they both involve contexts :)

PR #467, opened 2020-02-06
Implement ValidateVolumeCapabilities and refactor parameter handling for more comprehensive validation of existing disks in all cloud calls

### LINKED ISSUE #162: Implement ValidateVolumeCapabilities
Currently this function is unimplemented and just returns a "Unimplemented" message.

This should be done ASAP
/cc @saad-ali

### LINKED ISSUE #384: Validate volume capabilities for file system type and other non-access mode things
/assign

davidz627:
TODO: 
- [x] Add some unit tests for new functions :)

/kind cleanup
/kind feature
/assign @msau42 @misterikkit 

Partially does #384 (Still need to do `Node` side validation in a seperate PR)
Fixes #162 

```release-note
ValidateVolumeCapabilities validates that the given volume conforms to all capabilities in the request. Validation of existing volumes during inserts also improved to check all parameters.
```

### DIFF
--- a/pkg/common/parameters.go
+++ b/pkg/common/parameters.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	ParameterKeyType                 = "type"
+	ParameterKeyReplicationType      = "replication-type"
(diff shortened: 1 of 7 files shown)


pkg/common/parameters.go
```diff
@@ -0,0 +1,75 @@
...
+		DiskEncryptionKMSKey: "",                  // Default
+	}
+	for k, v := range parameters {
+		if k == "csiProvisionerSecretName" || k == "csiProvisionerSecretNamespace" {
+			// These are hardcoded secrets keys required to function but not needed by GCE PD
+			continue
+		}
+		switch strings.ToLower(k) {
+		case ParameterKeyType:
+			if v != "" {
```
msau42: should it be considered an error if a user specified a key without a value?

davidz627: in-tree PD just treats that case as the default. I tried setting `type: ""` and it just created a `pd-standard` without complaining

davidz627:
#415 seems to have stalled and already has conflicts

davidz627:
@msau42 added unit tests, PTAL

pkg/common/parameters.go
```diff
@@ -0,0 +1,75 @@
...
+				p.DiskType = v
```
msau42: do we want toLower() here?

davidz627: hmmmm for in-tree we `toLower()` replication type but literally none of the other parameters. It feels like it doesn't matter so much but can definitely be added without issue

PR #479, opened 2020-02-26
Validate that existing mount point is compatible with volume and capabilities when it already exists for Stage and Publish

### LINKED ISSUE #253: Validate VolumeCapability matches with "AlreadyExist"ing volume in Publish/Stage calls
We do static validation of volume capability now. However, if an existing volume is found we must validate that the volume capabilities of the existing volume match with the requested capabilities, otherwise we must return an "ALREADY_EXIST" error

davidz627:
/kind feature
/assign @msau42 

Fixes #253 

```release-note
Staging or Publishing to existing mountpoint's are now return an AlreadyExists error if the volume or capabilities don't match, success otherwise
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -15,8 +15,10 @@ limitations under the License.
 package gceGCEDriver
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -59,6 +61,8 @@ var _ csi.NodeServer = &GCENodeServer{}
 const (
 	volumeLimitSmall int64 = 15
 	volumeLimitBig   int64 = 127
+
+	defaultFSType string = "ext4"
 )
 
 func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
@@ -95,13 +99,20 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 		return nil, status.Error(codes.Internal, fmt.Sprintf("cannot validate mount point: %s %v", targetPath, err))
 	}
 	if !notMnt {
-		// TODO(#95): check if mount is compatible. Return OK if it is, or appropriate error.
-		/*
-			1) Target Path MUST be the vol referenced by vol ID
-			2) TODO(#253): Check volume capability matches for ALREADY_EXISTS
-			3) Readonly MUST match
+		// Validate that the existing volume is compatible
+		partition := ""
+		if part, ok := req.GetVolumeContext()[common.VolumeAttributePartition]; ok {
+			partition = part
+		}
+		devicePath, err := ns.getDevicePath(volumeID, partition)
+		if err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Error when getting device path: %v", err))
+		}
 
-		*/
+		if err := ns.volumeCompatible(targetPath, devicePath, volumeCapability); err != nil {
+			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("Mount point %s already exists but found to be incompatible: %v", stagingTargetPath, err))
+		}
+
 		klog.V(4).Infof("NodePublishVolume succeeded on volume %v to %s, mount already exists.", volumeID, targetPath)
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
@@ -118,8 +129,7 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 		if mnt.FsType != "" {
 			fstype = mnt.FsType
 		} else {
-			// Default fstype is ext4
-			fstype = "ext4"
+			fstype = defaultFSType
 		}
 
 		klog.V(4).Infof("NodePublishVolume with filesystem %s", fstype)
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -327,6 +332,41 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
 
+func (ns *GCENodeServer) volumeCompatible(mountedPath, devicePath string, volumeCapability *csi.VolumeCapability) error {
+	// Part 1: Check that volume mounted at mountedPath is the same as one at
+	// devicePath
+	devicePathDev, err := filepath.EvalSymlinks(devicePath)
+	if err != nil {
+		return fmt.Errorf("failed to find backing disk for devicePath %s: %v", devicePath, err)
+	}
+
+	mountedPathDevBytes, err := ns.Mounter.Exec.Command("sh", "-c", fmt.Sprintf("df -P | awk '$6==\"%s\"{print $1}'", mountedPath)).CombinedOutput()
```
msau42: Can you add a comment that shows the syntax of the line returned and the output post processing?

pkg/gce-pd-csi-driver/node.go
```diff
@@ -95,13 +96,20 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
...
+		partition := ""
+		if part, ok := req.GetVolumeContext()[common.VolumeAttributePartition]; ok {
+			partition = part
+		}
+		devicePath, err := ns.getDevicePath(volumeID, partition)
+		if err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Error when getting device path: %v", err))
+		}
+
+		if err := ns.volumeCompatible(stagingTargetPath, devicePath, volumeCapability); err != nil {
```
msau42: This assumes that `targetPath` is bind mounted correctly to `devicePath`

pkg/gce-pd-csi-driver/node.go
```diff
@@ -327,6 +332,41 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
+		return fmt.Errorf("devices at paths %s and %s were not the same, got %s %s respectively", devicePath, mountedPath, devicePathDev, mountedPathDev)
+	}
+
+	// Part 2: Check volumeCapability format compatibility
+	format, err := ns.Mounter.GetDiskFormat(devicePath)
+	if err != nil {
+		return fmt.Errorf("failed to get the format of disk %s: %v", devicePath, err)
+	}
+	if mnt := volumeCapability.GetMount(); mnt != nil {
+		if mnt.FsType != format {
```
msau42: CSI spec allows mnt.FsType to be empty, and in that case we should default.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -327,6 +332,41 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
+			return fmt.Errorf("device at %s has format %s but volume capability requires %s", devicePath, format, mnt.FsType)
+		}
+	}
+
+	// If the volume capability request is type "Block" we don't care what
```
msau42: Can we skip the GetDiskFormat call entirely in this case?

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -159,6 +161,147 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+				klog.Errorf("Failed to unstage volume: %v", err)
+			}
+			fp := filepath.Join("/tmp/", volNameA)
+			err = testutils.RmAll(instance, fp)
+			if err != nil {
+				klog.Errorf("Failed to rm file path %s: %v", fp, err)
+			}
+		}()
+
+		// Assert: Stage Disk B to Disk A position and fail even both as EXT4
```
msau42: Can we run each scenario as a separate test case?

pkg/gce-pd-csi-driver/node.go
davidz627: what do you mean by this? We are checking whether it is bind mounted correctly. If it is not it will throw an AlreadyExists error which is the expected behavior

test/e2e/tests/single_zone_e2e_test.go
davidz627: There's a lot of boilerplate setup for "later assertions" as asserting things for publish depends on doing a attach and stage. I actually think this is a lot cleaner especially since i've documented it quite thoroughly. In the passing case we're happy, in the failing case I'm not really concerned about "knock on errors" since a failure is a failure and you can debug based on the first error seen.

davidz627:
@msau42 resolved comments. This is going to be fairly difficult to unit test given its almost complete dependence on actual filesystem behavior

pkg/gce-pd-csi-driver/node.go
msau42: I mean that it is checking that `devicePath` (/dev/blah) is mounted to `stagingTargetPath` (NodeStage path), but not necessarily that `stagingTargetPath` is bind mounted to `targetPath` (NodePublish path).

Regardless, this is still a good check to have, it's just not actually verifying the Publish path.

davidz627: I see, I think in NodePublish we check that `/dev/sdx` is mounted to `publishTargetPath`. But you're right in that we never explicitly check that `stagingTargetPath` is bindmounted to `publishTargetPath`.

Edit: I just noted we check `stagingTargetPath` in `NodePublish` as well, it should actually be `targetPath` to make my above statement too.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -327,6 +333,58 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
+	devicePathDev, err := filepath.EvalSymlinks(devicePath)
+	if err != nil {
+		return fmt.Errorf("failed to find backing disk for devicePath %s: %v", devicePath, err)
+	}
+	// df -P returns with the rows containing the backing disk and the location
+	// of the mount. awk proccesses the result by getting the line with the
+	// location of the mount we're looking for and printing out the backing
+	// disk. the resulting output should be a path to a device such as
+	// "/dev/sda"
+	mountedPathDevBytes, err := ns.Mounter.Exec.Command("sh", "-c", fmt.Sprintf("df -P | awk '$6==\"%s\"{print $1}'", mountedPath)).CombinedOutput()
```
msau42: I'm not sure if df works for bind mounts. Mounter I believe has GetMountRefs() which I think handles bind mounts

Issue #387, opened 2019-09-18
Add metadata information in tag on PD


davidz627:
info like namespace and pvc name (check with feature parity for in-tree plugin)

davidz627:
In-tree plugin provisions with `Description` field example:
```
Description
{"kubernetes.io/created-for/pv/name":"pvc-6a10b78e-4cae-4c9e-8dd2-1cb2c60211e1","kubernetes.io/created-for/pvc/name":"test-disk","kubernetes.io/created-for/pvc/namespace":"default"}
```

No such thing as a "tag" AFAIK. Do people really rely on json description text? I guess we should continue this for backwards compatibility reasons. Considering adding some more description fields to specify the driver name that provisioned to volume.

/cc @msau42

msau42:
What about "labels"? https://cloud.google.com/kubernetes-engine/docs/how-to/creating-managing-labels

davidz627:
IIRK labels are new and basically what we were trying to approximate with the description field before. Labels are not used by in-tree currently. They are empty on my test PD. 

IIUC description should stay the same because of backwards compatibility (who is actually depending on this functionality...?). Do we even have any expectations for backwards compatibility for a freeform text field meant for human-readable descriptions? 

We could duplicate whats in the description into the labels but that's almost uglier than only having them in the description. However, I would prefer to push to put this information *only* in the labels and break the descriptions field.

msau42:
It has the possibility to break users once migration is on though.

davidz627:
Depends on: https://github.com/kubernetes-csi/external-provisioner/issues/370

davidz627:
depends on: https://github.com/kubernetes-csi/external-provisioner/pull/399

verult:
kubernetes-csi/external-provisioner#399 now available in v1.6.0

shay-berman:
what is the status of this ticket?
I would like to know when CSI driver will be GA in GKE?

PR #487, opened 2020-04-08
Base deployment: enabled metric endpoint for all controllers

verult:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```


deploy/kubernetes/overlays/prow-gke-release-staging-rc/enable_provisioner_extra_metadata.yaml
```diff
@@ -0,0 +1,4 @@
+# for external-provisioner
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: "--extra-create-metadata"
```
msau42, line 4: It's not enough to just enable the flag. We actually need to plumb the information that's passed in to PD creation.

verult: Right. Plumbing the information is a longer term task, and for the time being I wanted to first enable this flag and let it soak.

verult:
The driver validates `CreateVolume()` params and rejects any call with unrecognized params, so the staging RC spec can only be updated with a PD CSI driver release that contains the fix for https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/387. Closing.

verult:
Reopening and changing the PR to only include the metric change.

verult:
Superseded by https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/517

PR #489, opened 2020-04-15
Add build rule for Windows container

jingxu97:
Add build rule for windows container. Here we try to enable buildx for
buiding windows container image on a linux VM

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Adds new build rules using docker buildx feature for allowing to build windows container image on either windows or linux node.
```

### DIFF
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,0 +1,9 @@
+ARG BASE_IMAGE
+ARG BASE_IMAGE_TAG
+
+FROM mcr.microsoft.com/windows/${BASE_IMAGE}:${BASE_IMAGE_TAG}
+LABEL description="PD CSI driver"
+COPY bin/gce-pd-csi-driver.exe /gce-pd-csi-driver.exe
+
+USER ContainerAdministrator
+ENTRYPOINT ["/gce-pd-csi-driver.exe"]
--- a/Makefile
+++ b/Makefile
@@ -14,35 +14,46 @@
 
 # Args:
 # GCE_PD_CSI_STAGING_IMAGE: Staging image repository
-
+REV=$(shell git describe --long --tags --match='v*' --dirty 2>/dev/null || git rev-list -n1 HEAD)
+ifdef GCE_PD_CSI_STAGING_VERSION
+	STAGINGVERSION=${GCE_PD_CSI_STAGING_VERSION}
+else
+	STAGINGVERSION=${REV}
+endif
 STAGINGIMAGE=${GCE_PD_CSI_STAGING_IMAGE}
-STAGINGVERSION=${GCE_PD_CSI_STAGING_VERSION}
 DRIVERBINARY=gce-pd-csi-driver
 DRIVERWINDOWSBINARY=${DRIVERBINARY}.exe
 
 all: gce-pd-driver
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -28,12 +29,9 @@ ifndef GCE_PD_CSI_STAGING_VERSION
...
 	go build -ldflags "-X main.vendorVersion=${STAGINGVERSION}" -o bin/${DRIVERBINARY} ./cmd/
 
-build-windows:
+gce-pd-driver-windows:
 	mkdir -p bin
-ifndef GCE_PD_CSI_STAGING_VERSION
-	$(error "Must set environment variable GCE_PD_CSI_STAGING_VERSION to staging version")
-endif
-	GOOS=windows go build -ldflags "-X main.vendorVersion=${STAGINGVERSION}" -o bin/${DRIVERWINDOWSBINARY} ./cmd/
+	GOOS=windows go build -ldflags -X=main.vendorVersion=$(REV) -o bin/${DRIVERWINDOWSBINARY} ./cmd/
```
msau42, line 34: How come we use REV instead of STAGINGVERSION like the rest of the rules?

Makefile
```diff
@@ -44,6 +42,16 @@ ifndef GCE_PD_CSI_STAGING_VERSION
 endif
 	docker build --build-arg TAG=$(STAGINGVERSION) -t $(STAGINGIMAGE):$(STAGINGVERSION) .
 
+build-push-windows-container:
```
msau42, line 45: Does this invoke gce-pd-driver-windows first?

jingxu97: no, it just does the docker build

msau42: Maybe remove the "build" from the name?

jingxu97: here build means docker build.
I see it is used in build the linux container?

Makefile
jingxu97: we changed it to use a default value getting from git

msau42:
/lgtm
/approve
Can you also add a release note

Dockerfile.Windows
```diff
@@ -0,0 +1,9 @@
+ARG BASE_IMAGE
+ARG IMAGE_TAG
```
msau42, line 2: A better name may be BASE_IMAGE_TAG

Dockerfile.Windows
```diff
@@ -0,0 +1,9 @@
...
+FROM mcr.microsoft.com/windows/${BASE_IMAGE}:${IMAGE_TAG}
+ENV CSI_PROXY_PIPE "\\\\.\\pipe\\csipipe"
```
msau42, line 4: Why do we define the ENV here instead of in the Daemonset spec? Doesn't the pipe path change depending on which API endpint we're talking to?

Dockerfile.Windows
jingxu97: ack and fixed

Dockerfile.Windows
jingxu97: right, it is not needed here. removed it

PR #503, opened 2020-05-11
Provide optional flag and add a new default volumesnapshotclass to run

### LINKED ISSUE #457: Update deployments to include new snapshot controller (for v1beta1 APIs), start running tests, and update docs
Looks like some new snapshot APIs went Beta in k8s `1.17`. The corresponding `external-snapshotter` is also available and tested with the PD Driver. However, the stable deployment in this repo still does not contain the snapshotter, tests are not run by default, and the documentation for snapshotting may be out of date.

saikat-royc:
PD CSI driver snapshot tests.

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind bug
> /kind flake

Summary of Changes:
1. Added a snapshot-class option to pass volumesnapshotclass.
2. Added a default volume snapshot class definition, which uses PD CSI as a snapshot driver.
3. Remove “Feature:” tag from skipString, since snapshot tests have a [Feature:VolumeSnapshotDataSource] tag.
4. Prevent snapshot tests from running pre 1.17 kubernetes version.
5. Allow snapshot tests run only in release-staging-head and release-staging-rc overlays, since the snapshotter is not yet enabled in stable.

Fixes #457 

```release-note
Enable snapshot tests for PD CSI driver for release-staging-head and release-staging-rc overlays.
```


test/k8s-integration/driver-config.go
```diff
@@ -72,9 +71,18 @@ func generateDriverConfigFile(pkgDir, storageClassFile, deploymentStrat string)
 		return "", fmt.Errorf("got unknown deployment strat %s, expected gce or gke", deploymentStrat)
 	}
 
+	var absSnapshotClassFilePath string
+	// If snapshot class is passed in as argument, include snapshot specific driver capabiltiites.
+	if snapshotClassFile != "" {
+		caps = append(caps, "snapshotDataSource", "pvcDataSource")
```
msau42, line 77: pvcDataSource refers to the "cloning" feature, which pd driver does not support.

test/k8s-integration/main.go
```diff
@@ -333,6 +333,8 @@ func generateTestSkip(normalizedVersion string) string {
 		// bug-fix introduced in 1.17
 		// (https://github.com/kubernetes/kubernetes/pull/81163)
 		skipString = skipString + "|volumeMode\\sshould\\snot\\smount\\s/\\smap\\sunused\\svolumes\\sin\\sa\\spod"
```
msau42, line 335: Side issue: This bug fix is actually in 1.16, so we can enable this test case in 1.16.

test/run-k8s-integration.sh
```diff
@@ -44,4 +44,10 @@ else
   base_cmd="${base_cmd} --gce-region=${gce_region}"
 fi
 
+# TODO: When beta csi snapshotter sidecar is promoted to stable, add stable
```
msau42, line 47: I think we can eventually enable it on all overlays so we don't need to have an overlay check.

test/k8s-integration/driver-config.go
saikat-royc: Ack. Will fix and retest.

test/k8s-integration/main.go
saikat-royc: Should we track this in a different issue?

test/run-k8s-integration.sh
saikat-royc: If we remove this check right now --snapshotclass-file will be passed as input to main.go, and stable driver running against >=1.17 will be running snapshot tests and would fail. hence I put a todo.

test/k8s-integration/main.go
msau42: yes can be tracked separately

test/run-k8s-integration.sh
msau42: yes, I was referring to the todo comment that says we should add the other overlays to the check. I'm thinking we can do the opposite and remove the check once we support snapshots in all the overlays.

test/k8s-integration/driver-config.go
msau42: Enabling the test should have caused a test failure since the driver doesn't support it. I'm curious to understand why it passed when you were testing it?

saikat-royc: I was testing only focussed snapshot tests using "External.Storage.*snapshot" instead of "External.Storage", hence only the snapshot tests were run. 
I will now run with just "External.Storage" and see the test results.

test/run-k8s-integration.sh
saikat-royc: Ah got it. Yes if we plan to put it in all, we can remove the check completely eventually.

saikat-royc:
issue with commit will re-submit again.

PR #509, opened 2020-05-18
Migrate instructions

mattcary:
/kind documentation

Add guidance on migrating from alpha snapshot version

```release-note
NONE
```


mattcary:
#457

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +122,141 @@
...
+  * snapshot-controller:v2.0.1
+  * csi-provisioner:v1.5.0-gke.0
+  * gke.gcr.io/csi-attacher:v2.1.1-gke.0
+  * gke.gcr.io/csi-resizer:v0.4.0-gke.0
+  * gke.gcr.io/csi-snapshotter:v2.1.1-gke.0
+  * gke.gcr.io/gcp-compute-persistent-disk-csi-driver:v0.7.0-gke.0
+
+#### Overview
+
+The idea is to provision a `VolumeSnapshotContents` in your new or upgraded
```
msau42, line 152: clarify the new object is the beta version

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +122,141 @@
...
+cluster with the PD snapshot handle from the alpha snapshot, then bind that to a
+new `VolumeSnapshot`.
```
msau42, line 154: beta

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +122,141 @@
...
+
+If you are able to set the `deletionPolicy` of any existing
+`VolumeSnapshotContents` to `Retain`, or have already thus set the
+d`deletionPolicy` of the creating `VolumeSnapshotClass`, then deleting the
```
msau42, line 158: I don't remember if deletion of snapshots looks at the field from the VolumeSnapshotClass.

Maybe it's safer to just update all of them to Retain.

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +122,141 @@
...
+  * gke.gcr.io/csi-snapshotter:v2.1.1-gke.0
+  * gke.gcr.io/gcp-compute-persistent-disk-csi-driver:v0.7.0-gke.0
+
+#### Overview
+
+The idea is to provision a `VolumeSnapshotContents` in your new or upgraded
+cluster with the PD snapshot handle from the alpha snapshot, then bind that to a
+new `VolumeSnapshot`.
+
+If you are able to set the `deletionPolicy` of any existing
```
msau42, line 156: Throughout the doc, it would be good to clarify if you're referring to the alpha version or beta version of the object.

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +122,141 @@
...
+d`deletionPolicy` of the creating `VolumeSnapshotClass`, then deleting the
+snapshot CRs will not delete the underlying PD snapshot. This can provision a
+`VolumeSnapshotContents` as described below.
+
+If you are creating a new cluster rather than upgrading an existing one, the
+method below could also be used before deleting the alpha cluster and/or alpha
+PD CSI driver deployment. This may be a safer option if you are not sure about
+the deletion policy of your alpha snapshot.
+
+#### Steps to restore from an existing snapshot.
```
msau42, line 167: These steps are useful even outside the context of trying to migrate from alpha to beta. Maybe this should be pulled out of the migration section and made more top level. Then in the migration section, you can refer to the "import an existing snapshot" section.

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +122,141 @@
...
+           apiGroup: snapshot.storage.k8s.io
+     EOF
+     ```
+  1. Finally, create a pod referring to the `PersistentVolumeClaim`. The PD CSI
+     driver will provision a `PersistentVolume` and populate it from the
+     snapshot. The pod from `examples/kubernetes/snapshot/restored-pod.yaml` is
+     set to use the PVC created above. After `kubectl apply`'ing the pod, run
+     `kubectl exec restored-pod -- ls -l /demo/data/` to confirm that the
+     snapshot has been restored correctly.
+
```
msau42, line 238: You also need to uninstall the alpha CRD after you're done?

docs/kubernetes/user-guides/snapshots.md
mattcary: Done

docs/kubernetes/user-guides/snapshots.md
mattcary: done

docs/kubernetes/user-guides/snapshots.md
mattcary: Done. It worked for me with just changing the contents but you have a good point.

docs/kubernetes/user-guides/snapshots.md
mattcary: done

docs/kubernetes/user-guides/snapshots.md
mattcary: Yeah, that's a good point. I was a little surprised myself that I couldn't find any guide on manually provisioning snapshot contents.

docs/kubernetes/user-guides/snapshots.md
mattcary: Clarified by changing the intro above to say that this is only operating on the beta cluster.

docs/kubernetes/user-guides/snapshots.md
mattcary: I've done this, creating a migration section and referring to that from the migration tips section.

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
     lost+found
     sample-file.txt
     ```
+
+### Manually Provisioning Snapshot Contents.
```
msau42, line 129: What do you think of calling it something like "Import a pre-existing snapshot"? I think this description aligns better with what a user would want to accomplish without using Kubernetes terms.

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
...
+     driver will provision a `PersistentVolume` and populate it from the
+     snapshot. The pod from `examples/kubernetes/snapshot/restored-pod.yaml` is
+     set to use the PVC created above. After `kubectl apply`'ing the pod, run
+     `kubectl exec restored-pod -- ls -l /demo/data/` to confirm that the
+     snapshot has been restored correctly.
+
+#### Troubleshooting
+
+If the `VolumeSnapshot`, `VolumeSnapshotContents` and `PersistentVolumeClaim`
+are not all mutually syncrhonized, the pod will not start. Try `kubectl describe
```
msau42, line 207: typo: synchronized

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
...
+kubectl logs -n gce-pd-csi-driver csi-gce-pd-controller-0 -c csi-provisioner | tail
+```
+
+### Tips on Migrating from Alpha Snapshots
+
+The api version has changed between the alpha and beta releases of the CSI
+snapshotter. In many cases, snapshots created with the alpha will not be able to
+be used with a beta driver. In some cases, they can be migrated.
+
+*You may lose all data in your snapshot!*
```
msau42, line 232: Bold and add "Attention" similar to the other notes

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
...
+snapshot!*
+
+These instructions happened to work with a particular GKE configuration. Because
+of the variety of alpha deployments, it is not possible to verify these steps
+will work all of the time. The following has been tested with a GKE cluster of
+master version 1.17.5-gke.0 using the following image versions for the
+snapshot-controller and PD CSI driver:
+
+  * snapshot-controller:v2.0.1
+  * csi-provisioner:v1.5.0-gke.0
```
msau42, line 244: gke.gcr.io?

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
...
+  * gke.gcr.io/csi-attacher:v2.1.1-gke.0
+  * gke.gcr.io/csi-resizer:v0.4.0-gke.0
+  * gke.gcr.io/csi-snapshotter:v2.1.1-gke.0
+  * gke.gcr.io/gcp-compute-persistent-disk-csi-driver:v0.7.0-gke.0
+
+#### Migrating by Restoring from a Manually Provisioned Snapshot
+
+The idea is to provision a `VolumeSnapshotContents` in your new or upgraded beta
```
msau42, line 252: clarify this is a beta VolumeSnapshotContent

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
...
+`Retain`, then deleting the snapshot CRs will not delete the underlying PD
+snapshot.
+
+If you are creating a new cluster rather than upgrading an existing one, the
+method below could also be used before deleting the alpha cluster and/or alpha
+PD CSI driver deployment. This may be a safer option if you are not sure about
+the deletion policy of your alpha snapshot.
+
+After confirming that the PD snapshot is available, restore it using the beta PD
+CSI driver as described above in "Manually Provisioning Snapshot Contents".  The
```
msau42, line 267: Can you make this a relative link?

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,150 @@
...
+restoration should in a cluster with the beta PD CSI driver installed. At not
```
msau42, line 268: typo: At no point

docs/kubernetes/user-guides/snapshots.md
msau42, line 252: I think this actually won't work if you upgrade a cluster right? Because the alpha CRD will be still be installed, preventing the beta CRD from being installed. So creation of the beta objects won't succeed until the alpha CRD is deleted.

docs/kubernetes/user-guides/snapshots.md
mattcary: Done

docs/kubernetes/user-guides/snapshots.md
mattcary: done

docs/kubernetes/user-guides/snapshots.md
mattcary: Done

docs/kubernetes/user-guides/snapshots.md
mattcary: Updated

docs/kubernetes/user-guides/snapshots.md
mattcary: done

docs/kubernetes/user-guides/snapshots.md
mattcary: I think so...

docs/kubernetes/user-guides/snapshots.md
mattcary: done

docs/kubernetes/user-guides/snapshots.md
mattcary: I'll clarify that an upgrade means the alpha components are removed. 

I don't want to get in to uninstalling everything; I tried it and it's a major PITA due to all the finalizers cross-linked between everything. It's really easy to get into a state where you have to patch out finalizers manually just to get resources to go away. Not to mention the problems of accidentally trying to delete CRDs while there are still resources around (eg, because of the finalizers).

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,153 @@
...
+snapshot.
+
+If you are creating a new cluster rather than upgrading an existing one, the
+method below could also be used before deleting the alpha cluster and/or alpha
+PD CSI driver deployment. This may be a safer option if you are not sure about
+the deletion policy of your alpha snapshot.
+
+After confirming that the PD snapshot is available, restore it using the beta PD
+CSI driver as described above in [Import a Pre-Existing
+Snapshot](#import-a-pre-existing-snapshot). The restoration should in a cluster
```
msau42, line 271: typo: should be

docs/kubernetes/user-guides/snapshots.md
```diff
@@ -122,3 +125,153 @@
...
+  * gke.gcr.io/gcp-compute-persistent-disk-csi-driver:v0.7.0-gke.0
+
+#### Migrating by Restoring from a Manually Provisioned Snapshot
+
+The idea is to provision a beta `VolumeSnapshotContents` in your new or upgraded
+beta PD CSI driver cluster with the PD snapshot handle from the alpha snapshot,
+then bind that to a new beta `VolumeSnapshot`.
+
+Note that if using the same upgraded cluster where the alpha driver was
+installed, the alpha driver must be completely removed, including its CRDs.
```
msau42, line 257: Would it be useful to note somewhere here that uninstalling the CRD will delete all the CR objects, so it's best to save all the snapshot handles before doing that?

msau42: I guess you also should save the information about what workload it was associated with too

mattcary: How does that sound?

FWIW I wasn't actually able to uninstall the driver & CRDs without wedging things. That's why I added the advice about spinning up a new cluster.

msau42: If you weren't able to get it to work, then maybe let's just say in the docs that you'll need to migrate to a new cluster. The advice about saving the objects from the old cluster still stand.

mattcary: OK, I've made it slightly stronger.

PR #510, opened 2020-05-19
Add windows driver installation support

jingxu97:
This PR adds windows driver support. It adds a windows base dir to
install base yaml files. It also adds a windows alpha kustomization
file. To install driver for windows, first set env
GCE_PD_DRIVER_VERSION=windows/alpha and run
deploy/kubernetes/deploy-driver-win.sh script.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


deploy/kubernetes/base_windows/controller.yaml
```diff
@@ -0,0 +1,82 @@
+kind: StatefulSet
```
msau42, line 1: Since almost all the specs are the same except for the paths and container images, can we use the normal "base" specs, and just have a windows overlay that patches the paths and images?

That way we have one less place to make changes to the spec.

jingxu97: I was thinking about this option. My concern was when we make some new changes that only apply to Linux, not Windows, we might break Windows installation. Now checking the "base" part, it seems not related to windows, so maybe there is no concern. What do you think?

msau42: Agree, I'm not sure I see anything very specific to Linux or Windows besides the paths and images.

jingxu97:
updated with the same base. This way is kind of tricky with using $patch: delete. 
@msau42 PTAL

deploy/kubernetes/overlays/windows/alpha/gcepd.yaml
```diff
@@ -0,0 +1,82 @@
...
+            - name: udev-rules-lib
+              mountPath: /lib/udev
+              $patch: delete
+            - name: udev-socket
+              mountPath: /run/udev
+              $patch: delete
+            - name: sys
+              mountPath: /sys
+              $patch: delete
+      nodeSelector:
```
msau42, line 47: is there a node selector for linux that needs to be deleted? or do we need to add a linux node selector?

deploy/kubernetes/overlays/windows/alpha/noderegistrar.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: csi-gce-pd-node
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-driver-registrar
+          args:
+            - --v=5
+            - --csi-address=unix://C:\\csi\\csi.sock
+            - --kubelet-registration-path=C:\\var\\lib\\kubelet\\plugins\\pd.csi.storage.gke.io\\csi.sock
```
msau42, line 13: The linux paths don't need to be removed?

deploy/kubernetes/overlays/windows/alpha/gcepd.yaml
```diff
@@ -0,0 +1,82 @@
...
+            - name: csi-proxy-disk-pipe
+              mountPath: \\.\pipe\csi-proxy-disk-v1alpha1
+            - name: csi-proxy-volume-pipe
+              mountPath: \\.\pipe\csi-proxy-volume-v1alpha1
+            - name: csi-proxy-filesystem-pipe
+              mountPath: \\.\pipe\csi-proxy-filesystem-v1alpha1
+            - name: kubelet-dir
+              mountPath: /var/lib/kubelet
+              mountPropagation: "Bidirectional"
+              $patch: delete
```
msau42, line 28: Another possibility we could do so you don't need to delete all the linux paths, is:
1. Have the base yamls not include any linux specific stuff
2. Create linux overlay based on base to add the linux specifics
3. Base all the existing overlays on the linux overlay

Then Windows can build on top of the base overlay.

What do you think of that approach?

deploy/kubernetes/overlays/windows/alpha/gcepd.yaml
jingxu97: no, the base set os to linux, here it will changed to windows

deploy/kubernetes/overlays/windows/alpha/noderegistrar.yaml
jingxu97: this is for windows path, I think.

deploy/kubernetes/overlays/windows/alpha/gcepd.yaml
jingxu97: I was thinking this too, and estimate there will be more changes and testing involved, so just go ahead with trying  this one first. 
Comparing these two, I do think the way you mentioned looks more clean. Let me try to update it with this approach..

deploy/kubernetes/base/node.yaml
```diff
@@ -15,21 +15,14 @@ spec:
       # Host network must be used for interaction with Workload Identity in GKE
       # since it replaces GCE Metadata Server with GKE Metadata Server. Remove
       # this requirement when issue is resolved and before any exposure of
-      # metrics ports.
-      hostNetwork: true
+      # metrics ports. But hostNetwork is not working for Windodws, might be an issue
```
msau42, line 18: typo: windows

Do we support workload identity on Windows? If so, then not being on host network may be problematic.

deploy/kubernetes/base/node.yaml
```diff
@@ -15,21 +15,14 @@ spec:
...
+      # when deploying on GKE windows node. See related issue https://buganizer.corp.google.com/issues/156382455
```
msau42, line 19: let's not reference internal issues here

deploy/kubernetes/base/node.yaml
```diff
@@ -41,34 +34,11 @@ spec:
             - name: registration-dir
               mountPath: /registration
         - name: gce-pd-driver
-          securityContext:
-            privileged: true
-          # Don't change base image without changing pdImagePlaceholder in
```
msau42, line 46: Keep this comment

deploy/kubernetes/deploy-driver.sh
```diff
@@ -51,7 +52,7 @@ function check_service_account()
 	# Grepping for a line with client email returning anything quoted after the colon
 	readonly IAM_NAME=$(grep -Po '"client_email": *\K"[^"]*"' ${GCE_PD_SA_DIR}/cloud-sa.json | tr -d '"')
 	readonly PROJECT=$(grep -Po '"project_id": *\K"[^"]*"' ${GCE_PD_SA_DIR}/cloud-sa.json | tr -d '"')
-	readonly GOTTEN_BIND_ROLES=$(gcloud projects get-iam-policy ${PROJECT} --flatten="bindings[].members" --format='table(bindings.role)' --filter="bindings.members:${IAM_NAME}")
+	readonly GOTTEN_BIND_ROLES=$(gcloud projects get-iam-policy $PROJECT --flatten="bindings[].members" --format='table(bindings.role)' --filter="bindings.members:${IAM_NAME}")
```
msau42, line 55: nit: keep the brackets for consistency with the other usage in this file

deploy/kubernetes/overlays/linux/base/noderegistrar.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-driver-registrar
+          args:
+            - "--v=5"
+            - "--csi-address=/csi/csi.sock"
+            - "--kubelet-registration-path=/var/lib/kubelet/plugins/pd.csi.storage.gke.io/csi.sock"
+          lifecycle:
```
msau42, line 14: You can actually remove the entire prestop hook, it doesn't work anymore and the solution is not going to use this: https://github.com/kubernetes-csi/node-driver-registrar/issues/21

deploy/kubernetes/overlays/linux/base/volumes.yaml
```diff
@@ -0,0 +1,44 @@
+kind: DaemonSet
```
msau42, line 1: Can this be combined with gcepd.yaml?

deploy/kubernetes/overlays/linux/dev/WARNING.md
```diff
@@ -0,0 +1,15 @@
...
+DISCLAIMER: THE LATEST IMAGE IS CONSTANTLY CHANGING WITH DEVELOPMENT AND CAN BE
+BROKEN AT ANY TIME
+
+This is the absolute cutting edge development Driver, it is intended for testing
+and development only and can have vast differences in
+functionality/behavior/configuration. Use only to try the newest features that
+are not guaranteed to work yet.
+
+APPROXIMATE CHANGELOG in latest:
+* Topology
```
msau42, line 11: I think all of these features are now in the latest release and can be removed

deploy/kubernetes/overlays/windows/alpha/kustomization.yaml
```diff
@@ -0,0 +1,34 @@
...
+- noderegistrar.yaml
+- volumes.yaml
+images:
+- name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
+  # Don't change stable image without changing pdImagePlaceholder in
+  # test/k8s-integration/main.go
+  newName: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
+  newTag: "v0.7.0-gke.0"
+- name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver-win
+  # Temporarly set to the private repo. Will swtich to public one
```
msau42, line 16: Maybe let's move these to a "dev" overlay for now until we have public images?

deploy/kubernetes/overlays/linux/alpha/WARNING.md
```diff
@@ -0,0 +1,3 @@
+WARNING: DO NOT USE THE ALPHA VERSION OF THE DRIVER FOR PRODUCTION
```
msau42, line 1: Can you delete the original files so that git can recognize these changes as a rename and we can retain git history?

deploy/kubernetes/overlays/linux/base/volumes.yaml
jingxu97: we could, but from kustomization doc, they suggested to make small changes for each file, since volumes and gcepd container are two different parts, I use two files.

deploy/kubernetes/base/node.yaml
jingxu97: fixed

deploy/kubernetes/overlays/linux/base/volumes.yaml
msau42: I think it makes sense to keep independent patches in separate files, but in this case, since container.volumeMounts and pod.spec.volumes are tied together, I think it makes sense to keep them together.

deploy/kubernetes/base/node.yaml
jingxu97: fixed. we might use a different way to workaround it. Currently we don't need to set it for windows.

deploy/kubernetes/overlays/linux/base/volumes.yaml
jingxu97: got it. I updated it.

jingxu97:
comments are addressed. PTAL

deploy/kubernetes/base/node.yaml
```diff
@@ -41,34 +34,13 @@ spec:
...
-            - name: udev-rules-lib
-              mountPath: /lib/udev
-            - name: udev-socket
-              mountPath: /run/udev
-            - name: sys
-              mountPath: /sys
-      nodeSelector:
-        kubernetes.io/os: linux
       volumes:
         - name: registration-dir
```
msau42, line 45: Should these volumes be removed too since they are linux specific?

deploy/kubernetes/overlays/windows/dev/drivertest.yaml
```diff
@@ -0,0 +1,72 @@
+kind: DaemonSet
```
msau42, line 1: is this file used?

deploy/kubernetes/overlays/windows/dev/securityContext.yaml
```diff
@@ -0,0 +1,12 @@
+kind: DaemonSet
```
msau42, line 1: is this file used?

deploy/kubernetes/overlays/windows/dev/tmp.yaml
```diff
@@ -0,0 +1,541 @@
+apiVersion: v1
```
msau42, line 1: is this file used?

deploy/kubernetes/overlays/windows/dev/drivertest.yaml
jingxu97: no, I think I added the whole dir and forgot to remove those tmp files. Removed them.

deploy/kubernetes/overlays/windows/dev/securityContext.yaml
jingxu97: removed

deploy/kubernetes/base/node.yaml
jingxu97: no, windows used it too, I think.

msau42: Windows seems to be overriding them? https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/510/files#diff-c6cce7581a0b791e1a6268807fdf0a3fR38

deploy/kubernetes/overlays/windows/dev/tmp.yaml
jingxu97: fixed

deploy/kubernetes/base/node.yaml
jingxu97: it will not override. just merging the two parts. But I think I need to remove lifecycle part similar to linux.

deploy/kubernetes/overlays/linux/base/noderegistrar.yaml
verult: Does the same need to be done for the windows registrar?

verult:
/lgtm

Consider adding a release note for the new `OS` env var in deployment script.

And a nit: maybe change "base" overlay name to something like "common". It could be confusing which "base" is referred to in other overlays.

deploy/kubernetes/base/node.yaml
msau42: What is the purpose of keeping the volumes here, if this base spec doesn't have the corresponding volumeMounts? It may be easier to read if we keep both volumeMounts and volumes together in one file.

deploy/kubernetes/delete-driver.sh
```diff
@@ -13,11 +13,12 @@ set -o errexit
 readonly NAMESPACE="${GCE_PD_DRIVER_NAMESPACE:-gce-pd-csi-driver}"
 readonly DEPLOY_VERSION="${GCE_PD_DRIVER_VERSION:-stable}"
 readonly PKGDIR="${GOPATH}/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver"
+readonly OS="${OS:-linux}"
 source "${PKGDIR}/deploy/common.sh"
 
 ensure_kustomize
 
-${KUSTOMIZE_PATH} build ${PKGDIR}/deploy/kubernetes/overlays/${DEPLOY_VERSION} | ${KUBECTL} delete -v="${VERBOSITY}" --ignore-not-found -f -
+${KUSTOMIZE_PATH} build ${PKGDIR}/deploy/kubernetes/overlays/${OS}/${DEPLOY_VERSION} | ${KUBECTL} delete -v="${VERBOSITY}" --ignore-not-found -f -
```
msau42, line 21: Since the DaemonSet has an OS node selector, what do you think of deploying both Linux + Windows daemonsets in one overlay?  Then this easily supports mixed OS nodes.

jingxu97: What do you mean by deploying in one overlay?

deploy/kubernetes/base/node.yaml
jingxu97: Now I look at the node.yaml again, the most part is about volumes and volumemounts, which are quite different from linux and windows. I think it is more clean just deploy node.yaml in overlay layer, and remove it from base layer.

deploy/kubernetes/base/setup-cluster.yaml
```diff
@@ -199,6 +190,19 @@ subjects:
...
+metadata:
+  name: csi-gce-pd-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: csi-gce-pd-node-deploy
+subjects:
+- kind: ServiceAccount
+  name: csi-gce-pd-controller-sa
+
```
jingxu97, line 205: I added this binding, because without it, when pod security policy admission controller is enabled, service account csi-gce-pd-controller-sa cannot create pod at all.

jingxu97:
I updated the PR. PTAL. Thanks!

deploy/kubernetes/delete-driver.sh
msau42: I mean actually have 2 DaemonSets in the same overlay (and not have os-specific overlays), one for linux and one for windows. Then when you run "deploy-driver.sh", you actually deploy both linux + windows daemonsets.

msau42: It means we need 2 DaemonSets in the base, and if we end up changing something, we'll have to remember to update both DaemonSets.

But this way, we can support clusters that have mixed windows + linux nodes.

msau42:
cc @mattcary

deploy/kubernetes/delete-driver.sh
jingxu97: I think I prefer os-specific overlay for clear structure. Not only daemonset is different, also have different other things such as pod security policy.
For cluster with mixed windows and linux, with current script, user can set os twice and deploy it twice. Or I can update script to accept multiple OSes.

msau42: I'm not sure if running the script twice will work. The script does things like delete/create the controller secret which should only be done once.

If you are planning to remove the node.yaml from the base anyway, and not have it share anything, then it may be simple to just have a node_linux.yaml and node_windows.yaml

jingxu97: I updated with changes similar to what you mentioned. The deployment env can take both os and version information, and also allow you to set multiple values. I tested the script and it can install drivers for a mixed cluster.

deploy/kubernetes/overlays/windows/base/allowedHostPaths.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: csi-gce-pd-node-psp
+spec:
+  allowedHostPaths:
+  - pathPrefix: \var\lib\kubelet
```
msau42, line 7: I think this is going to replace the current PSP with these rules, instead of add, which is not desirable if we're going to be running both linux + windows.

deploy/kubernetes/overlays/windows/base/kustomization.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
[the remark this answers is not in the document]
jingxu97: So with one overlay, we have both windows and linux together. What if user need to deploy with mixed OS and versions, e.g, linux stable, and windows dev etc? Or just because some versions are only available in linux, not windows?

deploy/kubernetes/overlays/windows/base/allowedHostPaths.yaml
jingxu97: good point, will try to make a merge here.

deploy/kubernetes/overlays/windows/base/kustomization.yaml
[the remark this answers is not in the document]
msau42: I think we'll still need to have different overlays for dev/alpha/staging/stable and we can control the versions used in each. For example, the dev overlay could use stable versions for linux, and unstable versions for windows. And in terms of development flow, we can add Windows first to the dev overlay, then once we're ready, add it to staging, and then when we're ready to release, add it to stable.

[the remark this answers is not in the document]
jingxu97: Maybe I miss something here, so you think user should not need a mixed version for windows and linux. If they want to deploy stable version, it will be stable version for both linux and windows? But what if windows don't have stable version yet, user must to deploy dev version for both linux and windows, instead of one stable version and one dev version?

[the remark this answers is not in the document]
jingxu97: Also in the case user already deploy. a Linux driver, and want to add a windows driver, with one overlay, it will try to deploy both linux and windows. Maybe some setup might be messed up with different versions of deployment?

deploy/kubernetes/overlays/linux/base/enableHostNetwork.yaml
```diff
@@ -0,0 +1,8 @@
+kind: DaemonSet
```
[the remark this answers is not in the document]
jingxu97: right, fix it

deploy/kubernetes/base/controller.yaml
```diff
@@ -82,5 +84,10 @@ spec:
         - name: cloud-sa-volume
           secret:
             secretName: cloud-sa
+      # https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+      # See "special case". This will tolerate everything. Node component should
+      # be scheduled on all nodes.
+      tolerations:
+      - operator: Exists
```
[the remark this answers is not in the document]
jingxu97: good point, will remove it. I added it because I see node has it. And my test setup has linux tainted.

deploy/kubernetes/overlays/linux/base/noderegistrar.yaml
```diff
@@ -0,0 +1,13 @@
+kind: DaemonSet
```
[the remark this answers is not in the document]
jingxu97: fix it.

deploy/kubernetes/overlays/windows/base/noderegistrar.yaml
```diff
@@ -0,0 +1,17 @@
+kind: DaemonSet
```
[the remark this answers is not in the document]
jingxu97: fix it.

deploy/kubernetes/overlays/windows/base/allowedHostPaths.yaml
jingxu97: actually, I think it is better to have separate policy for linux and windows instead of merging

deploy/kubernetes/overlays/windows/base/kustomization.yaml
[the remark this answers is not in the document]
jingxu97: I have been thinking the structure again last night :), and realized this is more like a multi-stage process. So instead of base/overlay structure, I propose to use the following as multiple stages and use them as bases. I am working on this structure now and will update. Please let me know what you think about this structure.

- cluster_controller_setup : includes cluster setup and controller yaml file
- node_setup: 
      *  Linux: node and other related yaml files for Linux
      *  Windows:    node and other related yaml files for Windows

- image_update
      *  alpha:
      *. stable:
      *. staging:
      * dev:
...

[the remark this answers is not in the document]
msau42: I'm not quite sure how this would work out but am curious to see. One thing to note is that alpha/staging/stable is not necessarily just different versions. Different features may be added depending on their maturity.

>  what if windows don't have stable version yet, user must deploy dev version for both linux and windows, instead of one stable version and one dev version?

In terms of looking at the driver itself, I think we should consider both Linux + Windows parts of the driver to be part of one complete driver instead of 2 independent drivers. So considering that Windows is still alpha right now, I think it is fine if it uses the Linux alpha versions as well (which currently don't have any difference from stable). I don't think a user will want to have a mixed Linux + Windows cluster where the Linux part is stable and the Windows part is alpha. I would expect a user to have 2 separate clusters: a Linux-only cluster used for production, and an experimental Windows-only cluster.

> Also in the case user already deploy a Linux driver, and want to add a windows driver (might be a different version from Linux), with one overlay, it will try to deploy both linux and windows. Some setup might be messed up with different versions of deployment?

I agree with one deployment and one place to define all the versions, it's easier to make sure the versions are in-sync with each other.

jingxu97:
opened https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/520 to replace this one.

PR #515, opened 2020-05-27
Add option to deploy GKE managed PD CSI driver for integration tests

### LINKED ISSUE #511: Driver integration test should support option to deploy GKE managed PD CSI driver
In https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go, an option should be provided to deploy a GKE Managed driver, so that kubernetes e2e test suites can be run for this managed driver.

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Add hook to deploy GKE with GCE PD CSI driver, and run kubernetes e2e tests against the managed driver.
    Summary of changes:
    1. Add a new hook to pass release-channel for GKE cluster deployment.
    2. Add a new hook to deploy GKE managed PD CSI driver.
    3. Added a version comparision util with supporting UTs.

Fixes #511 

```release-note
Add hook to deploy GKE with GCE PD CSI driver, and run kubernetes e2e tests against the managed driver
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -9,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	apimachineryversion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/klog"
 )
 
@@ -131,7 +133,7 @@ func setImageTypeEnvs(imageType string) error {
 	return nil
 }
 
-func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string) error {
+func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, useManagedDriver bool) error {
 	locationArg, locationVal, err := gkeLocationArgs(gceZone, gceRegion)
 	if err != nil {
 		return err
@@ -150,9 +152,23 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string) err
 			return err
 		}
 	}
-	cmd := exec.Command("gcloud", "container", "clusters", "create", gkeTestClusterName,
-		locationArg, locationVal, "--cluster-version", *gkeClusterVer, "--num-nodes", strconv.Itoa(numNodes),
-		"--quiet", "--machine-type", "n1-standard-2", "--image-type", imageType)
+
+	var cmd *exec.Cmd
+	cmdParams := []string{"container", "clusters", "create", gkeTestClusterName,
+		locationArg, locationVal, "--num-nodes", strconv.Itoa(numNodes)}
+	if isVariableSet(gkeClusterVer) {
+		cmdParams = append(cmdParams, "--cluster-version", *gkeClusterVer)
+	} else {
+		cmdParams = append(cmdParams, "--release-channel", *gkeReleaseChannel)
+	}
+
+	if useManagedDriver {
+		// PD CSI Driver add on is enabled only in gcloud beta.
+		cmdParams = append([]string{"beta"}, cmdParams...)
+		cmdParams = append(cmdParams, "--addons", "GcePersistentDiskCsiDriver")
+	}
+
+	cmd = exec.Command("gcloud", cmdParams...)
 	err = runCommand("Staring E2E Cluster on GKE", cmd)
 	if err != nil {
 		return fmt.Errorf("failed to bring up kubernetes e2e cluster on gke: %v", err)
@@ -289,4 +305,31 @@ func getNormalizedVersion(kubeVersion, gkeVersion string) (string, error) {
 	}
 	return strings.Join(toks[:2], "."), nil
 
+}
+
+func getKubeClusterVersion() (string, error) {
+	out, err := exec.Command("kubectl", "version", "-o=json").CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("failed to obtain cluster version, error: %v", err)
+	}
+	type version struct {
+		ClientVersion *apimachineryversion.Info `json:"clientVersion,omitempty" yaml:"clientVersion,omitempty"`
+		ServerVersion *apimachineryversion.Info `json:"serverVersion,omitempty" yaml:"serverVersion,omitempty"`
+	}
+
+	var v version
+	err = json.Unmarshal(out, &v)
+	if err != nil {
+		return "", fmt.Errorf("Failed to parse kubectl version output, error: %v", err)
+	}
+
+	return v.ServerVersion.GitVersion, nil
+}
+
+func mustGetKubeClusterVersion() string {
+	ver, err := getKubeClusterVersion()
+	if err != nil {
+		klog.Fatalf("Error: %v", err)
+	}
+	return ver
 }
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -22,27 +22,28 @@ import (
 	"path/filepath"
 	"syscall"
 
-	testutils "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/test/e2e/utils"
-
 	"k8s.io/apimachinery/pkg/util/uuid"
+	apimachineryversion "k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/klog"
+	testutils "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/test/e2e/utils"
 )
 
 var (
 	// Kubernetes cluster flags
-	teardownCluster  = flag.Bool("teardown-cluster", true, "teardown the cluster after the e2e test")
-	teardownDriver   = flag.Bool("teardown-driver", true, "teardown the driver after the e2e test")
-	bringupCluster   = flag.Bool("bringup-cluster", true, "build kubernetes and bringup a cluster")
-	gceZone          = flag.String("gce-zone", "", "zone that the gce k8s cluster is created/found in")
-	gceRegion        = flag.String("gce-region", "", "region that gke regional cluster should be created in")
-	kubeVersion      = flag.String("kube-version", "", "version of Kubernetes to download and use for the cluster")
-	testVersion      = flag.String("test-version", "", "version of Kubernetes to download and use for tests")
-	kubeFeatureGates = flag.String("kube-feature-gates", "", "feature gates to set on new kubernetes cluster")
-	localK8sDir      = flag.String("local-k8s-dir", "", "local prebuilt kubernetes/kubernetes directory to use for cluster and test binaries")
-	deploymentStrat  = flag.String("deployment-strategy", "gce", "choose between deploying on gce or gke")
-	gkeClusterVer    = flag.String("gke-cluster-version", "", "version of Kubernetes master and node for gke")
-	numNodes         = flag.Int("num-nodes", -1, "the number of nodes in the test cluster")
-	imageType        = flag.String("image-type", "cos", "the image type to use for the cluster")
+	teardownCluster   = flag.Bool("teardown-cluster", true, "teardown the cluster after the e2e test")
+	teardownDriver    = flag.Bool("teardown-driver", true, "teardown the driver after the e2e test")
+	bringupCluster    = flag.Bool("bringup-cluster", true, "build kubernetes and bringup a cluster")
+	gceZone           = flag.String("gce-zone", "", "zone that the gce k8s cluster is created/found in")
+	gceRegion         = flag.String("gce-region", "", "region that gke regional cluster should be created in")
+	kubeVersion       = flag.String("kube-version", "", "version of Kubernetes to download and use for the cluster")
+	testVersion       = flag.String("test-version", "", "version of Kubernetes to download and use for tests")
+	kubeFeatureGates  = flag.String("kube-feature-gates", "", "feature gates to set on new kubernetes cluster")
+	localK8sDir       = flag.String("local-k8s-dir", "", "local prebuilt kubernetes/kubernetes directory to use for cluster and test binaries")
+	deploymentStrat   = flag.String("deployment-strategy", "gce", "choose between deploying on gce or gke")
+	gkeClusterVer     = flag.String("gke-cluster-version", "", "version of Kubernetes master and node for gke")
+	numNodes          = flag.Int("num-nodes", -1, "the number of nodes in the test cluster")
+	imageType         = flag.String("image-type", "cos", "the image type to use for the cluster")
+	gkeReleaseChannel = flag.String("gke-release-channel", "", "GKE release channel to be used for cluster deploy. One of 'rapid', 'stable' or 'regular'")
 
 	// Test infrastructure flags
 	boskosResourceType = flag.String("boskos-resource-type", "gce-project", "name of the boskos resource type to reserve")
@@ -51,10 +52,11 @@ var (
 	inProw             = flag.Bool("run-in-prow", false, "is the test running in PROW")
 
 	// Driver flags
-	stagingImage      = flag.String("staging-image", "", "name of image to stage to")
-	saFile            = flag.String("service-account-file", "", "path of service account file")
-	deployOverlayName = flag.String("deploy-overlay-name", "", "which kustomize overlay to deploy the driver with")
-	doDriverBuild     = flag.Bool("do-driver-build", true, "building the driver from source")
+	stagingImage        = flag.String("staging-image", "", "name of image to stage to")
+	saFile              = flag.String("service-account-file", "", "path of service account file")
+	deployOverlayName   = flag.String("deploy-overlay-name", "", "which kustomize overlay to deploy the driver with")
+	doDriverBuild       = flag.Bool("do-driver-build", true, "building the driver from source")
+	useGKEManagedDriver = flag.Bool("use-gke-managed-driver", false, "use GKE managed PD CSI driver for the tests")
 
 	// Test flags
 	migrationTest = flag.Bool("migration-test", false, "sets the flag on the e2e binary signalling migration")
@@ -75,12 +77,23 @@ func init() {
 func main() {
 	flag.Parse()
 
-	if !*inProw {
+	if !*inProw && !*useGKEManagedDriver {
 		ensureVariable(stagingImage, true, "staging-image is a required flag, please specify the name of image to stage to")
 	}
 
+	if *useGKEManagedDriver {
+		ensureVariableVal(deploymentStrat, "gke", "deployment strategy must be GKE for using managed driver")
+		ensureFlag(doDriverBuild, false, "'do-driver-build' must be false when using GKE managed driver")
+		ensureFlag(teardownDriver, false, "'teardown-driver' must be false when using GKE managed driver")
+		ensureVariable(stagingImage, false, "'staging-image' must not be set when using GKE managed driver")
+		ensureVariable(deployOverlayName, false, "'deploy-overlay-name' must not be set when using GKE managed driver")
+	}
+
 	ensureVariable(saFile, true, "service-account-file is a required flag")
-	ensureVariable(deployOverlayName, true, "deploy-overlay-name is a required flag")
+	if !*useGKEManagedDriver {
+		ensureVariable(deployOverlayName, true, "deploy-overlay-name is a required flag")
+	}
+
 	ensureVariable(testFocus, true, "test-focus is a required flag")
 	ensureVariable(imageType, true, "image type is a required flag. Available options include 'cos' and 'ubuntu'")
 
@@ -103,7 +116,8 @@ func main() {
 	if *deploymentStrat == "gke" {
 		ensureFlag(migrationTest, false, "Cannot set deployment strategy to 'gke' for migration tests.")
 		ensureVariable(kubeVersion, false, "Cannot set kube-version when using deployment strategy 'gke'. Use gke-cluster-version.")
-		ensureVariable(gkeClusterVer, true, "Must set gke-cluster-version when using deployment strategy 'gke'.")
+		ensureExactlyOneVariableSet([]*string{gkeClusterVer, gkeReleaseChannel},
+			"For GKE cluster deployment, exactly one of 'gke-cluster-version' or 'gke-release-channel' must be set")
 		ensureVariable(kubeFeatureGates, false, "Cannot set feature gates when using deployment strategy 'gke'.")
 		if len(*localK8sDir) == 0 {
 			ensureVariable(testVersion, true, "Must set either test-version or local k8s dir when using deployment strategy 'gke'.")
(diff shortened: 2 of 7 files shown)


test/k8s-integration/cluster.go
```diff
@@ -150,9 +150,19 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string) err
...
 		}
 	}
-	cmd := exec.Command("gcloud", "container", "clusters", "create", gkeTestClusterName,
-		locationArg, locationVal, "--cluster-version", *gkeClusterVer, "--num-nodes", strconv.Itoa(numNodes),
-		"--quiet", "--machine-type", "n1-standard-2", "--image-type", imageType)
+
+	var cmd *exec.Cmd
+	if useManagedDriver {
+		// PD CSI Driver add on is enabled only in gcloud beta.
+		cmd = exec.Command("gcloud", "beta", "container", "clusters", "create", gkeTestClusterName,
```
mattcary, line 157: You should put the args in an array and then pass to Command as Command(args...)

That way all the common arguments only have to be stated once which will help maintenance long-term.

test/k8s-integration/main.go
```diff
@@ -75,12 +76,21 @@ func init() {
...
 	flag.Parse()
 
-	if !*inProw {
+	if !*inProw && !*useGKEManagedDriver {
 		ensureVariable(stagingImage, true, "staging-image is a required flag, please specify the name of image to stage to")
 	}
 
+	if *useGKEManagedDriver {
+		ensureVariableVal(deploymentStrat, "gke", "deployment strategy must be GKE for using managed driver")
+		ensureFlag(doDriverBuild, false, "driver build flag will be ignored when using GKE managed driver")
```
mattcary, line 85: more accurate to say that the flag must be false, as that's what you're checking.

test/k8s-integration/cluster.go
```diff
@@ -150,9 +150,19 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string) err
...
+			locationArg, locationVal, "--cluster-version", *gkeClusterVer, "--num-nodes", strconv.Itoa(numNodes),
```
mattcary, line 158: I wonder if we're going to need an option for release channels instead of just specifying cluster-version. For example, now the rapid channel is the only way to get 1.17; the highest --cluster-version one can pass is 1.16.8.

test/k8s-integration/cluster.go
saikat-royc: Ack

test/k8s-integration/main.go
saikat-royc: Ack

test/k8s-integration/cluster.go
saikat-royc: Ack, have added support for release channels.

test/k8s-integration/main.go
```diff
@@ -344,6 +367,21 @@ func generateTestSkip(normalizedVersion string) string {
 	return skipString
 }
 
+// Helper function to generate skip string when using GKE managed PD CSI driver.
+func generateGKEBasedTestSkip(clusterVersion string) string {
+	skipString := "\\[Disruptive\\]|\\[Serial\\]"
+	curVer := MustParseVersion(clusterVersion)
+	if (*curVer).IsLessThan(MustParseVersion("1.17.0")) {
+		skipString = skipString + "|volumeMode\\sshould\\snot\\smount\\s/\\smap\\sunused\\svolumes\\sin\\sa\\spod"
+	}
```
saikat-royc, line 376: @msau42  is this the right version to skip the mount tests?

msau42: The fix was in 1.16.0, it should be < 1.16.0. 

Alternatively, can we share the same skip function?

mattcary: Maybe run the cluster version through getNormalizedVersion, apply generateTestSkip, and then append additional skips based on the full version?

saikat-royc: @msau42  Its tricky to share the same skipstring, since that works on major.minor version only. However, for using GKE managed driver, say snapshots, we need to check the full version 1.18.3.
Perhaps, using both generateTestSkip, and additional filters based on GKE full version is the way to go as @mattcary  suggested.

test/k8s-integration/version.go
```diff
@@ -0,0 +1,113 @@
...
+
+	versionRegex         = regexp.MustCompile(`^` + versionNum + `\.` + versionNum + `\.` + versionNum + internalPatchVersion + "?$")
+	gkeExtraVersionRegex = regexp.MustCompile(`^(?:gke)\.(0|[1-9][0-9]*)$`)
+)
+
+type Version struct {
+	version [4]int
+	extra   string
+}
+
```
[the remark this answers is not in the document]
saikat-royc: yes, we can keep the normalized version in the Version struct, so its all available in one place. If thats what you mean @mattcary

test/k8s-integration/main.go
msau42: Can we normalize the "x.y" version to "x.y.z" so it'll work with the new "lessThan" checks?

saikat-royc: does x.y normalize to x.y.0?

test/k8s-integration/version.go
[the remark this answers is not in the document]
mattcary: Yes. Then getNormalizedVersion can go away, that was my point.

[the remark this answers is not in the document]
saikat-royc: ok that sounds like a cleaner approach. Let me know if we are on the same page, 
where we get kubectl version in all cases and then apply the new IsLessThan() based comparison for generating skip strings.
The function
generateskipstring(clusterversion, use_managed_gke_driver) {
    // skip stirng for flaky tests
    if version.lessthan("1.16.0") {
        // add mount tests to skip string
    }

    if use_managed_gke_driver {
       if version.lessthan("1.18.3") {
            // add snapshot to skip string
        }
    } else {
        version.lessthan("1.17.0") {
            // add snapshot skip string
         }
     }
}

Thoughts?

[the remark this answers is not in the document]
mattcary: Yeah, then we can have one if statement per skip, rather than the currently sort-of-confusing logic using the fall-through in the switch.

test/k8s-integration/version.go
```diff
@@ -0,0 +1,113 @@
+package main
```
verult, line 1: Could https://github.com/kubernetes/apimachinery/blob/master/pkg/util/version/version.go be reused?

test/k8s-integration/main.go
```diff
@@ -75,12 +77,21 @@ func init() {
...
 
-	if !*inProw {
+	if !*inProw && !*useGKEManagedDriver {
 		ensureVariable(stagingImage, true, "staging-image is a required flag, please specify the name of image to stage to")
 	}
 
+	if *useGKEManagedDriver {
+		ensureVariableVal(deploymentStrat, "gke", "deployment strategy must be GKE for using managed driver")
+		ensureFlag(doDriverBuild, false, "'do-driver-build' must be false when using GKE managed driver")
+		ensureFlag(teardownDriver, false, "'teardown-driver' must be false when using GKE managed driver")
```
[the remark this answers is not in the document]
saikat-royc: good point to avoid confusion. In order to add this check we need a base_cmd without the 'deploy-overlay-name' flag in run-k8s-integration.sh

test/k8s-integration/cluster.go
```diff
@@ -150,9 +152,26 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string) err
...
 	}
-	cmd := exec.Command("gcloud", "container", "clusters", "create", gkeTestClusterName,
-		locationArg, locationVal, "--cluster-version", *gkeClusterVer, "--num-nodes", strconv.Itoa(numNodes),
-		"--quiet", "--machine-type", "n1-standard-2", "--image-type", imageType)
+
+	var cmd *exec.Cmd
+	cmdParams := []string{"container", "clusters", "create", gkeTestClusterName,
+		locationArg, locationVal, "--num-nodes", strconv.Itoa(numNodes)}
+
+	ensureExactlyOneVariableSet([]*string{gkeClusterVer, gkeReleaseChannel},
```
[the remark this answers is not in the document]
saikat-royc: Ack

test/k8s-integration/cluster.go
```diff
@@ -290,3 +309,30 @@ func getNormalizedVersion(kubeVersion, gkeVersion string) (string, error) {
...
+	var v version
+	err = json.Unmarshal(out, &v)
+	if err != nil {
+		return "", fmt.Errorf("Failed to parse kubectl version output, error: %v", err)
+	}
+
+	return v.ServerVersion.GitVersion, nil
+}
+
+func MustGetKubClusterVersion() string {
```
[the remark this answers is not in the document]
saikat-royc: Done

test/k8s-integration/main.go
```diff
@@ -295,12 +312,11 @@ func handle() error {
 		}
 	}
 
-	normalizedVersion, err := getNormalizedVersion(*kubeVersion, *gkeClusterVer)
-	if err != nil {
-		return fmt.Errorf("failed to get cluster minor version: %v", err)
-	}
+	clusterVersion := mustGetKubClusterVersion()
+	fmt.Printf("Kubernetes cluster version %q\n", clusterVersion)
+	testSkip := generateTestSkip(clusterVersion, *useGKEManagedDriver)
```
saikat-royc, line 317: There is a problem in this current logic of unified comparision logic for gce and gke.
For example kubetest --extract=release/latest --up deploys a server version of "v1.19.0-beta.0". Current comparision logic wont work. Needs additional investigation if we can use https://github.com/kubernetes/apimachinery/blob/master/pkg/util/version/version.go for parse and comparision of gke builds. (e.g. v1.19.0-gke.0)

test/k8s-integration/version.go
saikat-royc: Thanks for pointing this out

test/k8s-integration/version.go
```diff
@@ -0,0 +1,113 @@
...
+
+var (
+	versionNum           = `(0|[1-9][0-9]*)`
+	internalPatchVersion = `(\-[a-zA-Z0-9_.+-]+)`
+
+	versionRegex         = regexp.MustCompile(`^` + versionNum + `\.` + versionNum + `\.` + versionNum + internalPatchVersion + "?$")
+	gkeExtraVersionRegex = regexp.MustCompile(`^(?:gke)\.(0|[1-9][0-9]*)$`)
+)
+
+type Version struct {
```
verult, line 20: nit: unexported

test/k8s-integration/version.go
```diff
@@ -0,0 +1,113 @@
...
+	version [4]int
+	extra   string
+}
+
+func (v Version) String() string {
```
verult, line 25: nit: pointer as receiver

test/k8s-integration/main.go
```diff
@@ -317,30 +344,34 @@ func handle() error {
...
+}
+
+func generateGKETestSkip(clusterVersion string, use_gke_managed_driver bool) string {
+	skipString := "\\[Disruptive\\]|\\[Serial\\]"
+	curVer := mustParseVersion(clusterVersion)
+	if (*curVer).lessThan(mustParseVersion("1.16.0")) {
+		skipString = skipString + "|volumeMode\\sshould\\snot\\smount\\s/\\smap\\sunused\\svolumes\\sin\\sa\\spod"
+	}
+
+	if (use_gke_managed_driver && (*curVer).lessThan(mustParseVersion("1.18.3-gke.0"))) ||
```
verult, line 370: Will this need to be updated soon to include a 1.17 bound? Maybe include a TODO

test/k8s-integration/version.go
verult, line 1: Add comments on why this file is needed

test/k8s-integration/version.go
```diff
@@ -0,0 +1,113 @@
...
+	versionNum           = `(0|[1-9][0-9]*)`
+	internalPatchVersion = `(\-[a-zA-Z0-9_.+-]+)`
+
+	versionRegex         = regexp.MustCompile(`^` + versionNum + `\.` + versionNum + `\.` + versionNum + internalPatchVersion + "?$")
+	gkeExtraVersionRegex = regexp.MustCompile(`^(?:gke)\.(0|[1-9][0-9]*)$`)
+)
+
+type Version struct {
+	version [4]int
+	extra   string
```
verult, line 22: Does this need to be a struct field?

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,157 @@
...
+			},
+		},
+		{
+			version: "v1.18.0-gke.0",
+			expectedV: Version{
+				version: [4]int{1, 18, 0, 0},
+			},
+		},
+		{
+			version: "1.18.0",
```
verult, line 28: the next two are duplicates

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,157 @@
...
+			expectErr: true,
+		},
+		{
+			version:   "1.-18.9",
+			expectErr: true,
+		},
+		{
+			version:   "1.18",
+			expectErr: true,
+		},
```
verult, line 75: maybe a test with `-alpha.x`

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+}
+
+func TestIsVersionLessThan(t *testing.T) {
+	tests := []struct {
+		leftVersion  string
+		rightVersion string
+		expectRes    bool
+	}{
+		// Positive cases.
```
verult, line 109: Should cover left > right as well.

Also include a case like "1.18.0" and "1.18.1" where both versions lack the gke part and they are unequal

test/k8s-integration/main.go
mattcary: I don't think so, PD CSI Snapshots are never going to be available from the managed driver before this 1.18 version. I don't think we care about testing non-managed versions on GKE (?)

msau42: I think we should consider getting this into 1.17 since the general snapshots feature is beta in 1.17.

mattcary: In order to get a managed driver snapsnot test in 1.17 we'll have to add a case to deploy only the sidecar manually?

I guess 1.17 will be around long enough it's worth having a test for that combination.

saikat-royc: GKE 1.17.5-gke.9 will have the pd csi snapshot support. I will put in the change here.

test/k8s-integration/version.go
saikat-royc: Done

test/k8s-integration/version_test.go
saikat-royc: Done

test/k8s-integration/version_test.go
saikat-royc: Done

test/k8s-integration/version_test.go
saikat-royc: Done

test/k8s-integration/version.go
saikat-royc: Done

test/k8s-integration/version.go
saikat-royc: Done

test/k8s-integration/version.go
saikat-royc: can you elaborate please? array helps in easier range index based comparison.

test/k8s-integration/main.go
```diff
@@ -317,30 +344,37 @@ func handle() error {
...
+
+func generateGKETestSkip(clusterVersion string, use_gke_managed_driver bool) string {
+	skipString := "\\[Disruptive\\]|\\[Serial\\]"
+	curVer := mustParseVersion(clusterVersion)
+	if curVer.lessThan(mustParseVersion("1.16.0")) {
+		skipString = skipString + "|volumeMode\\sshould\\snot\\smount\\s/\\smap\\sunused\\svolumes\\sin\\sa\\spod"
+	}
+
+	// For GKE deployed PD CSI snapshot is enabled in 1.17.5-gke.9(and higher), 1.17.6-gke.4(and higher), 1.18.3(and higher).
+	if (use_gke_managed_driver && curVer.lessThan(mustParseVersion("1.17.5-gke.9"))) ||
```
mattcary, line 371: Would this read better writen like

if (use_gke_managed_driver) {
  if curVer.lessThan(1.17.5-gke.9) {
    skipString = skipString + "|VolumeSnapshotDataSource"
  } else if curVer.LessThen(1.17.6.0) {
   // don't skip
  } else if curVer.LessThan(17.6-gke.4) {
   skipString = ...
  } else if curVer.LessThan(1.18.0) {
   // don't skip
  } else if curVer.LessThan(1.18.3) {
   skipStyring = ...
 }

It's still complicated but at least one can just read down and quickly see that status of each version rather than trying to figure out more complicated nested logic.

msau42: I think we also don't need to be so exact with the versions considering that we control the test job configurations too.

saikat-royc: @msau42  keeping the logic accurate would make sure the test skip strings logic robust and gives clarity which version exactly supports snapshots. Thoughts?

msau42: I think we can handle the supported versions in the top level README (I doubt anyone will actually come down to the test code to figure it out), and maybe clarify as a comment in the test code if we want to.

On the GKE side, I think we can handle the complexity of only adding the test job for supported versions.

saikat-royc: From the prow configs I see we add cluster versions like -extract=gke-latest-1.17 (instead of adding specific gke releases)
so for e.g if a version < 1.17.6-gke.4 is picked test breaks. Feel it's easier to control and be deterministic in the go code. Let's discuss on this offline and reach a conclusion.

test/k8s-integration/main.go
```diff
@@ -295,12 +312,22 @@ func handle() error {
...
+	case "gce":
+		testSkip = generateGCETestSkip(clusterVersion)
+	case "gke":
+		testSkip = generateGKETestSkip(clusterVersion, *useGKEManagedDriver)
+	default:
+		return fmt.Errorf("Unknown deployment strategy %s", *deploymentStrat)
 	}
 
-	testSkip := generateTestSkip(normalizedVersion)
+	fmt.Printf("Test skip string %s\n", testSkip)
```
mattcary, line 330: Debugging code that should be removed?

test/k8s-integration/main.go
```diff
@@ -295,12 +312,22 @@ func handle() error {
...
 	}
 
-	normalizedVersion, err := getNormalizedVersion(*kubeVersion, *gkeClusterVer)
-	if err != nil {
-		return fmt.Errorf("failed to get cluster minor version: %v", err)
+	// Kubernetes version of GKE deployments are expected to be of the pattern x.y.z-gke.k,
+	// hence we use the main.Version utils to parse and compare GKE managed cluster versions.
+	// For clusters deployed on GCE, use the apimachinery version utils (which supports non-gke based semantic versioning).
+	clusterVersion := mustGetKubeClusterVersion()
+	fmt.Printf("Kubernetes cluster version %q\n", clusterVersion)
```
mattcary, line 319: Debugging code that should be removed?

test/k8s-integration/main.go
```diff
@@ -317,30 +344,37 @@ func handle() error {
...
-		fallthrough
-	case "1.15":
-		fallthrough
-	case "1.16":
-		// "volumeMode should not mount / map unused volumes in a pod" tests a
-		// bug-fix introduced in 1.17
-		// (https://github.com/kubernetes/kubernetes/pull/81163)
+	v := apimachineryversion.MustParseSemantic(clusterVersion)
+
+	// "volumeMode should not mount / map unused volumes in a pod" tests a
```
mattcary, line 351: You dropped the "bug-fix introduced in 1.17" line

test/k8s-integration/main.go
saikat-royc: Ack

test/k8s-integration/main.go
saikat-royc: Ack

test/k8s-integration/main.go
saikat-royc: Ack

test/k8s-integration/main.go
saikat-royc: based on @msau42 's recommendation, given the prow jobs ensure to test against latest of any cluster version, keep the logic simple to check for < 1.17.5-gke.9 for skip strings, and added a comment.

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,309 @@
...
+		{
+			version:   "1.1",
+			expectErr: true,
+		},
+		{
+			version:   "1.18",
+			expectErr: true,
+		},
+		{
+			version:   "1.18-gke.0",
```
mattcary, line 93: Add a negative test for something like 1.2.3.4 to make it clear how the gke suffix works.

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,309 @@
...
+			version:   "1.18-alpha.3.673+73326ef01d2d7c",
+			expectErr: true,
+		},
+		{
+			version:   "1.18.3-alpha.3.673+73326ef01d2d7c",
+			expectErr: true,
+		},
+	}
+
+	fmt.Println("Num test cases:", len(tests))
```
mattcary, line 118: Aren't these fmt's a little noisy?

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,309 @@
...
+		if err == nil && tc.expectErr {
+			t.Fatal("Got no error but expected one")
+			continue
+		}
+
+		if gotV.version[0] != tc.expectedV.version[0] ||
+			gotV.version[1] != tc.expectedV.version[1] ||
+			gotV.version[2] != tc.expectedV.version[2] ||
+			gotV.version[3] != tc.expectedV.version[3] {
+			t.Fatalf("Got version: %s, expected: %s", gotV.String(), tc.expectedV.String())
```
mattcary, line 138: Adding the test index might make tracking down errors easier if there's ambiguity in the expected/got output.

msau42: A test pattern I've seen is to use `t.Run` and which gives each test case a descriptive name. Then you also don't have the issue where a `Fatalf` in one test case aborts the remaining test cases

test/k8s-integration/version.go
verult: Sorry I meant whether we need `extra` inside the struct rather than a temporary var inside a function

test/k8s-integration/version_test.go
```diff
@@ -0,0 +1,309 @@
...
+		},
+		{
+			leftVersion:  "1.18.0",
+			rightVersion: "1.18.0",
+		},
+		{
+			leftVersion:  "1.18.0-gke.0",
+			rightVersion: "1.18.0-gke.0",
+		},
+		// Negative test cases.(right > left)
```
verult, line 242: nit: Should there be the same number of right > left cases as left < right?

Issue #516, opened 2020-05-28
Race when running multiple controller replicas


ffilippopoulos:
We are running the following configuration:
kube `v1.17.3`
os-image `Flatcar Container Linux by Kinvolk 2512.2.0 (Oklo)`
kernel `4.19.124-flatcar`
docker `docker://18.6.3`
and deploy csi driver from `github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/deploy/kubernetes/overlays/stable?ref=v0.7.0`

We decided to use 2 replicas for the controller, since we saw that this is a statefulset and assumed it will be supported (we couldn't find any relevant doc if that is allowed)
```
csi-gce-pd-controller-0                                     4/4     Running   5          45h
csi-gce-pd-controller-1                                     4/4     Running   5          45h
```

That setup sporadically results in containers stuck at `ContainerCreating` state as the kube sees a `Multi-Attach` error:

```
53s         Warning   FailedAttachVolume             pod/prometheus-1                          Multi-Attach error for volume "pvc-72f6c84c-6e0b-4e33-90e8-8ad22bdb653b" Volume is already exclusively attach
ed to one node and can't be attached to another                                                                                                                                                             
29s         Warning   FailedAttachVolume             pod/prometheus-1                          AttachVolume.Attach failed for volume "pvc-72f6c84c-6e0b-4e33-90e8-8ad22bdb653b" : rpc error: code = Internal
 desc = unknown Attach error: failed when waiting for zonal op: operation operation-1590671984460-5a6b52e968410-035a9d7d-a615df1e failed (RESOURCE_IN_USE_BY_ANOTHER_RESOURCE): The disk resource 'projects/
uw-dev/zones/europe-west2-c/disks/pvc-72f6c84c-6e0b-4e33-90e8-8ad22bdb653b' is already being used by 'projects/uw-dev/zones/europe-west2-c/instances/worker-k8s-pbdr'
```

We have noticed that these logs come from one of the two controllers (the one that issues the attach command second):
```
csi-gce-pd-controller-1 csi-attacher I0528 13:17:12.188224       1 csi_handler.go:99] Error processing "csi-8fb2cee8e7ba0ead8206f1ba5a8f66d3d6b8273fc110e9be399eec76355051ca": failed to attach: rpc error: 
[...]

PR #518, opened 2020-06-03
Adding leader election to controllers in staging rc

verult:
**What type of PR is this?**
/kind feature

First step in releasing change for #516

```release-note
NONE
```

/assign @msau42 
/cc ffilippopoulos

### DIFF
--- a/deploy/kubernetes/overlays/prow-gke-release-staging-rc/kustomization.yaml
+++ b/deploy/kubernetes/overlays/prow-gke-release-staging-rc/kustomization.yaml
@@ -21,3 +21,20 @@ images:
 - name: gke.gcr.io/csi-snapshotter
   newName: gcr.io/gke-release-staging/csi-snapshotter
   newTag: "v2.1.1-gke.0"
+
+resources:
+- leader-election-rbac.yaml
+
+patchesJson6902:	
+- target:	
+    group: apps	
+    version: v1	
+    kind: StatefulSet	
+    name: csi-gce-pd-controller	
+  path: enable-leader-election.yaml
+- target:	
+    group: apps	
+    version: v1	
+    kind: StatefulSet	
+    name: csi-gce-pd-controller	
+  path: statefulset-to-deployment.yaml
--- a/deploy/kubernetes/overlays/prow-gke-release-staging-rc/enable-leader-election.yaml
+++ b/deploy/kubernetes/overlays/prow-gke-release-staging-rc/enable-leader-election.yaml
@@ -1,0 +1,74 @@
+# for external-provisioner
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: "--enable-leader-election"
+
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: "--leader-election-type=leases"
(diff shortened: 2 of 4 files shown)


verult:

Only one test failed, so assuming it's a flake

deploy/kubernetes/overlays/prow-gke-release-staging-rc/kustomization.yaml
```diff
@@ -21,3 +21,14 @@ images:
...
   newTag: "v2.1.1-gke.0"
+
+resources:
+- leader-election-rbac.yaml
+
+patchesJson6902:	
+- target:	
+    group: apps	
+    version: v1	
+    kind: StatefulSet
```
msau42, line 32: Do we want to switch to a Deployment as well? I think it's not easily done without modifying the base too

verult: I'm leaning towards optimizing for the common case of 1 replica. I think this is the common case, but if we have indicators otherwise we could change it

msau42: Maybe let's set the pod management policy to "parallel" then? That way the statefulset won't get stuck if one of the replicas fail.

verult: Actually looking at the [design doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md#cluster-level-deployment) again, StatefulSet is good for a dedicated node, which is probably usually not the case for this deployment. I can switch it to Deployment.

At some point I managed to change the `kind` with a Kustomize JSON patch. Will see if I can still do it :)

msau42: I think this low risk enough to just do it in the base. There's probably other spec differences between StatefulSet and Deployment too.

verult: Got the patching figured out. I pushed a new version

I believe the default of maxUnavailable = 25% (rounded down, so 0) and maxSurge = 25% (rounded up, so 1) is OK - a new replica will be created first, in standby, before the old replica is deleted.

deploy/kubernetes/overlays/prow-gke-release-staging-rc/kustomization.yaml
```diff
@@ -21,3 +21,20 @@ images:
...
+    version: v1	
+    kind: StatefulSet	
+    name: csi-gce-pd-controller	
+  path: enable-leader-election.yaml
+- target:	
+    group: apps	
+    version: v1	
+    kind: StatefulSet	
+    name: csi-gce-pd-controller	
+  path: statefulset-to-deployment.yaml
```
msau42, line 40: File is missing

verult: agh sorry, added

PR #520, opened 2020-06-04
New kustomization for pd driver

jingxu97:
This can work for both linux and windows.
The current structure for kustomization is as follows

-  base: it contains the setup that is common to different driver versions.
    *    controller_setup: includes cluster setup and controller yaml files
    *    node_setup:
           *  Linux: includes node yaml file and related setting that is only applicable for Linux
           *  Windows: includes node yaml file and related setting that is only applicable for Windows
- images: it has a list of images for different versions
    *   stable: currently has only image list for Linux stable version
    *   alpha: besides image list in stable, also add windows image updates
    *   dev: based on alpha, and also contains user's image
    *   prow-xxx: image list used for prow tests. Linux only right now.
-  overlays: it has the version specific setup.
    *    alpha: kustomize will generate alpha version of the drivers which includes both Linux and Windows
    *    stable: kustomize will generate alpha version of the drivers which only has Linux.  
    *    dev: this is based on alpha version and images are updated with appropriate versions. 
    *    prow-gke-release-xxx: this is based on stable version and images are updated with appropriate versions. It is used for CI testing only.

This PR also updates kustomize latest versino. "bases" is deprecated, use "resources" instead.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/deploy-driver.sh
+++ b/deploy/kubernetes/deploy-driver.sh
@@ -97,4 +97,3 @@ ${KUBECTL} version
 readonly tmp_spec=/tmp/gcp-compute-persistent-disk-csi-driver-specs-generated.yaml
 ${KUSTOMIZE_PATH} build ${PKGDIR}/deploy/kubernetes/overlays/${DEPLOY_VERSION} | tee $tmp_spec
 ${KUBECTL} apply -v="${VERBOSITY}" -f $tmp_spec
-
--- a/deploy/kubernetes/images/alpha/image.yaml
+++ b/deploy/kubernetes/images/alpha/image.yaml
@@ -1,0 +1,18 @@
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-gcepd-driver-alpha-win
+imageTag:
+  name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver-win
+  newName: gcr.io/jing-k8s-dev/gce-pd-windows-2019
+  newTag: "0.2.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-node-registrar-win
+imageTag:
+  name: gke.gcr.io/k8s-staging-csi/csi-node-driver-registrar-win
+  newName: gcr.io/k8s-staging-csi/csi-node-driver-registrar
+  newTag: "amd64-windows-v20200428-v1.3.0-26-g510710d5"
--- a/deploy/kubernetes/images/dev/kustomization.yaml
+++ b/deploy/kubernetes/images/dev/kustomization.yaml
@@ -1,0 +1,5 @@
+namespace:
+  gce-pd-csi-driver
+resources:
+- ../alpha/
+- image.yaml
--- a/deploy/kubernetes/images/stable/image.yaml
+++ b/deploy/kubernetes/images/stable/image.yaml
@@ -1,0 +1,55 @@
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-provisioner
+imageTag:
+  name: gke.gcr.io/csi-provisioner
+  newTag: "v1.6.0-gke.0"
+
+---
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-attacher
+imageTag:
+  name: gke.gcr.io/csi-attacher
+  newTag: "v2.2.0-gke.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-resizer
+imageTag:
+  name: gke.gcr.io/csi-resizer
+  newTag: "v0.5.0-gke.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-snapshotter
+imageTag:
+  name: gke.gcr.io/csi-snapshotter
+  newTag: "v2.1.1-gke.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-node-registrar
+imageTag:
+  name: gke.gcr.io/csi-node-driver-registrar
+  newTag: "v1.3.0-gke.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-gcepd-driver
+imageTag:
+  name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
+  # Don't change stable image without changing pdImagePlaceholder in
+  # test/k8s-integration/main.go
+  newTag: "v0.7.0-gke.0"
+---
--- a/deploy/kubernetes/overlays/stable/kustomization.yaml
+++ b/deploy/kubernetes/overlays/stable/kustomization.yaml
@@ -1,25 +1,9 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-bases:
-- ../../base
-images:
-- name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
-  # Don't change stable image without changing pdImagePlaceholder in
-  # test/k8s-integration/main.go
-  newName: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
-  newTag: "v0.7.0-gke.0"
-- name: gke.gcr.io/csi-provisioner
-  newName: gke.gcr.io/csi-provisioner
-  newTag: "v1.6.0-gke.0"
-- name: gke.gcr.io/csi-attacher
-  newName: gke.gcr.io/csi-attacher
-  newTag: "v2.2.0-gke.0"
-- name: gke.gcr.io/csi-node-driver-registrar
-  newName: gke.gcr.io/csi-node-driver-registrar
-  newTag: "v1.3.0-gke.0"
-- name: gke.gcr.io/csi-resizer
-  newName: gke.gcr.io/csi-resizer
-  newTag: "v0.5.0-gke.0"
-- name: gke.gcr.io/csi-snapshotter
-  newName: gke.gcr.io/csi-snapshotter
-  newTag: "v2.1.1-gke.0"
+namespace:
+  gce-pd-csi-driver
+resources:
+- ../../base/controller
+- ../../base/node_linux
+transformers:
+- ../../images/stable
--- a/deploy/kubernetes/base/controller/cluster_setup.yaml
+++ b/deploy/kubernetes/base/controller/cluster_setup.yaml
@@ -147,33 +147,6 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 
 ---
-apiVersion: policy/v1beta1
-kind: PodSecurityPolicy
-metadata:
-  name: csi-gce-pd-node-psp
-spec:
-  seLinux:
-    rule: RunAsAny
-  supplementalGroups:
-    rule: RunAsAny
-  runAsUser:
-    rule: RunAsAny
-  fsGroup:
-    rule: RunAsAny
-  privileged: true
-  volumes:
-  - '*'
-  hostNetwork: true
-  allowedHostPaths:
-  - pathPrefix: "/var/lib/kubelet/plugins_registry/"
-  - pathPrefix: "/var/lib/kubelet"
-  - pathPrefix: "/var/lib/kubelet/plugins/pd.csi.storage.gke.io/"
-  - pathPrefix: "/dev"
-  - pathPrefix: "/etc/udev"
-  - pathPrefix: "/lib/udev"
-  - pathPrefix: "/run/udev"
-  - pathPrefix: "/sys"
----
 
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
@@ -187,6 +160,18 @@ rules:
     - csi-gce-pd-node-psp
 ---
 
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: csi-gce-pd-node-deploy-win
+rules:
+  - apiGroups: ['policy']
+    resources: ['podsecuritypolicies']
+    verbs:     ['use']
+    resourceNames:
+    - csi-gce-pd-node-psp-win
+---
+
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
@@ -198,6 +183,32 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: csi-gce-pd-node-sa
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: csi-gce-pd-node-win
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: csi-gce-pd-node-deploy-win
+subjects:
+- kind: ServiceAccount
+  name: csi-gce-pd-node-sa
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: csi-gce-pd-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: csi-gce-pd-node-deploy
+subjects:
+- kind: ServiceAccount
+  name: csi-gce-pd-controller-sa
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
@@ -231,7 +242,6 @@ roleRef:
   kind: ClusterRole
   name: csi-gce-pd-snapshotter-role
   apiGroup: rbac.authorization.k8s.io
-
 ---
 
 kind: Role
@@ -262,3 +272,4 @@ roleRef:
   kind: Role
   name: csi-gce-pd-leaderelection-role
   apiGroup: rbac.authorization.k8s.io
+
--- a/deploy/kubernetes/base/controller/controller.yaml
+++ b/deploy/kubernetes/base/controller/controller.yaml
@@ -17,6 +17,8 @@ spec:
       # this requirement when issue is resolved and before any exposure of
       # metrics ports
       hostNetwork: true
+      nodeSelector:
+        kubernetes.io/os: linux
       serviceAccountName: csi-gce-pd-controller-sa
       priorityClassName: csi-gce-pd-controller
       containers:
--- a/deploy/kubernetes/base/controller/kustomization.yaml
+++ b/deploy/kubernetes/base/controller/kustomization.yaml
@@ -1,6 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
+namespace:
+  gce-pd-csi-driver
 resources:
-- ../stable
-transformers:
-- ../../images/prow-gke-release-staging-head
+- cluster_setup.yaml
+- controller.yaml
--- a/deploy/kubernetes/base/kustomization.yaml
+++ b/deploy/kubernetes/base/kustomization.yaml
@@ -1,8 +1,8 @@
-commonLabels:
-  k8s-app: gcp-compute-persistent-disk-csi-driver
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 namespace:
   gce-pd-csi-driver
 resources:
-- node.yaml
-- controller.yaml
-- setup-cluster.yaml
+- controller
+- node_linux
+- node_windows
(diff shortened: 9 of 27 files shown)


jingxu97:
cc @msau42 @verult

deploy/kubernetes/deploy-driver.sh
```diff
@@ -95,6 +96,36 @@ fi
 ${KUBECTL} version
 
 readonly tmp_spec=/tmp/gcp-compute-persistent-disk-csi-driver-specs-generated.yaml
-${KUSTOMIZE_PATH} build ${PKGDIR}/deploy/kubernetes/overlays/${DEPLOY_VERSION} | tee $tmp_spec
-${KUBECTL} apply -v="${VERBOSITY}" -f $tmp_spec
 
+if [[ ${DEPLOY_OS} = "mixed" ]]; then
```
msau42, line 100: I would suggest to simplify things and just always include both linux and windows node specs. I guess we may have a problem in the beginning where the windows spec is not available for staging/stable.

What if the kustomization.yaml in  image_setup/{dev/staging/stable} apppropriately pulls in node_base/windows as needed?

Then we don't actually need to auto-generate a top level kustomization.yaml and can continue to use the appropriate image_setup/X/kustomization.yaml

deploy/kubernetes/kustomization/image_setup/alpha/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
msau42, line 1: naming nit: we have more than just image version patches in these directories. We may enable new features as well.

Maybe I would still keep the directory name as "overlays". So the main change is that the node deployment is separated from controller deployment, but we still use the kustomization.yaml from the "overlays" directory to determine which version of the driver to deploy.

deploy/kubernetes/kustomization/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace:
+  gce-pd-csi-driver
+bases:
+- base_setup
```
msau42, line 6: Maybe call this one controller_setup

deploy/kubernetes/deploy-driver.sh
jingxu97: At first, I thought with multiple bases, it is easy to have arbitrary combination, e.g. linux + stable_image, windows + alpha_image. But the image part does not work as expected in multiple bases. I will simply this with one combined driver instead of having a complicated script here.

deploy/kubernetes/kustomization/image_setup/alpha/kustomization.yaml
jingxu97: updated.

deploy/kubernetes/kustomization/kustomization.yaml
jingxu97: updated

deploy/kubernetes/base/controller_setup/controller.yaml
```diff
@@ -82,5 +84,10 @@ spec:
         - name: cloud-sa-volume
           secret:
             secretName: cloud-sa
+      # https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
```
msau42, line 87: From the other PR, I think we should not include this otherwise the controller won't restart if the node becomes unhealthy. It's only really meant for DaemonSets.

deploy/kubernetes/base/node_setup/windows/node.yaml
```diff
@@ -0,0 +1,82 @@
+#TODO(#40): Force DaemonSet to not run on master.
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: csi-gce-pd-node-win
+spec:
+  selector:
+    matchLabels:
+      app: gcp-compute-persistent-disk-csi-driver
```
msau42, line 9: We may need linux and windows DaemonSets to use different labels. Otherwise the linux DaemonSet may start managing the windows Pods, and vice versa.

deploy/kubernetes/overlays/stable/kustomization.yaml
```diff
@@ -1,25 +1,27 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-bases:
-- ../../base
+namespace:
+  gce-pd-csi-driver
 images:
-- name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
   # Don't change stable image without changing pdImagePlaceholder in
```
msau42, line 6: Keep this comment with gcp-compute-persistent-disk-csi-driver

deploy/kubernetes/base/node_setup/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace:
+  gce-pd-csi-driver
+resources:
+- linux/node.yaml
+- windows/node.yaml
```
msau42, line 7: This means windows deployment automatically gets added to stable. Is that what we want? Or should we first add windows to the alpha overlay, and eventually when it becomes stable, we add it to the base?

msau42: Another alternative is to have the overlays directly specify the os/node.yaml resources that they want.

deploy/kubernetes/base/controller_setup/controller.yaml
jingxu97: right, forgot to update it.

deploy/kubernetes/overlays/stable/kustomization.yaml
jingxu97: fixed

deploy/kubernetes/base/node_setup/kustomization.yaml
jingxu97: Thanks to @Liujingfang1, we can make image update lists reusable for different versions. I think with this change, it can customize every possible way easily.

deploy/kubernetes/base/node_setup/windows/node.yaml
jingxu97: updated.

deploy/kubernetes/base/images/image.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: builtin
+kind: ImageTagTransformer
```
msau42, line 2: What is the difference between using this and the `images` field that's in the kustomization?

deploy/kubernetes/base/node_setup/windows/node.yaml
```diff
@@ -0,0 +1,82 @@
...
+      # metrics ports. But hostNetwork is not working for Windows, might be an issue
+      # when deploying on GKE Windows node.
+      # hostNetwork: true
+      priorityClassName: csi-gce-pd-node
+      serviceAccountName: csi-gce-pd-node-sa
+      nodeSelector:
+        kubernetes.io/os: windows
+      containers:
+        - name: csi-driver-registrar
+          image: gcr.io/k8s-staging-csi/csi-node-driver-registrar:amd64-windows-v20200428-v1.3.0-26-g510710d5
```
msau42, line 27: These images should be substituted in the alpha overlay

deploy/kubernetes/base/images/image.yaml
jingxu97: Before, we have use append these images with some base. For example, in previous structure, we list most of images in stable version. And stable version needs a base, which point to the base that has both linux and windows setup. Now if all other versions e.g., alpha, dev need to base on stable, all of them must share the same base setup as stable (which contains both linux and windows.). If for alpha, we only want to have linux as base instead of using stable as a base, we have to copy the same image list to the alpha version. 
Now with this ImageTagTransformer, we can do any combination easily. The image is just like a resource to use. We can allow alpha to use mixed setup as base, plus images, while stable to use only linux setup, plus images.

deploy/kubernetes/base/node_setup/windows/node.yaml
jingxu97: we don't have another image yet, I think. Currently I see this is in staging. We need to check when it is able to release.

deploy/kubernetes/base/images/image.yaml
msau42: > And stable version needs a base, which point to the base that has both linux and windows setup.

Isn't it the opposite? Stable has just linux, and alpha has linux + windows? So everything can still continue to be off of stable, and then override the images. Unless the issue is you can't have 1 kustomization.yaml with `images` that uses another kustomization.yaml that also has `images` as its base?

deploy/kubernetes/base/node_setup/windows/node.yaml
msau42: Yes, but I think we should follow the pattern of the linux node.yaml and not specify a tag in the base, and instead override the tag in the kustomization.yaml

deploy/kubernetes/base/images/image.yaml
jingxu97: I think I used the wrong example. My main point is this way we can have arbitrary combination  of os + image as needed. This will give the flexibility in case we do want e.g., windows+image for security setup reason (we don't want to have linux security set up mix with it). And putting all common things in the same place seems clean.

jingxu97:
updated the PR. PTAL

deploy/kubernetes/base/images/image.yaml
msau42: The tags in this file are only for the "stable" overlay. If we want to follow this model, should we have a file for the images in every overlay?

deploy/kubernetes/base/images/image.yaml
```diff
@@ -0,0 +1,78 @@
...
+metadata:
+  name: imagetag9
+imageTag:
+  name: gcr.io/k8s-staging-csi/csi-node-driver-registrar-win
+  # Don't change stable image without changing pdImagePlaceholder in
+  # test/k8s-integration/main.go
+  newName: gcr.io/k8s-staging-csi/csi-node-driver-registrar
+  newTag: "amd64-windows-v20200428-v1.3.0-26-g510710d5"
```
msau42, line 78: This should go in an image file that is for alpha clusters right?

deploy/kubernetes/base/node_setup/linux/node.yaml
```diff
@@ -15,21 +15,20 @@ spec:
       # Host network must be used for interaction with Workload Identity in GKE
       # since it replaces GCE Metadata Server with GKE Metadata Server. Remove
       # this requirement when issue is resolved and before any exposure of
-      # metrics ports.
+      # metrics ports. But hostNetwork is not working for Windows, might be an issue
```
msau42, line 18: Do we need the windows comment in the linux file?

deploy/kubernetes/base/node_setup/windows/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace:
+  gce-pd-csi-driver
+resources:
+- node.yaml
+- psp.yaml
```
msau42, line 7: Do we want psp.yaml for linux too?

deploy/kubernetes/base/node_setup/linux/node.yaml
jingxu97: removed it.

deploy/kubernetes/overlays/prow-gke-release-staging-head/kustomization.yaml
```diff
@@ -1,23 +1,7 @@
...
-  newName: quay.io/k8scsi/csi-node-driver-registrar
-  newTag: "canary"
-- name: gke.gcr.io/csi-resizer
-  newName: quay.io/k8scsi/csi-resizer
-  newTag: "canary"
-- name: gke.gcr.io/csi-snapshotter
-  newName: quay.io/k8scsi/csi-snapshotter
-  newTag: "canary"
+resources:
+- ../../base/controller_setup
```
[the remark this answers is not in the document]
jingxu97: yes, updated

msau42:
/approve
Would like to get at least one more set of eyes on this since it's changing the structure a bit. Also I still had one outsanding question on the ClusterRole

deploy/kubernetes/images/alpha/image.yaml
```diff
@@ -0,0 +1,22 @@
...
+  newName: gcr.io/jing-k8s-dev/gce-pd-windows-2019
+  newTag: "0.2.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-node-registrar-win
+imageTag:
+  name: gcr.io/k8s-staging-csi/csi-node-driver-registrar-win
```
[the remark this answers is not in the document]
jingxu97: yes, updated

deploy/kubernetes/base/node_setup/linux/psp.yaml
```diff
@@ -0,0 +1,27 @@
...
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  privileged: true
+  volumes:
+  - '*'
```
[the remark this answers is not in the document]
jingxu97: got it.

jingxu97:
finally, the test passed. PTAL @msau42

deploy/kubernetes/base/node_setup/linux/psp.yaml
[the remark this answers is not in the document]
jingxu97: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/524

deploy/kubernetes/base/controller_setup/cluster_setup.yaml
```diff
@@ -187,6 +160,18 @@ rules:
...
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: csi-gce-pd-node-deploy-win
+rules:
+  - apiGroups: ['policy']
+    resources: ['podsecuritypolicies']
+    verbs:     ['use']
+    resourceNames:
+    - csi-gce-pd-node-psp-win
```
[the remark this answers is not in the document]
jingxu97: I tried it, it is ok.
I was thinking about it, but think putting cluster role and binding in once place is also good. Either way is fine for me.

jingxu97:
@verult could you please review this again since lots of changes from last time.

deploy/kubernetes/images/dev/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+namespace:
+  gce-pd-csi-driver
+resources:
+- image.yaml
+- ../alpha/
```
verult, line 5: If we ever need to change the tag for node-driver-registrar in the dev overlay, should it use the image name in base or alpha?

jingxu97: if it is only for dev, then change the image in the image.yaml file under /dev

jingxu97:
Previously, we have the following structure. Compared to the current one, the main problem here is linux and windows cannot share the image list (stable). With current structure, image list can be shared. Basically once it is setup, future update do not need to touch overlays. For changes that are related to controller or cluster_setup, it does to base/controller_setup. Any change related to node yaml files go to base/node_setup (linux or windows). Any changes related to images go to images (stable, alpha, dev etc.). In overlay, we could set up a overlay that is a combination of base + node_setup + image. e.g,
base + node/linux + image/dev
base + node/windows + image/alpha. 


- base
    *. controller.yaml
    *. setup-cluster.yaml
- overlays
    * linux
         * base: psp.yaml, node.yaml
         * stable: image list
         * dev: yaml and image list
         * alpha: dev
         ..
    * windows
          * base: psp.yaml, node.yaml
          * stable: image list

deploy/kubernetes/images/stable/image.yaml
```diff
@@ -0,0 +1,55 @@
...
+  newTag: "v1.3.0-gke.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-gcepd-driver
+imageTag:
+  name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
+  # Don't change stable image without changing pdImagePlaceholder in
```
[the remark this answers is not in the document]
mattcary: Let me think out loud briefly. I haven't used kustomize much so I could be missing something.

Do we want the yaml in base/ to be meaningful, or is it only supported by deploying through customize? In that case we could replace the images in base/ with placeholders like ATTACHER_IMAGE and have a newName in all of the image transformers. Then it's clear what is being changed.

With respect to the ordering of multiple changes, though, it looks like there's a bug in kustomize. If one applies two patches that refer to the same field, kustomize errors. But if one applies multiple image transformations, the last one wins (as defined by the list in the kustomization.yaml in the transformer directory).

So that's unfortunate... I can ask in the kustomize repo to see if that really is a bug or for some reason WAI. From our perspective, we just have to be careful I guess, and trace the order of things from kustomization files starting in the overlay.

deploy/kubernetes/images/stable/image.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-rpovisioner
```
mattcary, line 4: typo: rpovisioner

deploy/kubernetes/images/stable/image.yaml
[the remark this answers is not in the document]
jingxu97: I think using placeholders like ATTACHER_IMAGE sounds good to me. All the image name placeholder used in /base should not be touched.  
All the image replacement will be done in /images dir. For testing, we only need to touch/replace /images/dev/image.yaml file to create ImageTagTransformer when necessary.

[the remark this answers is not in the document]
jingxu97: Also I suggest to modify placeholder in a different PR. This PR has already gone through lots of changes.

deploy/kubernetes/images/alpha/image.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-gcepd-driver-alpha-win
+imageTag:
+  name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver-win
+  # Don't change stable image without changing pdImagePlaceholder in
```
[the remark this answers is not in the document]
jingxu97: removed it

deploy/kubernetes/images/alpha/image.yaml
```diff
@@ -0,0 +1,22 @@
...
+  newTag: "0.2.0"
+---
+
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-node-registrar-win
+imageTag:
+  name: gke.gcr.io/k8s-staging-csi/csi-node-driver-registrar-win
+  # Don't change stable image without changing pdImagePlaceholder in
```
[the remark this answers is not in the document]
jingxu97: removed

deploy/kubernetes/images/stable/image.yaml
jingxu97: fixed

deploy/kubernetes/images/dev/kustomization.yaml
jingxu97: there is no image name gke.gcr.io/gcp-compute-persistent-disk-csi-driver-win (this is for windows) in base (stable). 
Generally, the following can happen
in stable, change image name --> nameA
in alpha, change image name --> nameB
if alpha is also based on stable, I think the final image name will be nameA.
Or,
in stable, change image name --> nameA
in alpha, change image nameA --> nameB
if alpha is also based on stable, I think the final image name will be nameB.

This is thought ImageTagTransformer. I also tested using both ImageTagTransformer and images section directly in kustomization.yaml. The latter (images section) seems taking the final effect.
There are so many ways to construct this, we just have to try it out.

deploy/kubernetes/images/stable/image.yaml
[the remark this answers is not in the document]
mattcary: SGTM for doing that in a different PR

mattcary:
/lgtm

If there's agreement on putting the image placeholders in a follow-on CL I'm happy enough with this now.

deploy/kubernetes/images/stable/image.yaml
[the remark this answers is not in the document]
mattcary: I've filed https://github.com/kubernetes-sigs/kustomize/issues/2590 about conflict with image transformers.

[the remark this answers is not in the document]
jingxu97: Thanks! I cc @Liujingfang1 who is the expert in this area.

PR #527, opened 2020-06-12
Return success in CreateVolume when disk is READY

### LINKED ISSUE #526: PD CSI Driver CreateVolume should wait for Disk to reach READY status before returning success
For CreateVolume,
1. if an existing disk is found a success is directly returned without checking ready status:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-pd-csi-driver/controller.go#L140 

2. Similarly at the end of the operation we should check for ready status.
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L304
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L323

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

In PD CSI Driver CreateVolume call, wait for disk to reach a READY status, before returning success to caller.

Fixes #526

```release-note
In GCE PersistentDisk CSI Driver CreateVolume call, wait for disk to reach a READY status, before returning success to caller.

```

### DIFF
--- a/pkg/gce-cloud-provider/compute/cloud-disk.go
+++ b/pkg/gce-cloud-provider/compute/cloud-disk.go
@@ -92,6 +92,17 @@ func (d *CloudDisk) GetKind() string {
 	}
 }
 
+func (d *CloudDisk) GetStatus() string {
+	switch d.Type() {
+	case Zonal:
+		return d.ZonalDisk.Status
+	case Regional:
+		return d.RegionalDisk.Status
+	default:
+		return "Unknown"
+	}
+}
+
 // GetPDType returns the type of the PD as either 'pd-standard' or 'pd-ssd' The
 // "Type" field on the compute disk is stored as a url like
 // projects/project/zones/zone/diskTypes/pd-standard
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -62,6 +62,29 @@ const (
 	replicationTypeRegionalPD = "regional-pd"
 )
 
+func isDiskReady(disk *gce.CloudDisk) (bool, error) {
+	status := disk.GetStatus()
+	switch status {
+	case "READY":
+		return true, nil
+	case "FAILED":
+		return false, fmt.Errorf("Disk %s status is FAILED", disk.GetName())
+	case "CREATING":
+		klog.V(4).Infof("Disk %s status is CREATING", disk.GetName())
+		return false, nil
+	case "DELETING":
+		klog.V(4).Infof("Disk %s status is DELETING", disk.GetName())
+		return false, nil
+	case "RESTORING":
+		klog.V(4).Infof("Disk %s status is RESTORING", disk.GetName())
+		return false, nil
+	default:
+		klog.V(4).Infof("Disk %s status is: %s", disk.GetName(), status)
+	}
+
+	return false, nil
+}
+
 func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	var err error
 	// Validate arguments
@@ -143,6 +166,16 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		if err != nil {
 			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("CreateVolume disk already exists with same name and is incompatible: %v", err))
 		}
+
+		ready, err := isDiskReady(existingDisk)
+		if err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("CreateVolume disk %v had error checking ready status: %v", volKey, err))
+		}
+
+		if !ready {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("CreateVolume disk %v is not ready", volKey))
+		}
+
 		// If there is no validation error, immediately return success
 		klog.V(4).Infof("CreateVolume succeeded for disk %v, it already exists and was compatible", volKey)
 		return generateCreateVolumeResponse(existingDisk, capBytes, zones), nil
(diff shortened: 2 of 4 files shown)


msau42:
/ok-to-test
will let @mattcary review first :-)

pkg/gce-cloud-provider/compute/cloud-disk.go
```diff
@@ -92,6 +92,17 @@ func (d *CloudDisk) GetKind() string {
...
 }
 
+func (d *CloudDisk) GetStatus() string {
+	switch d.Type() {
+	case Zonal:
+		return d.ZonalDisk.Status
+	case Regional:
+		return d.RegionalDisk.Status
+	default:
+		return ""
```
mattcary, line 102: nit: this will end up with an obscure error message in isDiskReady. Maybe something more informative could be returned here?

saikat-royc:
UTs and sanity test needs to be looked into, before final review

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -187,6 +220,15 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 	default:
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", params.ReplicationType))
 	}
+
+	ready, err := isDiskReady(disk)
```
msau42, line 224: Can we potentially extend `waitForOp` instead, so that we don't immediately fail? I imagine this delay will happen frequently anytime someone restores from a snapshot.

saikat-royc: Will look into it. I was following the snapshot model of ready check.

msau42: Ah the snapshot api specifically says that it should return once it's cut and before it's ready, especially if it would take a signficant amount of time to get ready. But the createvolume api doesn't have a requirement like that.

saikat-royc: Looking into the code and test logs, insert call returns an insert op, that the driver polls to reach completion. Here is the timeout behavior:
The timeout for the insert operation polling is 5 min (polled every 3 secs).
However the context is provided a deadline of 10 secs by the caller (csi-provisioner side car in this case). Which means the poll operation itself gets cancelled at the end of 10 secs.

The current implementation and placement of isDiskReady(in CreateVolume call) looks to me the right place, as its a common place for both regional/zonal disks.

Possible changes in external provisioner to reduce number of errors:
We can look at changing the timeout to a larger value (in the order of minutes say 1-2 mins). An additional optimization can be done to use a large timeout only for the scenario where the a volume datasource is provided for provisioning (https://github.com/saikat-royc/external-provisioner/blob/master/pkg/controller/controller.go#L459).
To keep the external provisioner code generic, a new flag (say provisionFromDatasourceOperationTimeout = 10 secs default) can be exposed (https://github.com/saikat-royc/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L65) and set in the PD CSI driver pod spec (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/base/controller/controller.yaml#L34)

Let me know your thoughts? @msau42 @mattcary

msau42: I think we can use the same timeout value for all operations that the provisioner calls. 10 seconds is probably too short.

mattcary: +1 Increasing the timeout is straightforward.

Can we get stats from GKE clusters to see how long attach time typically is? (not for snapshots, just to get a sense of how often these errors are happening)

msau42: Saikat explained to me that the scenario we're seeing right now is that the waitforop times out after 10 seconds and then we run into the "check if disk already exists" logic that returns immediately. And if we increase the timeout then the waitForOp here should not return until the disk is ready. So given that, I think increasing the timeout is sufficient and we don't need to add an additional wait for ready (returning error is fine).

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -143,6 +166,16 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		if err != nil {
 			return nil, status.Error(codes.AlreadyExists, fmt.Sprintf("CreateVolume disk already exists with same name and is incompatible: %v", err))
 		}
+
+		ready, err := isDiskReady(existingDisk)
```
msau42, line 170: Can we add some unit tests for the various scenarios? The fake can probably be extended so that we can toggle the ready status.

saikat-royc: Done

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1549,3 +1549,111 @@ func TestVolumeOperationConcurrency(t *testing.T) {
...
+				VolumeId:           testVolumeID,
+				VolumeContext:      nil,
+				AccessibleTopology: stdTopology,
+			},
+		},
+	}
+
+	// Run test cases
+	for _, tc := range testCases {
+		t.Logf("test case: %s", tc.name)
```
msau42, line 1626: Can we wrap this in t.Run()? That way the Fatals below won't cause the later test cases to be skipped.

saikat-royc: Done

PR #533, opened 2020-06-24
Update golang version in dockerfile

verult:
Fixes failing tests

```release-note
NONE
```

/assign @mattcary @msau42 @saad-ali @saikat-royc

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 
 
 
-FROM golang:1.13.4-alpine3.10 as builder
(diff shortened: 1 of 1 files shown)


verult:
Not sure if this PR is necessary - the test failure is due to compiling kubernetes/kubernetes. Will wait until that change is in.

If the Dockerfile is updated, go.mod may need to be updated too

verult:
Not aware of any failures since the fix went in.

/close

PR #542, opened 2020-07-01
Increase PD CSI sidecar operation timeout

### LINKED ISSUE #541: Increase operation timeouts for csi sidecars
While fixing kubernetes-sigs/gcp-compute-persistent-disk-csi-driver#526, observed that Create Volume from a snapshot datasource takes time to get provisioned.

following behavior in provisioning seen in GCE PD CSI driver:
there are 2 cases for a CreateVolume call (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-pd-csi-driver/controller.go#L88):

disk already found. we check for ready status and return error if not ready.
no existing disk found, so we trigger an insert op and wait for the op to complete (timeout for the poll operation for insert op is 5 mins). The wait for op gets cancelled due to a short deadline of set in the context (10 secs), by the client csi-provisioner sidecar.
By increasing the timeout from 10 secs to the order of minutes, we significantly reduce round-trips os retries due to diskNotReady error.

Provisioner and Attacher sidecars timeout need to be increased.

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #541 

```release-note
Increase provisioner and attacher op timeout to reduce retries
```

### DIFF
--- a/deploy/kubernetes/overlays/prow-gke-release-staging-head/increase-sidecar-operation-timeout.yaml
+++ b/deploy/kubernetes/overlays/prow-gke-release-staging-head/increase-sidecar-operation-timeout.yaml
@@ -1,0 +1,12 @@
+# Increase operation timeout for provisioner sidecar.
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: "--timeout=250s"
+# Increase operation timeout for attacher sidecar.
+- op: add
+  path: /spec/template/spec/containers/1/args/-
+  value: "--timeout=250s"
+# Increase operation timeout for snapshotter sidecar.
+- op: add
+  path: /spec/template/spec/containers/3/args/-
+  value: "--timeout=300s"
--- a/deploy/kubernetes/overlays/prow-gke-release-staging-head/kustomization.yaml
+++ b/deploy/kubernetes/overlays/prow-gke-release-staging-head/kustomization.yaml
@@ -17,3 +17,9 @@ patchesJson6902:
     kind: Deployment
     name: csi-gce-pd-controller
   path: volume-inuse-error-handler.yaml
+- target:
+    group: apps
+    version: v1
+    kind: Deployment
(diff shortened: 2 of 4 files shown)


deploy/kubernetes/overlays/prow-gke-release-staging-head/increase-sidecar-operation-timeout.yaml
```diff
@@ -0,0 +1,8 @@
+# Increase operation timeout for provisioner sidecar.
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: "--timeout=250s"
+# Increase operation timeout for attacher sidecar.
+- op: add
+  path: /spec/template/spec/containers/1/args/-
+  value: "--timeout=250s"
```
msau42, line 8: Do we also need to consider raising the timeout for resizer and snapshotter too?

saikat-royc: Adding snapshotter timeout as per discussions. The resizer timeouts are fine.

deploy/kubernetes/overlays/prow-gke-release-staging-head/increase-sidecar-operation-timeout.yaml
```diff
@@ -0,0 +1,12 @@
+# Increase operation timeout for provisioner sidecar.
```
msau42, line 1: Should we add to staging-rc too?

saikat-royc: ok will do

Issue #140, opened 2018-10-26
Collect driver logs for k8s test


msau42:
We can do something like https://github.com/kubernetes/kubernetes/pull/69986

davidz627:
/remove-lifecycle stale
This would be incredibly helpful for debugging

msau42:
Sample driver logs: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/554/pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration/1283548649288110080/artifacts/pd-csi-driver/

Issue #139, opened 2018-10-26
Validate the driver successfully came up in k8s test


msau42:
Need to deal with the master DaemonSet pod pending

msau42:
https://github.com/kubernetes/kubernetes/pull/93120 may help validate the driver is registered successfully on all nodes.

It would be nice to also have a check that the controller deployment came up successfully

mattcary:
Is that a matter of polling for objects defined in the appropriate configs in cluster/addons?

msau42:
Yes, namely the Deployment in the gce-pd-csi-driver namespace.

mattcary:
/assign

Ok I'll take this.

mattcary:
Actually, the k/k change will not improve the k8s-integration test in this repo as it deploys the driver from a shell script.

I have a PR at https://github.com/mattcary/gcp-compute-persistent-disk-csi-driver/tree/wait (currently stacked up behind #610) that will fix the change in this repo too.

PR #551, opened 2020-07-14
Update CI test configuration

jingxu97:
update ci test configuration

Fix windows integration script

remove env virable GCE_PD_DO_DRIVER_BUILD direct setting

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/driver.go
+++ b/test/k8s-integration/driver.go
@@ -71,9 +71,8 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
 	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", driverNamespace)
 	err = runCommand("Checking driver pods", statusCmd)
 	if err != nil {
-		return fmt.Errorf("failed to check driver pods: %v", err)
+		return fmt.Errorf("failed to describe pods: %v", err)
 	}
-
 	return nil
 }
 
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -298,6 +298,23 @@ func handle() error {
 			}
 		}()
 	}
+	// For windows cluster, when cluster is up, all Windows nodes are tainted with NoSchedule to avoid linux pods
+	// being scheduled to Windows nodes. When running windows tests, we need to remove the taint.
+	if *platform == "windows" {
+		nodesCmd := exec.Command("kubectl", "get", "nodes", "-l", "beta.kubernetes.io/os=windows", "-o", "name")
+		out, err := nodesCmd.CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("failed to get nodes: %v", err)
+		}
+		nodes := strings.Fields(string(out))
+		for _, node := range nodes {
+			taintCmd := exec.Command("kubectl", "taint", "node", node, "node.kubernetes.io/os:NoSchedule-")
+			_, err = taintCmd.CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("failed to untaint windows node %v", err)
+			}
+		}
+	}
 
 	if !*useGKEManagedDriver {
 		// Install the driver and defer its teardown
--- a/test/run-windows-k8s-integration.sh
+++ b/test/run-windows-k8s-integration.sh
@@ -9,22 +9,20 @@
 set -o nounset
 set -o errexit
 
-readonly GCE_PD_DO_DRIVER_BUILD=false
 readonly PKGDIR=${GOPATH}/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
-readonly LOCAL_K8S_DIR=${GOPATH}/src/k8s.io/kubernetes
 readonly overlay_name="${GCE_PD_OVERLAY_NAME:-alpha}"
 readonly do_driver_build="${GCE_PD_DO_DRIVER_BUILD:-true}"
 readonly deployment_strategy=${DEPLOYMENT_STRATEGY:-gce}
 readonly test_version=${TEST_VERSION:-master}
 readonly gce_zone=${GCE_CLUSTER_ZONE:-us-central1-b}
-readonly teardown_driver=${GCE_CLUSTER_ZONE:-true}
+readonly teardown_driver=${GCE_PD_TEARDOWN_DRIVER:-true}
 
 make -C ${PKGDIR} test-k8s-integration
 
 base_cmd="${PKGDIR}/bin/k8s-integration-test \
-            --platform=windows --local-k8s-dir=${LOCAL_K8S_DIR} --bringup-cluster=false --teardown-cluster=false --teardown-driver=${teardown_driver}\
+            --platform=windows --bringup-cluster=false --teardown-cluster=false --teardown-driver=${teardown_driver}\
             --run-in-prow=true --deploy-overlay-name=${overlay_name} --service-account-file=${E2E_GOOGLE_APPLICATION_CREDENTIALS} \
-            --do-driver-build=${do_driver_build} --gce-zone=${gce_zone}\
+            --do-driver-build=${do_driver_build} --gce-zone=${gce_zone} --test-version=${test_version}\
             --storageclass-file=sc-windows.yaml --snapshotclass-file=pd-volumesnapshotclass.yaml --test-focus="External.Storage" \
             --deployment-strategy=${deployment_strategy}"
 


jingxu97:
cc @msau42 @verult

jingxu97:
replace #549

test/k8s-integration/driver.go
```diff
@@ -65,12 +67,18 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
...
 	// TODO (#139): wait for driver to be running
 	time.Sleep(time.Minute)
-	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", "default")
+	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", "gce-pd-csi-driver")
 	err = runCommand("Checking driver pods", statusCmd)
 	if err != nil {
 		return fmt.Errorf("failed to check driver pods: %v", err)
 	}
-
+	time.Sleep(time.Minute)
```
msau42, line 75: Do we need this second check?

test/k8s-integration/main.go
```diff
@@ -296,6 +296,21 @@ func handle() error {
...
 	}
+	if *platform == "windows" {
+		nodesCmd := exec.Command("kubectl", "get", "nodes", "-l", "beta.kubernetes.io/os=windows", "-o", "name")
+		out, err := nodesCmd.CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("failed to get nodes: %v", err)
+		}
+		nodes := strings.Fields(string(out))
+		for _, node := range nodes {
+			taintCmd := exec.Command("kubectl", "taint", "node", node, "node.kubernetes.io/os:NoSchedule-")
```
msau42, line 307: Is this required if we're not testing mixed os clusters?

test/run-windows-k8s-integration.sh
```diff
@@ -22,7 +21,7 @@ readonly teardown_driver=${GCE_CLUSTER_ZONE:-true}
 make -C ${PKGDIR} test-k8s-integration
 
 base_cmd="${PKGDIR}/bin/k8s-integration-test \
-            --platform=windows --local-k8s-dir=${LOCAL_K8S_DIR} --bringup-cluster=false --teardown-cluster=false --teardown-driver=${teardown_driver}\
+            --platform=windows --local-k8s-dir=${k8s_dir} --bringup-cluster=false --teardown-cluster=false --teardown-driver=${teardown_driver}\
```
msau42, line 24: local k8s dir is only for if you're running the test from your own local machine. Do we need it for CI? k8s-integration should download kubernetes and build in a tmp directory.

test/run-windows-k8s-integration.sh
```diff
@@ -9,9 +9,8 @@
 set -o nounset
 set -o errexit
 
-readonly GCE_PD_DO_DRIVER_BUILD=false
 readonly PKGDIR=${GOPATH}/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
-readonly LOCAL_K8S_DIR=${GOPATH}/src/k8s.io/kubernetes
+readonly k8s_dir=${PKGDIR}/kubernetes
```
verult, line 13: `PKGDIR` points to the PD CSI driver repo

test/run-windows-k8s-integration.sh
verult, line 24: `teardown_driver` references `GCE_CLUSTER_ZONE`. Is that intended?

test/run-windows-k8s-integration.sh
```diff
@@ -9,9 +9,8 @@
...
 readonly overlay_name="${GCE_PD_OVERLAY_NAME:-alpha}"
 readonly do_driver_build="${GCE_PD_DO_DRIVER_BUILD:-true}"
```
verult, line 15: `DO_DRIVER_BUILD` no longer exists

test/k8s-integration/driver.go
jingxu97: oh, I was thinking to keep "default" and add "gce-pd-csi-driver" namespace one. 
I am not sure why it was checking "default" before, or just a mistake.

Also this runCommand actually does not really check pod status. It will not return error if pods are not successfully running. In the second check I added, I print out the output for debugging.

msau42: When we changed namespaces, we forgot to update the test to check in the correct namespace. So we can remove the check for default namespace.

test/k8s-integration/main.go
jingxu97: this is only needed for windows clusters.

test/run-windows-k8s-integration.sh
jingxu97: right, in prow test, kuberentes will be installed under current dir which is the pd csi driver repo. in the test, kubetest needs to be in kubernetes dir.

test/run-windows-k8s-integration.sh
jingxu97: in prow test using kubetest, kuberentes will be installed under current dir which is the pd csi driver repo. in the test, kubetest needs to be running in kubernetes dir. We don't need to download kuberentes locally.

jingxu97:
comments were addressed. PTAL.

test/run-windows-k8s-integration.sh
msau42: kubetest should be run from `testDir`, not `k8sDir`: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go#L221

Maybe we need to set `--test-version` here.

Btw, does kubernetes_e2e support gke clusters?

jingxu97: I see there are some confusion here. For current pd driver ci tests for linux, we build kubernetes and install it under a tmp dir before running the test. But this step is not needed for the windows test setup. It uses script /workspace/scenarios/kubernetes_e2e.py which will install kuberentes like kube_up, I think. https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml#L65

This script will install kuberentes under current dir which is pd csi driver repo. The local-k8s-dir setting here does not mean the test will install kubernetes locally. I used this setting because the test use it change current dir here https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go#L421

So because I didn't set kubeversion or testversion, so the test code will not install kubernetes locally because it is not needed. https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/1c51cca720acaa91f2f00e55a5e44d5d5832ed0b/test/k8s-integration/main.go#L227

I plan to have another PR to edit the config parameter so that it is easier to read. Also I think for linux test, we can also use /workspace/scenarios/kubernetes_e2e.py to setup kubernetes instead of building it locally. We can set to use ci/latest or ci/master depending which kubernetes version we want to test.

test/k8s-integration/driver.go
jingxu97: got it. I plan to have follow up PR to check pod status and print out debug log etc.

test/run-windows-k8s-integration.sh
msau42: I think we should set "--test-version" instead. The reason is because we may use different Kubernetes versions to 1) bring up the cluster and 2) run e2e binary

We generally use "master" as the e2e version so that we can get new test cases and test fixes without relying on them to be backported to older Kubernetes versions.

"--local-k8s-dir" is intended for developer testing where you just want to run the e2es quickly without having to download and build kubernetes.

jingxu97: my understanding is since we already have prebuild kubernetes in different versions https://github.com/kubernetes/test-infra/blob/a4b7e151278a26fb45ee186b7fdc96394deed016/docs/kubernetes-versions.md
we could just use it by setting extract=ci/k8s-beta , ci/k8s-master , ... etc.
It will save k8s build time. Is there some specific version we want to test?

msau42: Yes, I think in release versions we can save a lot of time by using extract.  But if we want to test head, we still need to build. We do have https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go#L40 that I think can help with this.

But still, we have 3 distinct variables we use to distinguish different use cases:
- kubeVersion to set the cluster version
- testVersion to set the e2e version
- local for local development (you're making a change in kubernetes or the e2es that you want to test)

test/k8s-integration/driver.go
```diff
@@ -65,12 +67,12 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
...
-	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", "default")
-	err = runCommand("Checking driver pods", statusCmd)
+	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", "gce-pd-csi-driver")
+	out, err = statusCmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("failed to check driver pods: %v", err)
+		return fmt.Errorf("failed to get nodes: %v", err)
 	}
-
+	klog.V(2).Infof("GCE PD driver pods: \n %v", string(out))
```
msau42, line 75: BTW, this is fixed in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/140

test/k8s-integration/main.go
msau42: Should kubernetes-e2e.py be handling this as part of windows cluster bringup?

test/run-windows-k8s-integration.sh
msau42: I think we don't need this variable anymore?

test/k8s-integration/main.go
jingxu97: no, it is not.

msau42: Is there a reason why it shouldn't be part of standard cluster bringup tool? How come the end users need to do this manually?

test/k8s-integration/driver.go
jingxu97: ack.

test/run-windows-k8s-integration.sh
jingxu97: removed it. thanks!

test/k8s-integration/driver.go
msau42: I believe @mattcary is going to look into it as part of https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/139

test/run-windows-k8s-integration.sh
jingxu97: synced up offline. Will remove the need to locally download and build e2e binary in a separate PR. Instead we could set extract=ci/k8s-version.

jingxu97: fixed. Thanks1

test/run-windows-k8s-integration.sh
jingxu97: I don't see DO_DRIVER_BUILD? It has GCE_PD_DO_DRIVER_BUILD?

test/k8s-integration/main.go
jingxu97: I will check with windows team to see whether we could remove the taint in bringup tool.

jingxu97:
@verult @msau42 PTAL.

test/k8s-integration/driver.go
mattcary: I just wrote https://github.com/kubernetes/kubernetes/pull/93159, could we use that here as well?

I haven't figured out where the best places to wait for things are---if it has to be done here and in k8s/k8s

msau42: So for our pd driver repo tests, we need to add the logic in our k8s-integration test runner, which bring up the driver outside of the e2e tests.

jingxu97: thanks! I think we have to put the logic here to wait for pod running.

test/k8s-integration/main.go
jingxu97: because there are linux pods scheduled on windows cluster, so avoid pods scheduled to windows node, we currently taint windows node. We could add node selector to pods, but I think not all pods have it. So taint windows node is the easiest way for now.

msau42: I guess my question is why doesn't the cluster installer that brings up the windows cluster do it?

msau42: Oh sorry, I see this is removing the taint that got added. So in order to run e2e tests, we need to remove the taint so the e2e pods will run on it.  But is there a reason why kubernetes_e2e.py which is a e2e boostrapping script can't do it?

msau42: Anyway, let's think about if we should enhance kubernetes_e2e in the future so that all test runners don't have to do this. For now, can you add a comment explaining why we need to remove the taint?

test/k8s-integration/driver.go
```diff
@@ -71,9 +72,9 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
 	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", driverNamespace)
 	err = runCommand("Checking driver pods", statusCmd)
 	if err != nil {
-		return fmt.Errorf("failed to check driver pods: %v", err)
+		return fmt.Errorf("failed to get nodes: %v", err)
```
msau42, line 75: error message is not correct?

test/k8s-integration/driver.go
```diff
@@ -71,9 +72,9 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
...
 	}
-
+	klog.V(2).Infof("GCE PD driver pods: \n %v", string(out))
```
msau42, line 77: `out` is no longer relevant in this case

test/k8s-integration/driver.go
mattcary: @jingxu97 Do you want to look at kubernetes/kubernetes#93159 and make sure it has the right parameters? In particular the StatefulSet name is hardcoded, maybe I should move that to a parameter to make it usable from the pd driver repo? (I'd need to look again in the pd driver repo to see if that's necessary).

msau42: So this repo switched to use a Deployment instead of a StatefulSet, so importing it is not going to work as is.

Maybe we can switch k/k to use Deployment too?

jingxu97: @mattcary so far your pr looks good to me. In case some pod not running correctly, do we collect logs on events which help debug?

jingxu97: never mind, I think logs are collected https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/140

test/k8s-integration/driver.go
jingxu97: fixed.

test/k8s-integration/driver.go
jingxu97: fixed. PTAL

test/k8s-integration/driver.go
```diff
@@ -71,9 +71,8 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
 	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", driverNamespace)
 	err = runCommand("Checking driver pods", statusCmd)
 	if err != nil {
-		return fmt.Errorf("failed to check driver pods: %v", err)
+		return fmt.Errorf("failed to describe nodes: %v", err)
```
msau42, line 74: should be describe pods?

jingxu97: oh, not sure what I was thinking. fixed.

test/k8s-integration/driver.go
mattcary: Done, switched to deployment and moved the driver name to the exported function.

mattcary: Actually there already is a deployment-waiter in  framework/deployment/wait that could be used. That's what my change is using now, so there's nothing to export to you.

PR #554, opened 2020-07-15
Test logs

### LINKED ISSUE #140: Collect driver logs for k8s test

msau42:
**What type of PR is this?**
/kind cleanup

Dump driver logs to test artifacts to make it easier to debug test failures

Fixes #140

The first commit that is updating vendor has a separate PR https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/553

```release-note
NONE
```


test/k8s-integration/main.go
```diff
@@ -311,6 +314,27 @@ func handle() error {
...
+				StatusWriter:  os.Stdout,
+				LogPathPrefix: filepath.Join(artifactsDir, "pd-csi-driver") + "/",
+			}
+			ctx, cancel := context.WithCancel(context.Background())
+			defer func() {
+				cancel()
+			}()
+			if err = podlogs.CopyAllLogs(ctx, client, driverNamespace, out); err != nil {
+				return fmt.Errorf("failed to start pod logger: %v", err)
+			}
```
saikat-royc, line 335: 1. Can we move this function to enable logging to a helper function?
2. For useGKEManagedDriver = true, the driver name space is not 'gce-pd-csi-driver'. Should we enable logging only if useGKEManagedDriver is false?

msau42: 1. Done
2. Yes, this is only called in the !useGKEManagedDriver block

Dockerfile
```diff
@@ -14,11 +14,10 @@
 
 
 
-FROM golang:1.13.4-alpine3.10 as builder
+FROM golang:1.13.4 as builder
```
saikat-royc, line 17: any specific reason for not using 1.14?

msau42: nope I just wasn't sure if we were ready to change it yet. Sometimes the new kubekins images don't have the newer golang versions built in

saikat-royc: Thanks for the clarification

PR #559, opened 2020-07-21
update driver test code for Windows

jingxu97:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/test/k8s-integration/driver-config.go
+++ b/test/k8s-integration/driver-config.go
@@ -45,6 +45,7 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 
 	// Fill in template parameters. Capabilities can be found here:
 	// https://github.com/kubernetes/kubernetes/blob/b717be8269a4f381ab6c23e711e8924bc1f64c93/test/e2e/storage/testsuites/testdriver.go#L136
+
 	caps := []string{
 		"persistence",
 		"block",
@@ -56,6 +57,12 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 	var fsTypes []string
 	if platform == "windows" {
 		fsTypes = []string{"ntfs"}
+		caps = []string{
+			"persistence",
+			"exec",
+			"multipods",
+			"topology",
+		}
 	} else {
 		fsTypes = []string{
 			"ext2",
--- a/test/k8s-integration/driver.go
+++ b/test/k8s-integration/driver.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"k8s.io/klog"
 	"k8s.io/kubernetes/test/e2e/framework/podlogs"
 )
 
@@ -15,9 +16,10 @@ func getOverlayDir(pkgDir, deployOverlayName string) string {
 	return filepath.Join(pkgDir, "deploy", "kubernetes", "overlays", deployOverlayName)
 }
 
-func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayName string, doDriverBuild bool) error {
+func installDriver(platform, goPath, pkgDir, stagingImage, stagingVersion, deployOverlayName string, doDriverBuild bool) error {
 	if doDriverBuild {
 		// Install kustomize
+		klog.Infof("Installing kustomize")
 		out, err := exec.Command(filepath.Join(pkgDir, "deploy", "kubernetes", "install-kustomize.sh")).CombinedOutput()
 		if err != nil {
 			return fmt.Errorf("failed to install kustomize: %s, err: %v", out, err)
@@ -40,6 +42,17 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
 			fmt.Sprintf("%s=%s:%s", pdImagePlaceholder, stagingImage, stagingVersion)).CombinedOutput()
 		if err != nil {
 			return fmt.Errorf("failed to edit kustomize: %s, err: %v", out, err)
+		}
+		if platform == "windows" {
+			out, err = exec.Command(
+				filepath.Join(pkgDir, "bin", "kustomize"),
+				"edit",
+				"set",
+				"image",
+				fmt.Sprintf("%s-win=%s-win:%s", pdImagePlaceholder, stagingImage, stagingVersion)).CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("failed to edit kustomize: %s, err: %v", out, err)
+			}
 		}
 	}
 
@@ -65,11 +78,18 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
 	if err != nil {
 		return fmt.Errorf("failed to deploy driver: %v", err)
 	}
-
+	klog.Infof("Deploying driver")
 	// TODO (#139): wait for driver to be running
-	time.Sleep(time.Minute)
-	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", driverNamespace)
-	err = runCommand("Checking driver pods", statusCmd)
+	if platform == "windows" {
+		klog.Infof("Waiting 15 minutes for the driver to start on Windows")
+		time.Sleep(15 * time.Minute)
+	} else {
+		klog.Infof("Waiting 5 minutes for the driver to start on Linux")
+		time.Sleep(5 * time.Minute)
+	}
+	out, err = exec.Command("kubectl", "describe", "pods", "-n", driverNamespace).CombinedOutput()
+	klog.Infof("describe pods \n %s", string(out))
+
 	if err != nil {
 		return fmt.Errorf("failed to describe pods: %v", err)
 	}
@@ -99,18 +119,22 @@ func pushImage(pkgDir, stagingImage, stagingVersion, platform string) error {
 		return err
 	}
 	var cmd *exec.Cmd
-	if platform != "windows" {
-		cmd = exec.Command("make", "-C", pkgDir, "push-container",
-			fmt.Sprintf("GCE_PD_CSI_STAGING_VERSION=%s", stagingVersion),
-			fmt.Sprintf("GCE_PD_CSI_STAGING_IMAGE=%s", stagingImage))
-	} else {
-		cmd = exec.Command("make", "-C", pkgDir, "build-and-push-windows-container-1909",
-			fmt.Sprintf("GCE_PD_CSI_STAGING_VERSION=%s", stagingVersion),
-			fmt.Sprintf("GCE_PD_CSI_STAGING_IMAGE=%s", stagingImage))
-	}
-	err = runCommand("Pushing GCP Container", cmd)
+
+	cmd = exec.Command("make", "-C", pkgDir, "push-container",
+		fmt.Sprintf("GCE_PD_CSI_STAGING_VERSION=%s", stagingVersion),
+		fmt.Sprintf("GCE_PD_CSI_STAGING_IMAGE=%s", stagingImage))
+	err = runCommand("Pushing GCP Container for Linux", cmd)
 	if err != nil {
-		return fmt.Errorf("failed to run make command: err: %v", err)
+		return fmt.Errorf("failed to run make command for linux: err: %v", err)
+	}
+	if platform == "windows" {
+		cmd = exec.Command("make", "-C", pkgDir, "build-and-push-windows-container-ltsc2019",
+			fmt.Sprintf("GCE_PD_CSI_STAGING_VERSION=%s", stagingVersion),
+			fmt.Sprintf("GCE_PD_CSI_STAGING_IMAGE=%s-win", stagingImage))
+		err = runCommand("Building and Pushing GCP Container for Windows", cmd)
+		if err != nil {
+			return fmt.Errorf("failed to run make command for windows: err: %v", err)
+		}
 	}
 	return nil
 }
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -79,6 +79,8 @@ func init() {
 }
 
 func main() {
+	klog.InitFlags(nil)
+	flag.Set("logtostderr", "true")
 	flag.Parse()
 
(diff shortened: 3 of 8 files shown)


jingxu97:
pull-gcp-compute-persistent-disk-csi-driver-e2e-windows

jingxu97:
pull-gcp-compute-persistent-disk-csi-driver-e2e-windows

jingxu97:
pull-gcp-compute-persistent-disk-csi-driver-e2e-windows

jingxu97:
@msau42 PTAL

test/k8s-integration/driver-config.go
```diff
@@ -54,6 +55,12 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 	var fsTypes []string
 	if platform == "windows" {
 		fsTypes = []string{"ntfs"}
+		caps = []string{
+			"persistence",
+			"exec",
+			"multipods",
+			"topolify",
```
msau42, line 62: topology

test/k8s-integration/driver.go
```diff
@@ -65,11 +77,18 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
 	if err != nil {
 		return fmt.Errorf("failed to deploy driver: %v", err)
 	}
-
+	klog.Infof("Deploying driver")
+	fmt.Printf("Deploying driver")
```
msau42, line 81: Do we need both klog and fmt?

test/k8s-integration/driver.go
```diff
@@ -65,11 +77,18 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
...
 	// TODO (#139): wait for driver to be running
-	time.Sleep(time.Minute)
-	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", driverNamespace)
-	err = runCommand("Checking driver pods", statusCmd)
+	if platform == "windows" {
+		klog.Infof("Test waiting for 15 mins")
```
msau42, line 84: Can this be move descriptive, like "Waiting for 15 minutes to start the driver"?

Also it really takes 15 minutes?

test/k8s-integration/driver.go
```diff
@@ -65,11 +77,18 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
...
+		time.Sleep(15 * time.Minute)
+	} else {
+		time.Sleep(1 * time.Minute)
+	}
+	out, err = exec.Command("kubectl", "describe", "pods", "-n", driverNamespace).CombinedOutput()
+	klog.V(2).Infof("describe pods \n %s", string(out))
+	fmt.Printf("describe pods \n %s", string(out))
```
msau42, line 91: Same thing both klog and fmt?

test/k8s-integration/driver.go
```diff
@@ -99,14 +118,18 @@ func pushImage(pkgDir, stagingImage, stagingVersion, platform string) error {
...
+
+	cmd = exec.Command("make", "-C", pkgDir, "push-container",
+		fmt.Sprintf("GCE_PD_CSI_STAGING_VERSION=%s", stagingVersion),
+		fmt.Sprintf("GCE_PD_CSI_STAGING_IMAGE=%s", stagingImage))
+	err = runCommand("Pushing GCP Container", cmd)
+	if err != nil {
+		return fmt.Errorf("failed to run make command: err: %v", err)
+	}
+	if platform == "windows" {
+		cmd = exec.Command("make", "-C", pkgDir, "build-and-push-windows-container-ltsc2019",
```
msau42, line 130: Can you explain why we need to do push-container first?

test/k8s-integration/driver.go
jingxu97: from build-log.txt, I cannot see this log with klog only. It works with fmt.Printf. Is it normal?

test/k8s-integration/driver.go
jingxu97: I have seen it took about 12 mins before. Normally downloading images take longer time than linux ones. I will try to get more data about this.

test/k8s-integration/driver.go
jingxu97: klog result does not show in build-log.txt

msau42: I think you need to do something like [this](https://github.com/kubernetes-csi/external-provisioner/blob/master/cmd/csi-provisioner/csi-provisioner.go#L111) to get the klog to print to stdout

test/k8s-integration/driver-config.go
jingxu97: fixed.

test/k8s-integration/driver.go
jingxu97: this command build and push container image together for windows, I update the runCommand message to make it clear.

jingxu97:
pull-gcp-compute-persistent-disk-csi-driver-e2e-windows

test/k8s-integration/driver.go
msau42: Ok I understand now. We still need to build and push LInux images because the controller part still needs to run on a Linux machine.

test/k8s-integration/main.go
```diff
@@ -80,6 +80,7 @@ func init() {
 
 func main() {
 	flag.Parse()
+	klog.InitFlags(nil)
```
msau42, line 83: I forget if the `klog.InitFlags` has to come before `flag.Parse`

msau42: We may also have to set logtostderr = true

test/k8s-integration/driver.go
jingxu97: yes, linux image build and push will be always triggered.

jingxu97:
@msau42 PTAL. The presubmit job could not completely pass because it depends on other test fixes (e.g., timeout). But for this change, I do manual tests and think it is ok to merge. We can keep monitoring tests.

test/k8s-integration/driver.go
```diff
@@ -8,16 +8,18 @@ import (
...
 
 func getOverlayDir(pkgDir, deployOverlayName string) string {
 	return filepath.Join(pkgDir, "deploy", "kubernetes", "overlays", deployOverlayName)
 }
 
-func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayName string, doDriverBuild bool) error {
+func installDriver(platform, goPath, pkgDir, stagingImage, stagingVersion, deployOverlayName string, doDriverBuild bool) error {
 	if doDriverBuild {
 		// Install kustomize
+		klog.Infof("kustomize")
```
msau42, line 22: Can you make this more descriptive, like "Installing kustomize"

test/k8s-integration/driver.go
```diff
@@ -65,11 +78,19 @@ func installDriver(goPath, pkgDir, stagingImage, stagingVersion, deployOverlayNa
...
 	}
-
+	klog.Infof("Deploying driver")
 	// TODO (#139): wait for driver to be running
-	time.Sleep(time.Minute)
-	statusCmd := exec.Command("kubectl", "describe", "pods", "-n", driverNamespace)
-	err = runCommand("Checking driver pods", statusCmd)
+	for i := 0; i < 3; i++ {
+		if platform == "windows" {
+			klog.Infof("Test waiting for 5 mins")
```
msau42, line 85: Can you make this more descriptive, like "Waiting 5 minutes for the driver to start". Should we log for linux too?

test/k8s-integration/main.go
```diff
@@ -298,27 +300,52 @@ func handle() error {
...
 		for _, node := range nodes {
 			taintCmd := exec.Command("kubectl", "taint", "node", node, "node.kubernetes.io/os:NoSchedule-")
-			_, err = taintCmd.CombinedOutput()
+			out, err := taintCmd.CombinedOutput()
 			if err != nil {
 				return fmt.Errorf("failed to untaint windows node %v", err)
 			}
+			klog.Infof("untaint windows nodes: %s, output %s", node, string(out))
+		}
+		/*nodesCmd = exec.Command("kubectl", "get", "nodes", "-l", "beta.kubernetes.io/os=linux", "-o", "name")
```
msau42, line 321: can we remove this code?

test/k8s-integration/main.go
```diff
@@ -342,6 +369,25 @@ func handle() error {
 		}()
 	}
 
+	// For windows cluster, when cluster is up, all Windows nodes are tainted with NoSchedule to avoid linux pods
+	// being scheduled to Windows nodes. When running windows tests, we need to remove the taint.
```
msau42, line 373: The comment doesn't match the code

test/k8s-integration/main.go
```diff
@@ -487,7 +533,7 @@ func runTestsWithConfig(testDir, testFocus, testSkip, testConfigArg string, clou
 	}
 	ginkgoArgs := fmt.Sprintf("--ginkgo.focus=%s --ginkgo.skip=%s", testFocus, testSkip)
 	if *platform == "windows" {
-		ginkgoArgs = ginkgoArgs + fmt.Sprintf(" --node-os-distro=%s", *platform)
+		ginkgoArgs = ginkgoArgs + fmt.Sprintf(" --node-os-distro=%s --allowed-not-ready-nodes=3", *platform)
```
msau42, line 536: Can you add a comment explaining why there are not ready nodes?

test/k8s-integration/prepull-image.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+if [[ -z "${PREPULL_IMAGE}" ]]; then
+    # Pre-pull all the test images. The images are currently hard-coded.
```
msau42, line 8: Can you explain in a comment what issue prepulling images helps solve?

test/k8s-integration/main.go
jingxu97: oh, sorry. I forgot to push my latest update there.

test/k8s-integration/driver.go
jingxu97: updated

test/k8s-integration/prepull-image.sh
jingxu97: added. PTAL

test/k8s-integration/main.go
jingxu97: added

test/k8s-integration/driver.go
jingxu97: updated

test/k8s-integration/main.go
```diff
@@ -342,6 +358,30 @@ func handle() error {
 		}()
 	}
 
+	// For windows cluster, it has both Windows nodes and Linux nodes. Before triggering the tests, taint Linux nodes
+	// with NoSchedule to avoid test pods being scheduled on Linux. Need to do this step after driver is deployed.
```
msau42, line 362: We need to do this after so that the driver part that runs on Linux won't be evicted? Can we add a toleration to the driver instead?

test/run-windows-k8s-integration.sh
```diff
@@ -23,7 +23,7 @@ base_cmd="${PKGDIR}/bin/k8s-integration-test \
             --platform=windows --bringup-cluster=false --teardown-cluster=false --teardown-driver=${teardown_driver}\
             --run-in-prow=true --deploy-overlay-name=${overlay_name} --service-account-file=${E2E_GOOGLE_APPLICATION_CREDENTIALS} \
             --do-driver-build=${do_driver_build} --gce-zone=${gce_zone} --test-version=${test_version}\
-            --storageclass-files=sc-windows.yaml --snapshotclass-file=pd-volumesnapshotclass.yaml --test-focus='External.Storage' \
+            --storageclass-files=sc-windows.yaml --snapshotclass-file=pd-volumesnapshotclass.yaml --test-focus='External.Storage.*ntfs.*[sig-windows]' \
```
msau42, line 26: We need [sig-windows] because some test cases in the external storage framework are known to not work on Windows? Would it be better to have a skip on [LinuxOnly] instead?

test/k8s-integration/driver.go
jingxu97: updated

test/k8s-integration/main.go
jingxu97: actually we do have tolerations https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/base/node_linux/node.yaml#L103
but there is some issue when I try to taint linux node before deploying it. Let me double check.

test/run-windows-k8s-integration.sh
jingxu97: I think it is ok to remove sig-windows. updated

test/k8s-integration/main.go
jingxu97: I found it, the controller does not have the tolerations. I think we want to avoid setting tolerations to controller to make sure it can be scheduled to a "good" node.

msau42: In general, the driver should be more resilient to ordering, but I guess this is a test-only scenario where by default we don't want pods to be scheduled to linux nodes. 

In a normal scenario, is setting taints the recommended way? Or do users need to explicitly set nodeSelectors for the OS node they want?

jingxu97: I think nodeSelector is better for its flexibility. Taint windows nodes when starting a cluster to avoid some system pods scheduled to Windows ( I guess those Pods do not have node selector). After cluster is up, user should remove windows node taint, then schedule pods based on node selector.

test/k8s-integration/main.go
jingxu97: updated.

jingxu97:
@msau42 I addressed the comments. PTAL

test/k8s-integration/main.go
jingxu97: added

PR #565, opened 2020-07-27
Test support for Balanced PDs

mattcary:
/kind feature

Add testing for the new balanced PD feature
```release-note
NONE
```

### DIFF
--- a/pkg/common/parameters.go
+++ b/pkg/common/parameters.go
@@ -42,7 +42,7 @@ const (
 
 // DiskParameters contains normalized and defaulted disk parameters
 type DiskParameters struct {
-	// Values: pd-standard OR pd-ssd
+	// Values: pd-standard, pd-balanced, pd-ssd, or any other PD disk type. Not validated.
 	// Default: pd-standard
 	DiskType string
 	// Values: "none", regional-pd
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -51,7 +51,7 @@ var (
 
 	// Test infrastructure flags
 	boskosResourceType = flag.String("boskos-resource-type", "gce-project", "name of the boskos resource type to reserve")
-	storageClassFile   = flag.String("storageclass-file", "", "name of storageclass yaml file to use for test relative to test/k8s-integration/config")
+	storageClassFiles  = flag.String("storageclass-files", "", "name of storageclass yaml file to use for test relative to test/k8s-integration/config. This may be a comma-separated list to test multiple storage classes")
 	snapshotClassFile  = flag.String("snapshotclass-file", "", "name of snapshotclass yaml file to use for test relative to test/k8s-integration/config")
 	inProw             = flag.Bool("run-in-prow", false, "is the test running in PROW")
 
@@ -107,9 +107,9 @@ func main() {
 	}
 
 	if *migrationTest {
-		ensureVariable(storageClassFile, false, "storage-class-file and migration-test cannot both be set")
+		ensureVariable(storageClassFiles, false, "storage-class-file and migration-test cannot both be set")
 	} else {
-		ensureVariable(storageClassFile, true, "One of storageclass-file and migration-test must be set")
+		ensureVariable(storageClassFiles, true, "One of storageclass-file and migration-test must be set")
 	}
 
 	if !*bringupCluster {
@@ -367,8 +367,16 @@ func handle() error {
 	}
 
 	// Run the tests using the testDir kubernetes
-	if len(*storageClassFile) != 0 {
-		err = runCSITests(*platform, pkgDir, testDir, *testFocus, testSkip, *storageClassFile, *snapshotClassFile, cloudProviderArgs, *deploymentStrat)
+	if len(*storageClassFiles) != 0 {
+		storageClasses := strings.Split(*storageClassFiles, ",")
+		for _, scFile := range storageClasses {
+			if err = runCSITests(*platform, pkgDir, testDir, *testFocus, testSkip, scFile, *snapshotClassFile, cloudProviderArgs, *deploymentStrat, scFile); err != nil {
+				return fmt.Errorf("runCSITests failed: %w", err)
+			}
+		}
+		if err = mergeArtifacts(storageClasses); err != nil {
+			return fmt.Errorf("artifact merging failed: %w", err)
+		}
 	} else if *migrationTest {
 		err = runMigrationTests(pkgDir, testDir, *testFocus, testSkip, cloudProviderArgs)
 	} else {
@@ -376,7 +384,7 @@ func handle() error {
 	}
 
 	if err != nil {
-		return fmt.Errorf("failed to run tests: %v", err)
+		return fmt.Errorf("failed to run tests: %w", err)
 	}
 
 	return nil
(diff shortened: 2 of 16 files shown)


mattcary:
/assign msau42
/assign saikat-royc

Hi folks. This PR does testing for balanced PD. Is it excessive? It lets us pass a list of storage classes to the k8s-integration test and runs all tests through all storage classes.

Also, I split in a second commit cleaning up constants and such for PD types that we don't use. Either we should do that, or we should actually validate the type---but doing that seems unnecessarily fragile and limiting to me.

test/k8s-integration/main.go
```diff
@@ -51,7 +51,7 @@ var (
 
 	// Test infrastructure flags
 	boskosResourceType = flag.String("boskos-resource-type", "gce-project", "name of the boskos resource type to reserve")
-	storageClassFile   = flag.String("storageclass-file", "", "name of storageclass yaml file to use for test relative to test/k8s-integration/config")
+	storageClassFile   = flag.String("storageclass-file", "", "name of storageclass yaml file to use for test relative to test/k8s-integration/config. This may be a comma-separated list to test multiple storage classes")
```
msau42, line 54: should we rename it "storageclass-files"?

test/k8s-integration/main.go
```diff
@@ -368,7 +368,10 @@ func handle() error {
 
 	// Run the tests using the testDir kubernetes
 	if len(*storageClassFile) != 0 {
-		err = runCSITests(*platform, pkgDir, testDir, *testFocus, testSkip, *storageClassFile, *snapshotClassFile, cloudProviderArgs, *deploymentStrat)
+		storageClasses := strings.Split(*storageClassFile, ",")
+		for _, scFile := range storageClasses {
+			err = runCSITests(*platform, pkgDir, testDir, *testFocus, testSkip, scFile, *snapshotClassFile, cloudProviderArgs, *deploymentStrat)
```
msau42, line 373: How do the test artifacts look? Do all the test results get aggregated, overwritten, or do they look like 2 different runs?

test/k8s-integration/main.go
mattcary: I didn't know if that would break any existing jobs. Is everything on prow etc run through the run-k8s-integration*.sh scripts?

test/k8s-integration/main.go
mattcary: It looks like two different runs in terms of stdout. I'm not sure about the test artifacts, I'll have to look to see how they appear locally.

test/k8s-integration/main.go
msau42: it goes through the script, so if we modify the script we should be fine

mattcary: Cool thanks done.

test/k8s-integration/main.go
mattcary: The control plane logs are good, it's everything since the cluster was brought up AFAICT.

Not sure about the junit*.xml logs. Continuing to dig.

saikat-royc: We can setup the ARTIFACTS env variable and run one of the jobs in run-k8s-integration-local.sh and check if the logs are overwritten for the subsequent run of kubetest.
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go#L461

msau42: pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration should be running these changes

mattcary: Yeah, that's what I'm doing. It looks like the junit stuff is overwritten but I have a fix being tested (but does prow/testgrid care where the junit*.xml files end up?)

mattcary:
The commit I just pushed puts the output into two separate directories in ARTIFACTS.

I suspect this is going to not make prow happy?

I'm thinking that it will be easier to just keep the single storageclass and configure another prow job manually. Or is that too conservative?

msau42:
In our csi sidecar tests, we do invoke e2e multiple times, and we do some post processing to combine the junit files into one: https://github.com/kubernetes-csi/csi-release-tools/blob/0345a835e2177c3135bbba8f2793b012f0bbd2ea/prow.sh#L1187

msau42:
but in this case, I wonder if we will run into issues because we'll have test case names that are the same (since it's driver name in the test case name, not storageclass)

mattcary:
> but in this case, I wonder if we will run into issues because we'll have test case names that are the same (since it's driver name in the test case name, not storageclass)

Yeah, I started to look into adding the storageclass into the ginkgo test description string, but that was starting to get complicated as well.

saikat-royc:
Would it be simpler to create new jobs where we point to a balanced pd storage class in the job config, instead of running the tests multiple times for the list of storage classes?

mattcary:
> Would it be simpler to create new jobs where we point to a balanced pd storage class in the job config, instead of running the tests multiple times for the list of storage classes?

That's what I'm thinking, yeah. That or figure out how to combine the storageclass with the driver so that from ginkgo's perspective the driver would be something like "pdcsi-standard" or "pdcsi-balanced".

mattcary:
Ok, lots of rearrangements. The first commit is a general one, adding support for multiple storage classes and merging all results.

The second one is the PD-balanced-specific one.

I can split these into two PRs if you like.

Is there any way to test this with prow to make sure that my junit fiddling is ok? At the end the ARTIFACTS directory looks like:
junit_pdcsi.xml  (merged xml)
sc-standard.yaml/
  junit_*.xml (the original xml, that is now repeated in ../junit_pdcsi.xml)
sc-another-one.yaml/
  etc

msau42:
pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration  should be running these changes

mattcary:
The following prow job shows how the tests show up as two different csi drivers, one with the standard and the other with balanced PDs.

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/565/pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration/1289304359020007427/

msau42:
Excellent, I think the other thing we need to potentially do is bump up the timeout in our test jobs since we're doubling the number of tests.

Would be nice to add pd-ssd in there too for completeness, but could be done separately.

msau42:
I guess the other consideration is do we want to enable all the different pd configs in the pull job, or just in the ci job? Adding it all in the pull job means that prs will take longer to finish testing.

mattcary:
Adding PD-SSD is a good idea though I wanted to see if this made anything blow up first (like the timeout issue which I had forgotten).

> I guess the other consideration is do we want to enable all the different pd configs in the pull job, or just in the ci job? Adding it all in the pull job means that prs will take longer to finish testing.

I think just the CI job. It's all supposed to be the same API so I feel like this is more a double-check that GCE isn't changing anything under us.

mattcary:
I've updated this so that the pull test only tests pd-standard, and added a CI script that tests all the classes.

I think https://github.com/kubernetes/test-infra/pull/18847 is the right PR to update the CI jobs?

pkg/common/parameters.go
```diff
@@ -42,7 +42,7 @@ const (
 
 // DiskParameters contains normalized and defaulted disk parameters
 type DiskParameters struct {
-	// Values: pd-standard OR pd-ssd
+	// Values: PD disk type (see GCP documentation. This driver does not validate this field).
```
saikat-royc, line 45: nit: keeping it consistent with comments on other fields, we can put the actual values. ("pd-standard","pd-ssd", "pd-balanced")

mattcary: Done.

saikat-royc:
we also need to update https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/run-windows-k8s-integration.sh#L26, storageclass-file to storageclass-files

mattcary:
> we also need to update https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/run-windows-k8s-integration.sh#L26, storageclass-file to storageclass-files

good catch, done

saikat-royc:
thanks for making this complicated change!

PR #566, opened 2020-07-29
Add gce disk labels support via storageclass/CreateVolumeRequest parameters

### LINKED ISSUE #340: Support for disk labels parameter
# Problem

I use PVC and PV resources on GKE, and I also use PersistentDisk(PD) for ComputeEngine instances.
I usually filter PDs by labels when I delete it.
However, the PD which created by PV does not have labels for filtering.
This makes me difficult to filter to delete PDs.

# Solution

Adding parameter to StorageClass and use it in this driver.

```
https://kubernetes.io/docs/concepts/storage/storage-classes/#gce-pd
e.g.

apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: slow
provisioner: kubernetes.io/gce-pd
parameters:
  type: pd-standard
  replication-type: none
  disk-labels: 'foo=123,bar=456' # Add this parameter to set labels
```

# Additional Info

I read the related issue on [external-provisioner](https://github.com/kubernetes-csi/external-provisioner/issues/86), but in my use case, I think it's better to implement on this driver, since it depends on GCP PD.

Sample implementation:
https://github.com/katsew/gcp-compute-persistent-disk-csi-driver/tree/disk-labels-support/pkg

ernestas2k:
**What type of PR is this?**
/kind feature

Adds support to assign gce disk labels via storageclass / CreateVolumeRequest parameters.
 
Fixes #340

- Any test I shouldn't have altered or should additionally update?
- note: API contract has changed - support of new "labels" parameter added.

```release-note
Added support to assign gce disk labels by providing storageclass parameter `labels`.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ See Github [Issues](https://github.com/kubernetes-sigs/gcp-compute-persistent-di
 | type             | `pd-ssd` OR `pd-standard` | `pd-standard` | Type allows you to choose between standard Persistent Disks  or Solid State Drive Persistent Disks |
 | replication-type | `none` OR `regional-pd`   | `none`        | Replication type allows you to choose between Zonal Persistent Disks or Regional Persistent Disks  |
 | disk-encryption-kms-key | Fully qualified resource identifier for the key to use to encrypt new disks. | Empty string. | Encrypt disk using Customer Managed Encryption Key (CMEK). See [GKE Docs](https://cloud.google.com/kubernetes-engine/docs/how-to/using-cmek#create_a_cmek_protected_attached_disk) for details. |
+| labels           | `key1=value1,key2=value2` |               | Labels allow you to assign custom GCE Disk labels                                                  |
 
 ### Topology
 
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -18,6 +18,7 @@ package common
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
@@ -147,4 +148,61 @@ func CreateNodeID(project, zone, name string) string {
 
 func CreateZonalVolumeID(project, zone, name string) string {
 	return fmt.Sprintf(volIDZonalFmt, project, zone, name)
+}
+
+// ConvertLabelsStringToMap converts the labels from string to map
+// example: "key1=value1,key2=value2" gets converted into {"key1": "value1", "key2": "value2"}
+func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
+	const labelsDelimiter = ","
+	const labelsKeyValueDelimiter = "="
+
+	labelsMap := make(map[string]string)
+	if labels == "" {
+		return labelsMap, nil
+	}
+
+	regexKey, _ := regexp.Compile(`^\p{Ll}[\p{Ll}0-9_-]{0,62}$`)
+	checkLabelKeyFn := func(key string) error {
+		if !regexKey.MatchString(key) {
+			return fmt.Errorf("label value %q is invalid (should start with lowercase letter / lowercase letter, digit, _ and - chars are allowed / 1-63 characters", key)
+		}
+		return nil
+	}
+
+	regexValue, _ := regexp.Compile(`^[\p{Ll}0-9_-]{0,63}$`)
+	checkLabelValueFn := func(value string) error {
+		if !regexValue.MatchString(value) {
+			return fmt.Errorf("label value %q is invalid (lowercase letter, digit, _ and - chars are allowed / 0-63 characters", value)
+		}
+
+		return nil
+	}
+
+	keyValueStrings := strings.Split(labels, labelsDelimiter)
+	for _, keyValue := range keyValueStrings {
+		keyValue := strings.Split(keyValue, labelsKeyValueDelimiter)
+
+		if len(keyValue) != 2 {
+			return nil, fmt.Errorf("labels %q are invalid, correct format: 'key1=value1,key2=value2'", labels)
+		}
+
+		key := strings.TrimSpace(keyValue[0])
+		if err := checkLabelKeyFn(key); err != nil {
+			return nil, err
+		}
+
+		value := strings.TrimSpace(keyValue[1])
+		if err := checkLabelValueFn(value); err != nil {
+			return nil, err
+		}
+
+		labelsMap[key] = value
+	}
+
+	const maxNumberOfLabels = 64
+	if len(labelsMap) > maxNumberOfLabels {
+		return nil, fmt.Errorf("more than %d labels is not allowed, given: %d", maxNumberOfLabels, len(labelsMap))
+	}
+
+	return labelsMap, nil
 }
--- a/pkg/common/utils_test.go
+++ b/pkg/common/utils_test.go
@@ -295,3 +295,185 @@ func TestKeyToVolumeID(t *testing.T) {
 	}
 
 }
+
+func TestConvertLabelsStringToMap(t *testing.T) {
+	t.Run("parsing labels string into map", func(t *testing.T) {
+		testCases := []struct {
+			name           string
+			labels         string
+			expectedOutput map[string]string
+			expectedError  bool
+		}{
+			{
+				name:           "should return empty map when labels string is empty",
+				labels:         "",
+				expectedOutput: map[string]string{},
+				expectedError:  false,
+			},
+			{
+				name:   "single label string",
+				labels: "key=value",
+				expectedOutput: map[string]string{
+					"key": "value",
+				},
+				expectedError: false,
+			},
+			{
+				name:   "multiple label string",
+				labels: "key1=value1,key2=value2",
+				expectedOutput: map[string]string{
+					"key1": "value1",
+					"key2": "value2",
+				},
+				expectedError: false,
+			},
+			{
+				name:   "multiple labels string with whitespaces gets trimmed",
+				labels: "key1=value1, key2=value2",
+				expectedOutput: map[string]string{
+					"key1": "value1",
+					"key2": "value2",
+				},
+				expectedError: false,
+			},
+			{
+				name:           "malformed labels string (no keys and values)",
+				labels:         ",,",
+				expectedOutput: nil,
+				expectedError:  true,
+			},
+			{
+				name:           "malformed labels string (incorrect format)",
+				labels:         "foo,bar",
+				expectedOutput: nil,
+				expectedError:  true,
+			},
+			{
+				name:           "malformed labels string (missing key)",
+				labels:         "key1=value1,=bar",
+				expectedOutput: nil,
+				expectedError:  true,
+			},
+			{
+				name:           "malformed labels string (missing key and value)",
+				labels:         "key1=value1,=bar,=",
+				expectedOutput: nil,
+				expectedError:  true,
+			},
+		}
+
+		for _, tc := range testCases {
+			t.Logf("test case: %s", tc.name)
+			output, err := ConvertLabelsStringToMap(tc.labels)
+			if tc.expectedError && err == nil {
+				t.Errorf("Expected error but got none")
+			}
+			if err != nil {
+				if !tc.expectedError {
+					t.Errorf("Did not expect error but got: %v", err)
+				}
+				continue
+			}
+
+			if !reflect.DeepEqual(output, tc.expectedOutput) {
+				t.Errorf("Got labels %v, but expected %v", output, tc.expectedOutput)
+			}
+		}
+	})
+
+	t.Run("checking google requirements", func(t *testing.T) {
+		testCases := []struct {
+			name          string
+			labels        string
+			expectedError bool
+		}{
+			{
+				name: "64 labels at most",
+				labels: `k1=v,k2=v,k3=v,k4=v,k5=v,k6=v,k7=v,k8=v,k9=v,k10=v,k11=v,k12=v,k13=v,k14=v,k15=v,k16=v,k17=v,k18=v,k19=v,k20=v,
+                         k21=v,k22=v,k23=v,k24=v,k25=v,k26=v,k27=v,k28=v,k29=v,k30=v,k31=v,k32=v,k33=v,k34=v,k35=v,k36=v,k37=v,k38=v,k39=v,k40=v,
+                         k41=v,k42=v,k43=v,k44=v,k45=v,k46=v,k47=v,k48=v,k49=v,k50=v,k51=v,k52=v,k53=v,k54=v,k55=v,k56=v,k57=v,k58=v,k59=v,k60=v,
+                         k61=v,k62=v,k63=v,k64=v,k65=v`,
+				expectedError: true,
+			},
+			{
+				name:          "label key must start with lowercase char (# case)",
+				labels:        "#k=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key must start with lowercase char (_ case)",
+				labels:        "_k=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key must start with lowercase char (- case)",
+				labels:        "-k=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key can only contain lowercase chars, digits, _ and -)",
+				labels:        "k*=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key may not have over 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij1234=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key cannot contain . and /",
+				labels:        "kubernetes.io/created-for/pvc/namespace=v",
+				expectedError: true,
+			},
+			{
+				name:          "label value can only contain lowercase chars, digits, _ and -)",
+				labels:        "k1=###",
+				expectedError: true,
+			},
+			{
+				name:          "label value may not have over 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij1234=v",
+				expectedError: true,
+			},
+			{
+				name:          "label value cannot contain . and /",
+				labels:        "kubernetes_io_created-for_pvc_namespace=v./",
+				expectedError: true,
+			},
+			{
+				name:          "label key can have up to 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij123=v",
+				expectedError: false,
+			},
+			{
+				name:          "label value can have up to 63 characters",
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -55,6 +55,7 @@ See Github [Issues](https://github.com/kubernetes-sigs/gcp-compute-persistent-di
 |------------------|---------------------------|---------------|----------------------------------------------------------------------------------------------------|
 | type             | `pd-ssd` OR `pd-standard` | `pd-standard` | Type allows you to choose between standard Persistent Disks  or Solid State Drive Persistent Disks |
 | replication-type | `none` OR `regional-pd`   | `none`        | Replication type allows you to choose between Zonal Persistent Disks or Regional Persistent Disks  |
+| labels           | `key1=value1,key2=value2` |               | Labels allow you to assign custom GCE Disk labels                                                  |
```
mattcary, line 58: (dropping some notes here for future reference)

naming looks good, 'labels' is consistent with GCP label APIs:
https://cloud.google.com/sdk/gcloud/reference/compute/disks/add-labels
https://cloud.google.com/resource-manager/docs/creating-managing-labels

examples/kubernetes/demo-regional-restricted-sc.yaml
```diff
@@ -6,6 +6,7 @@ provisioner: pd.csi.storage.gke.io
 parameters:
   type: pd-standard
   replication-type: regional-pd
+  labels: key1=value1,key2=value2
```
mattcary, line 9: Is it necessary to add labels to all the sample storageclasses if we already have demo-labeled-sc.yaml?

Given the amount these examples are copied/pasted I fear having them everywhere would lead to a lot of key1=value1 labels out in the world which disturbs me aesthetically :-P

pkg/common/utils.go
```diff
@@ -148,3 +148,36 @@ func CreateNodeID(project, zone, name string) string {
...
+	if labels == "" {
+		return m, nil
+	}
+
+	errIncorrectFormat := fmt.Errorf("labels %q are invalid, correct format: 'key1=value1,key2=value2'", labels)
+
+	keyValueStrings := strings.Split(labels, labelsDelimiter)
+	for _, keyValue := range keyValueStrings {
+		keyValue := strings.Split(keyValue, labelsKeyValueDelimiter)
+
```
mattcary, line 168: Do we want to validate the characters in the key/values? It would be nice to get immediate errors for this sort of thing rather than waiting for a possibly obscured error from a GCP response. While generally I don't think we want to duplicate GCP validation logic, doing some reasonable sanity checks might make sense.

eg from https://cloud.google.com/resource-manager/docs/creating-managing-labels:

Keys have a minimum length of 1 character and a maximum length of 63 characters, and cannot be empty. Values can be empty, and have a maximum length of 63 characters.
Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes. All characters must use UTF-8 encoding, and international characters are allowed.
The key portion of a label must be unique. However, you can use the same key with multiple resources.
Keys must start with a lowercase letter or international character.

pkg/common/utils.go
```diff
@@ -148,3 +148,36 @@ func CreateNodeID(project, zone, name string) string {
...
+		if len(keyValue) != 2 {
+			return nil, errIncorrectFormat
+		}
+
+		key := strings.TrimSpace(keyValue[0])
+		if key == "" {
+			return nil, errIncorrectFormat
+		}
+
```
mattcary, line 177: Are empty values ok?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -668,6 +668,31 @@ func TestCreateVolumeArguments(t *testing.T) {
...
+			},
+			expVol: &csi.Volume{
+				CapacityBytes:      common.GbToBytes(20),
+				VolumeId:           testVolumeID,
+				VolumeContext:      nil,
+				AccessibleTopology: stdTopology,
+			},
+		},
+		{
+			name: "success with malformed labels parameter",
```
mattcary, line 687: Shouldn't this test be named "fail" and not "success"?

pkg/common/utils.go
ernestas2k: Yes, according to the link you shared above:

_Keys have a minimum length of 1 character and a maximum length of 63 characters, and cannot be empty. Values can be empty, and have a maximum length of 63 characters._

ernestas2k:
Thanks for fast feedback. I agree with all of the observations and I'll prepare updates.

Also, is there any integration test that I should update or actually introduce for this feature?

pkg/common/utils.go
mattcary: Well at least I found the link :-P

Thanks

mattcary:
> Regarding the validation, absolutely let's do it, with slight caveat that changes on google side will have to be reflected in the code too (hopefully labels api won't change often).
> 
Great, thanks.


> Also, is there any integration test that I should update or actually introduce for this feature?
> 
That's a great question. The storage class test/k8s-integrations/configs/sc-standard.yaml is what gets used for the integration tests, adding a label there would be a good idea.


> The kubernetes-integration and verify jobs failed:
> verify: `E0729 17:31:41.945] Build timed out`

It succeeded after your last push.  You can run it manually with hack/verify-all.sh.

> integration: `W0729 17:52:03.603] - The zone 'projects/k8s-boskos-gce-project-07/zones/us-central1-b' does not have enough resources available to fulfill the request. Try a different zone, or try again later.`

Trying again later :-)


> 
> Is there anything to be done from my side regarding these failures?
> 
> PS: just pushed newest changes to address the comments (please take a look if you are ok with regexps, error messages, etc). I'm ready to rework it so it's not too different from other functionality.

Will look shortly, thanks!

ernestas2k:
Awesome, I just added e2e test case (altough I lean towards updating existing test case, such as `It("Should create and delete disk with default zone", func() {`). Let me know which of the options you prefer 👍

pkg/common/utils.go
```diff
@@ -148,3 +149,60 @@ func CreateNodeID(project, zone, name string) string {
...
+func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
+	const labelsDelimiter = ","
+	const labelsKeyValueDelimiter = "="
+
+	labelsMap := make(map[string]string)
+	if labels == "" {
+		return labelsMap, nil
+	}
+
+	regexKey, _ := regexp.Compile("^[a-z][-_a-z0-9]{0,62}$")
```
mattcary, line 164: Use \p{Lowercase_Letter} instead of a-z? eg `^\p{Ll}[-_\p{Ll}0-9]{0,62}$`.

That way international characters are dealt with correctly, IIUC: https://www.regular-expressions.info/unicode.html

I tested it in go playground and it seems to work as expected.

mattcary:
> Awesome, I just added e2e test case `It("Should create and delete disk with labels", func() {`. To be honest, I lean more towards updating existing test case, such as `It("Should create and delete disk with default zone", func() {`.
> 
> What are your thoughts on this? 👍

I prefer adding a new test case, if it starts failing it will be a lot easier to tell what's going on; eg if the non-label version is still succeeding than we immediately know what's going on.

I think it would be good to add a label to the storage class in test/k8s-integeration/configs. This will be used for the k/k e2e tests and would be good to ferret out if something is depending on the storage class parameters in a way it shouldn't be. Also it will run on real PDs and give us, well, a real e2e test.

mattcary:
> I think it would be good to add a label to the storage class in test/k8s-integeration/configs. This will be used for the k/k e2e tests and would be good to ferret out if something is depending on the storage class parameters in a way it shouldn't be. Also it will run on real PDs and give us, well, a real e2e test.

Oops, you did that already. Thanks!

ernestas2k:
Updated the PR with requested changes. I appreciate the comments and quick feedback 👍

pkg/common/utils.go
```diff
@@ -161,15 +161,15 @@ func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
 		return labelsMap, nil
 	}
 
-	regexKey, _ := regexp.Compile("^[a-z][-_a-z0-9]{0,62}$")
+	regexKey, _ := regexp.Compile("^\\p{Ll}[\\p{Ll}0-9_-]{0,62}$")
```
mattcary, line 164: I'd prefer using a raw string (backtics) to avoid the doubled backslashes.

ernestas2k:
Done. Let me know if you find any more things to address.

saad-ali:
@ernestas2k could you also include the tags introduced in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/570 (i.e. `"kubernetes.io/created-for/pv/name": "pvc-32af883a-d7cc-11ea-86f9-42010a800053"`, `"kubernetes.io/created-for/pvc/name": "pvc-demo"`, `"kubernetes.io/created-for/pvc/namespace": "default"`, and `"storage.gke.io/created-by": "pd.csi.storage.gke.io"`) in this PR to add as labels as well?

ernestas2k:
> #570

Sure, let's get answers to few questions first:
- Google only supports alphanumerical chars + "-" + "_" in keys and values, the keys in the PR contain more than that.
- Are they non-overridable, meaning we add them to the Labels map[string]string last so users won't be able to override them?

One more thing, the PR has definitions of Tag on disk parameters. This make me a bit worried they're interrelated and (possibly) should even be the same thing as Label in this PR. Thoughts?

ernestas2k:
> @ernestas2k could you also include the tags introduced in #570 (i.e. `"kubernetes.io/created-for/pv/name": "pvc-32af883a-d7cc-11ea-86f9-42010a800053"`, `"kubernetes.io/created-for/pvc/name": "pvc-demo"`, `"kubernetes.io/created-for/pvc/namespace": "default"`, and `"storage.gke.io/created-by": "pd.csi.storage.gke.io"`) in this PR to add as labels as well?

Sure, let's get answers to few questions first:

Google only supports alphanumerical chars + "-" + "_" in keys and values, the keys in the PR contain more than that.
Are they non-overridable, meaning we add them to the Labels map[string]string last so users won't be able to override them?
One more thing, the PR has definitions of Tag on disk parameters. This make me a bit worried they're interrelated and (possibly) should even be the same thing as Label in this PR. Thoughts?

saad-ali:
Good questions.

> Sure, let's get answers to few questions first:
> 
> - Google only supports alphanumerical chars + "-" + "_" in keys and values, the keys in the PR contain more than that.

Looks like the values shouldn't be an issue, but the keys will be.

How about we have a function that "sanitizes" key and value strings: specifically this sanitize method could and replaces non-allowed characters with `_` (underscore) (e.g. "kubernetes.io/created-for/pv/name: pvc-32af883a-d7cc-11ea-86f9-42010a800053" would become "kubernetes.io_created-for_pv_name": "pvc-32af883a-d7cc-11ea-86f9-42010a800053"), it could trim strings longer than max allowed, etc.).

> - Are they non-overridable, meaning we add them to the Labels map[string]string last so users won't be able to override them?

There should be no good reason for user to want to override them so, yes let's make it so user won't be able to override.
 
> One more thing, the PR has definitions of Tag on disk parameters. This make me a bit worried they're interrelated and (possibly) should even be the same thing as Label in this PR. Thoughts?

The only reason we added those "tags" in "description" is for feature parity with the legacy in-tree GCE PD volume plugin (the old in-tree GCE PD volume plugin used put tags as JSON in the description field" since labels didn't exist at that time). Labels are the correct way of handling these so we want labels, but the description tags are there in case some old component still depends on them.

saikat-royc:
What would be the behavior of validating exisiting disk? do we comapre labels too? 
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L268
func (cloud *CloudProvider) ValidateExistingDisk
@saad-ali @mattcary

pkg/common/utils_test.go
```diff
@@ -295,3 +295,155 @@ func TestKeyToVolumeID(t *testing.T) {
...
+				labels:        "k*=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key may not have over 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij1234=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key can have up to 63 characters",
```
saikat-royc, line 414: nit: Can we group the failure cases and success cases for better readability

pkg/common/utils_test.go
```diff
@@ -295,3 +295,155 @@ func TestKeyToVolumeID(t *testing.T) {
...
+			{
+				name:          "label value may not have over 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij1234=v",
+				expectedError: true,
+			},
+			{
+				name:          "label value can have up to 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij123=v",
+				expectedError: false,
+			},
```
saikat-royc, line 432: Add some more positive test cases for keys and values which include combination of '-' and '_'. e.g blah-_

saikat-royc:
> Good questions.
> 
> > Sure, let's get answers to few questions first:
> > 
> > * Google only supports alphanumerical chars + "-" + "_" in keys and values, the keys in the PR contain more than that.
> 
> Looks like the values shouldn't be an issue, but the keys will be.
> 
> How about we have a function that "sanitizes" key and value strings: specifically this sanitize method could and replaces non-allowed characters with `_` (underscore) (e.g. "kubernetes.io/created-for/pv/name: pvc-32af883a-d7cc-11ea-86f9-42010a800053" would become "kubernetes.io_created-for_pv_name": "pvc-32af883a-d7cc-11ea-86f9-42010a800053"), it could trim strings longer than max allowed, etc.).
> 
> > * Are they non-overridable, meaning we add them to the Labels map[string]string last so users won't be able to override them?
> 
> There should be no good reason for user to want to override them so, yes let's make it so user won't be able to override.
> 
> > One more thing, the PR has definitions of Tag on disk parameters. This make me a bit worried they're interrelated and (possibly) should even be the same thing as Label in this PR. Thoughts?
> 
> The only reason we added those "tags" in "description" is for feature parity with the legacy in-tree GCE PD volume plugin (the old in-tree GCE PD volume plugin used put tags as JSON in the description field" since labels didn't exist at that time). Labels are the correct way of handling these so we want labels, but the description tags are there in case some old component still depends on them.

^\p{Ll}[\p{Ll}0-9_-]{0,62}$ does not accept '.' e.g. kubernetes.io_created-for_pv_name

pkg/common/utils_test.go
```diff
@@ -295,3 +295,155 @@ func TestKeyToVolumeID(t *testing.T) {
...
+			{
+				name:          "label key can only contain lowercase chars, digits, _ and -)",
+				labels:        "k*=v",
+				expectedError: true,
+			},
+			{
+				name:          "label key may not have over 63 characters",
+				labels:        "abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij1234=v",
+				expectedError: true,
```
saikat-royc, line 411: and couple of more invalid cases can be added for . and /
			{
				name:          "label key cannot contain . and /",
				labels:        "kubernetes.io/created-for/pvc/namespace=v",
				expectedError: true,
			},
			{
				name:          "label value cannot contain . and /",
				labels:        "kubernetes_io_created-for_pvc_namespace=v./",
				expectedError: true,
			},

mattcary:
> What would be the behavior of validating exisiting disk in CreateVolume calls? do we comapre labels too?

AFAICT we don't have the PV plumbed through so it would be hard to check even if we wanted to. We also would only want to check the labels in the case where the PV was dynamically provisioned, information which we also don't seem to have.

Also, if the user were to delete the labels, and we enforced validation, it would be very hard to recreate them correctly. Since AFAIU the labels are used for informational purposes (accounting etc) it would be better to have a tool to recreate them appropriately rather than risk breaking storage usage if they get out of sync.

mattcary:
@ernestas2k any updates---are you waiting on anything?

msau42:
Since the request for supporting the k8s metadata came in late, would it be simpler to split up into two PRs and treat the k8s metadata separately?

ernestas2k:
I'm in melancholic autumn mood right now, but since you pinged me I'll address your points this week 👍 . Unless you reconsider and split the task. Until then I'll address all of the points added.

mattcary:
> I'm in melancholic autumn mood right now, but since you pinged me I'll address your points this week 👍 . Unless you reconsider and split the task. Until then I'll address all of the points added.

Thanks! Summer is still in full swing here in Seattle and I still have a month before the autumn rains set in :-)

I think if you just make the test improvements Saikat mentioned, I'm okay with leaving the tag addition to a future CL. Let me know if you agree and I'll make in issue to track the tags (we're working on the filestore driver and will be adding labels in the same way).

ernestas2k:
Just updated the PR, addressed comments on test cases. 

Regarding kubernetes tags and sanitizing keys/values. Pretty straightforward change, but I would be surprized when strings that I've passed are accepted and **silently** converted. Take parameter `my_company/my_object_id=something/something-1-2-3` - conversion would change my key and value in some way that makes me not be able to query the disk anymore. At least, it would be a big surprise.

At most, I'd go with sanitizing the predefined keys of tag* constants (for example `tagKeyCreatedForClaimNamespace = "kubernetes.io/created-for/pvc/namespace"`). Tag values are also okay if it's guaranteed their values are always 'compliant' with google rules or the actual values aren't really that important. 

What are your thoughts?

mattcary:
> Just updated the PR, addressed comments on test cases.
> 
> Regarding kubernetes tags and sanitizing keys/values. Pretty straightforward change, but I would be surprized when strings that I've passed are accepted and **silently** converted. Take parameter `my_company/my_object_id=something/something-1-2-3` - conversion would change my key and value in some way that makes me not be able to query the disk anymore. At least, it would be a big surprise.
> 
> At most, I'd go with sanitizing the predefined keys of tag* constants (for example `tagKeyCreatedForClaimNamespace = "kubernetes.io/created-for/pvc/namespace"`). Tag values are also okay if it's guaranteed their values are always 'compliant' with google rules or the actual values aren't really that important.
> 
> What are your thoughts?

But since these labels are actually put on the PDs, what happens if an invalid character is passed through? Will PD creation fail or will something be silently dropped?

I think the common case would be that users would prefer to have labels actually created rather than having PD creation fail (which will occur some time after the SC creation and so be hard to debug).

We could also think about a validating webhook as is currently being done for snapshotting.

StokeHead:
Just wanted to check if this issue is still being worked on?
I followed #693 and #570 , however both dont seem to exactly implement what i would need.

We are applying different actions on GCE PDs according to the originating StorageClass. However with the description field implemented in #570 i am only able to see the PVC/PV name. Without querying the cluster itself its therefore not possible to get the originating StorageClass. (Or am i missing something?)
Having the possibility to add custom labels, or extending the description field with the originating storage class should either be of help in my use case.

mattcary:
I'll take over this PR and get it in. Watch #340 for the status.

/close

PR #571, opened 2020-08-07
Skip NodeExpandVolume for block volumes

### LINKED ISSUE #372: NodeExpand may resize fs for block volume
Looking back at `resizer.Resize` implementation, it checks the format on the disk before attempting a resize.  The problem is, the presence of a filesystem doesn't mean that the PVC is a Filesystem type.  A user could mount the volume as Block and manage an ext4/xfs filesystem themselves.

I think fixing this may require adding VolumeCapabilities to the Controller/NodeExpandVolumeRequest

/kind bug
/assign @davidz627 
cc @gnufied

saad-ali:
**What type of PR is this?**
>
/kind bug

Fixes #372

```release-note
Skip NodeExpandVolume for block volumes. 
```
/assign @msau42

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -461,6 +461,20 @@ func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpa
 		return nil, status.Error(codes.Internal, fmt.Sprintf("error when getting device path for %s: %v", volumeID, err))
 	}
 
+	volumeCapability := req.GetVolumeCapability()
+	if volumeCapability != nil {
+		// VolumeCapability is optional, if specified, validate it
+		if err := validateVolumeCapability(volumeCapability); err != nil {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("VolumeCapability is invalid: %v", err))
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -461,6 +461,20 @@ func (ns *GCENodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpa
...
 	}
 
+	volumeCapability := req.GetVolumeCapability()
+	if volumeCapability != nil {
+		// VolumeCapability is optional, if specified, validate it
+		if err := validateVolumeCapability(volumeCapability); err != nil {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("VolumeCapability is invalid: %v", err))
+		}
+
+		if blk := volumeCapability.GetBlock(); blk != nil {
```
msau42, line 471: Can you add a unit test for this?

saad-ali: The `TestNodeExpandVolume` unit tests are currently broken and require more extensive work to fix: see https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/573

saad-ali:
PTAL

PR #586, opened 2020-08-21
Import the kubernetes-csi/csi-release-tools subtree.

### LINKED ISSUE #585: Use kubernetes-csi/csi-release-tools instead of custom Makefile.
This repository currently uses a custom `Makefile` script to build the artifacts. CSI drivers are expected to use the kubernetes-csi/csi-release-tools repository for building.

See https://github.com/kubernetes-csi/csi-release-tools for details.

jeremyje:
**What type of PR is this?**

/kind feature

Imports the kubernetes-csi/csi-release-tools repository to align the build with other CSI projects.

This PR simply adds the repository, it does not modify the current `Makefile`, that'll be done in a future PR.

#585

I have done limited testing some things are currently broken but that's expected. Notably the e2e tests.

```release-note
NONE
```

### DIFF
--- a/release-tools/filter-junit.go
+++ b/release-tools/filter-junit.go
@@ -1,0 +1,133 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This command filters a JUnit file such that only tests with a name
+matching a regular expression are passed through. By concatenating
+multiple input files it is possible to merge them into a single file.
+*/
+package main
+
+import (
+	"encoding/xml"
+	"flag"
+	"io/ioutil"
+	"os"
+	"regexp"
+)
+
+var (
+	output = flag.String("o", "-", "junit file to write, - for stdout")
+	tests  = flag.String("t", "", "regular expression matching the test names that are to be included in the output")
+)
+
+/*
+ * TestSuite represents a JUnit file. Due to how encoding/xml works, we have
+ * represent all fields that we want to be passed through. It's therefore
+ * not a complete solution, but good enough for Ginkgo + Spyglass.
+ */
+type TestSuite struct {
+	XMLName   string     `xml:"testsuite"`
+	TestCases []TestCase `xml:"testcase"`
+}
+
+type TestCase struct {
+	Name      string     `xml:"name,attr"`
+	Time      string     `xml:"time,attr"`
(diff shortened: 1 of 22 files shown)


deploy/common.sh
```diff
@@ -26,5 +26,5 @@ function get_needed_roles()
 function ensure_kustomize()
 {
   ensure_var PKGDIR
-  ${PKGDIR}/deploy/kubernetes/install-kustomize.sh
+  "${PKGDIR}"/deploy/kubernetes/install-kustomize.sh
```
jingxu97, line 29: so this double quote is necessary?

Makefile
```diff
@@ -12,55 +12,9 @@
...
-REV=$(shell git describe --long --tags --match='v*' --dirty 2>/dev/null || git rev-list -n1 HEAD)
-ifdef GCE_PD_CSI_STAGING_VERSION
-	STAGINGVERSION=${GCE_PD_CSI_STAGING_VERSION}
-else
-	STAGINGVERSION=${REV}
-endif
-STAGINGIMAGE=${GCE_PD_CSI_STAGING_IMAGE}
-DRIVERBINARY=gce-pd-csi-driver
-DRIVERWINDOWSBINARY=${DRIVERBINARY}.exe
+.PHONY: push-multiarch-% push-multiarch
```
jingxu97, line 15: With this change, there are other places need to be change for building and releasing image. The parameters passed will be different.

msau42:
/ok-to-test
Can we also squash all of the commits into one?

jingxu97:
> Some of the content in csi-release-tools is very specific to the way we test csi-sidecars and do sidecar releases, and are not applicable here, such as bringing up a kind cluster for CI, and deploying the hostpath driver.
> 
> Will it be confusing if we include those irrelevant parts here? We are mainly interested in the build rules and not much else right?

bringing up a kind cluster is also useful for linux case? I feel csi-release-tool is gives a very organized way for building and releasing part. So we can use this consistently for all csi related components.

msau42:
I don't think we can test pd driver in a kind cluster

msau42:
I agree reusing the make rules is useful, but I don't know about the rest, like prow.sh and sidecar_release_process

jingxu97:
> I agree reusing the make rules is useful, but I don't know about the rest, like prow.sh and sidecar_release_process

another option is take the part we need from release-tool repo directly instead of vendering it if possible?

msau42:
/lgtm
/approve
Thanks!

release-tools/filter-junit.go
```diff
@@ -0,0 +1,133 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This command filters a JUnit file such that only tests with a name
```
msau42, line 18: @mattcary can you changes to combine the junit files be refactored now that it's actually imported in the repo?

mattcary: #591

Great point. It may be a little complicated as IIRC I made some changes to the junit-filter code which was built only as a command line tool and not a go library. But we should be able to dedup something, maybe by also extending the refactoring to the release tools.

PR #593, opened 2020-08-26
Do all storage class gingko runs and report errors at the end

### LINKED ISSUE #592: Run all storageclasses in k8s-integration even when test failure
Currently the test stops at the first ginkgo run. Errors should be combined.

/assign mattcary

mattcary:
/kind bug
/assign @saikat-royc 
Fixes #592

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/filter-junit.go
+++ b/test/k8s-integration/filter-junit.go
@@ -18,10 +18,13 @@ package main
 
 import (
 	"encoding/xml"
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"k8s.io/klog"
 )
 
 /*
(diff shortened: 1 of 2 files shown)


saikat-royc:
Can we potentially face issues in merging the junit files for failed test run for a given SC? The current [code](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/cc0326cabc686844aa5ec6328d048e0f41d512f3/test/k8s-integration/filter-junit.go#L65:6) seems like it would fail if a sub-dir is not found.

mattcary:
> Can we potentially face issues in merging the junit files for failed test run for a given SC? The current [code](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/cc0326cabc686844aa5ec6328d048e0f41d512f3/test/k8s-integration/filter-junit.go#L65:6) seems like it would fail if a sub-dir is not found.

My reasoning was that the common case is going to be for a test to fail partway through, so that all directories will be created and there will be some xmls to merge. I tested this by forcing a failure for one storage class and it worked fine.

If one of the ginkgo runs fails before it's even able to create a subdirectory, I think that would be the sign of something fundamentally broken. I'd expect that to be rare (?) and so okay if we fail during xml merge.

It's a good idea though to make the merging robust to a directory not being found, I'll make that change.

mattcary:

timeout error

PR #600, opened 2020-08-31
Move Dockerfile next to CSI main, remove $BASE_IMAGE arg.

jeremyje:
**What type of PR is this?**
/kind cleanup

Moves the `Dockerfile` from the root of the repository next to the binary. This better aligns with expectations of csi-release-tools and also allows for additional containers to be added in the future.

This PR also removes the `$BASE_IMAGE` variable from the `Dockerfile.windows` because it's always set to `servercore`. The base OS flavor is typically determined by the binary's dependencies so there's very little value to parameterize this field. This reduces the burden of implementing, https://github.com/kubernetes-csi/csi-release-tools/issues/100.

Partial fix for #585

```release-note
NONE
```


msau42:
We'll need to update our Louhi pipeline to use this new Dockerfile location too

msau42:
/assign @jingxu97 
to review windows changes.

msau42:
/hold

We're in the middle of releasing 1.0. I think we should wait on this until we're done to minimize disruption to our release pipeline

mattcary:
> We'll need to update our Louhi pipeline to use this new Dockerfile location too

Oh, good catch. It would be nice to have the louhi pipeline work off of the makefile rather than running docker directly. It looks like that's possible but we should make that change first (after the 1.0 release of course...)

PR #603, opened 2020-08-31
e2e test should take GKE node version as an input

saikat-royc:
For GKE clusters, node-version flag can be provided during cluster
creation to specify worker node k8s version. This change is needed
to test node and master version skew combinations.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #

```release-note
PD CSI e2e test infra should take GKE node version as an optional input argument.
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -167,6 +167,10 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 		cmdParams = append(cmdParams, "--enable-autorepair")
 	}
 
+	if isVariableSet(gkeNodeVersion) {
+		cmdParams = append(cmdParams, "--node-version", *gkeNodeVersion)
+	}
+
 	if useManagedDriver {
 		// PD CSI Driver add on is enabled only in gcloud beta.
 		cmdParams = append([]string{"beta"}, cmdParams...)
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -48,6 +48,7 @@ var (
 	imageType          = flag.String("image-type", "cos", "the image type to use for the cluster")
 	gkeReleaseChannel  = flag.String("gke-release-channel", "", "GKE release channel to be used for cluster deploy. One of 'rapid', 'stable' or 'regular'")
 	gkeTestClusterName = flag.String("gke-cluster-name", "gcp-pd-csi-driver-test-cluster", "GKE cluster name")
+	gkeNodeVersion     = flag.String("gke-node-version", "", "GKE cluster worker node version")
 
 	// Test infrastructure flags
 	boskosResourceType = flag.String("boskos-resource-type", "gce-project", "name of the boskos resource type to reserve")
--- a/test/run-k8s-integration-ci.sh
+++ b/test/run-k8s-integration-ci.sh
@@ -23,6 +23,7 @@ readonly image_type=${IMAGE_TYPE:-cos}
 readonly use_gke_managed_driver=${USE_GKE_MANAGED_DRIVER:-false}
 readonly gke_release_channel=${GKE_RELEASE_CHANNEL:-""}
 readonly teardown_driver=${GCE_PD_TEARDOWN_DRIVER:-true}
+readonly gke_node_version=${GKE_NODE_VERSION:-}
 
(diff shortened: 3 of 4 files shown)


test/run-k8s-integration-local.sh
```diff
@@ -73,6 +73,24 @@ make -C "${PKGDIR}" test-k8s-integration
...
+# This version of the command brings up (and subsequently tears down) a GKE
+# cluster using a release channel, and a given worker node version and with managed GCE PersistentDisk CSI driver add-on enabled, and points to
+# the local K8s repository to get the e2e test binary.
+# ${PKGDIR}/bin/k8s-integration-test --run-in-prow=false --service-account-file=${GCE_PD_SA_DIR}/cloud-sa.json \
+# --test-focus="External.Storage.*snapshot" --local-k8s-dir=$KTOP --storageclass-files=sc-standard.yaml,sc-balanced.yaml,sc-ssd.yaml \
+# --snapshotclass-file=pd-volumesnapshotclass.yaml --do-driver-build=false --teardown-driver=false \
+# --gce-zone="us-central1-c" --num-nodes=${NUM_NODES:-3} --gke-release-channel="rapid" --deployment-strategy="gke" \
+# --use-gke-managed-driver=true --teardown-cluster=true --node-version="1.16"
+
+# This version of the command brings up (and subsequently tears down) a GKE
```
mattcary, line 85: I don't know if this command is necessary.

Really all these commented versions I think are to easily test common cases. But I find if there's too many it's hard to confirm all the flags are right so I fiddle with the flags anyway in addition to uncommenting. If there's two many different commands the file gets hard to navigate.

msau42: What if we just have one or two commands, and we put each argument on a separate line and have it commented out if we don't want it by default?

saikat-royc: Removing this change. We can take up the cleanup in a different patch

mattcary: I like @msau42's suggestion.

It's not a huge deal either way as usually I have a locally tweaked version anyway. But I do want to make sure we don't clutter up this script too much.

+1 to removing the change for now and cleaning up after more bikeshedding....

test/k8s-integration/cluster.go
```diff
@@ -167,6 +167,10 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 		cmdParams = append(cmdParams, "--enable-autorepair")
 	}
 
+	if isVariableSet(gkeNodeVersion) {
```
verult, line 170: nit: should there be validation against specifying both release channel and node version?

mattcary: I don't think that's an error. I just ran

gcloud beta container clusters create --release-channel rapid --node-version 1.16.8-gke.15 skewed

and it succeeded. It might be useful to test such skewed deployments (although it looks like one might need to be careful about auto-upgrade---maybe that should be automatically disabled whenever both release channel and node version are specified? Or maybe we should just disable auto-upgrade all the time anyway as we probably never want that to happen during a test?)

saikat-royc: It works if we specify release channel and node version. Master version picks the release version, while nodes pickup the node version.

verult: ah ok thanks for checking. Yeah in hindsight if this were not the case, the skew tests we have planned with node version = master version - 2 would be tough.

Disabling auto-upgrade all the time makes sense to me. Removes the possible confusion during the rare occasion when a short-lived test cluster is running while an auto-upgrade is being pushed out.

verult:
/hold

Saikat do you think we should always disable auto upgrade? If so does it make sense to add it in this PR?

test/k8s-integration/cluster.go
saikat-royc: I see error when using release channels, and --no-enable-autoupgrade option.
e.g.
```
Auto_upgrade cannot be false when release_channel REGULAR is set
```

mattcary: Ok, let's defer the autoupgrade fiddling until we understand the use cases a bit better. It may be that it's not a problem for tests, we'd have to dig into the details more.

PR #604, opened 2020-09-02
Use a fixed version of kustomization

jingxu97:
The latest version of kustomize does not work for image tag transfer
This PR use a fixed version 3.8.0

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/install-kustomize.sh
+++ b/deploy/kubernetes/install-kustomize.sh
@@ -16,6 +16,50 @@ fi
 if [ -f "kustomize" ]; then
   rm kustomize
 fi
-echo "installing latest version of kustomize"
-curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+
+echo "installing kustomize"
+
+where=$PWD
+if [ -f $where/kustomize ]; then
+  echo "A file named kustomize already exists (remove it first)."
+  exit 1
+fi
+
+tmpDir=`mktemp -d`
+if [[ ! "$tmpDir" || ! -d "$tmpDir" ]]; then
+  echo "Could not create temp dir."
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


msau42:
I believe kubectl actually bundles in kustomize now. Maybe we don't need to even need to install kustomize separately?

Anyway, we can merge this to unblock testing.

/lgtm
/approve

deploy/kubernetes/install-kustomize.sh
```diff
@@ -16,6 +16,51 @@ fi
 if [ -f "kustomize" ]; then
   rm kustomize
 fi
-echo "installing latest version of kustomize"
-curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+
+echo "installing kustomize"
+#curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
```
mattcary, line 21: Can we remove this commented command in order to reduce clutter in the script?

One can always look at the git history if they want to know old ways of installing.

jingxu97:
> I believe kubectl actually bundles in kustomize now. Maybe we don't need to even need to install kustomize separately?
> 
> Anyway, we can merge this to unblock testing.
> 
> /lgtm
> /approve

I tested kubectl bundle, the version is too low and does not support image tag transfer

deploy/kubernetes/install-kustomize.sh
jingxu97: sure, I removed from my local and haven't pushed it yet .

jingxu97:
@msau42 could you add /lgtm again? thanks!

PR #607, opened 2020-09-08
Update to use csi proxy beta

jingxu97:
This PR updates driver to use csi-proxy beta version.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```


jingxu97:
cc @msau42 @verult

deploy/kubernetes/base/node_windows/node.yaml
```diff
@@ -43,6 +43,7 @@ spec:
           # Don't change base image without changing pdImagePlaceholder in
           # test/k8s-integration/main.go
           image: gke.gcr.io/gcp-compute-persistent-disk-csi-driver-win
+          imagePullPolicy: Always
```
msau42, line 46: We usually only set to always for dev environments where we're rebuilding the same image, but in general, we leave it unset (default)

pkg/mount-manager/safe-mounter_windows.go
```diff
@@ -22,16 +22,18 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
-	diskv1alpha1 "github.com/kubernetes-csi/csi-proxy/client/api/disk/v1alpha1"
-	diskclientv1alpha1 "github.com/kubernetes-csi/csi-proxy/client/groups/disk/v1alpha1"
+	diskv1beta1 "github.com/kubernetes-csi/csi-proxy/client/api/disk/v1beta1"
```
msau42, line 27: if we alias it to some common name, like "diskapi", then there's potentially less lines that need changing if we have to change api versions again

pkg/mount-manager/safe-mounter_windows.go
```diff
@@ -22,16 +22,18 @@ import (
...
+	diskclientv1beta1 "github.com/kubernetes-csi/csi-proxy/client/groups/disk/v1beta1"
 
 	fsv1alpha1 "github.com/kubernetes-csi/csi-proxy/client/api/filesystem/v1alpha1"
```
msau42, line 30: These apis don't need to be updated to beta?

pkg/mount-manager/safe-mounter_windows.go
```diff
@@ -142,30 +144,37 @@ func (mounter *CSIProxyMounter) Unmount(target string) error {
...
+	diskIDsResponse, err := mounter.DiskClient.ListDiskIDs(context.Background(), listRequest)
 	if err != nil {
 		return "", err
 	}
-	return getDiskNumberResponse.DiskNumber, nil
+	diskIDsMap := diskIDsResponse.GetDiskIDs()
+	for diskNum, diskInfo := range diskIDsMap {
+		klog.V(4).Infof("found disknum %s, diskInfo %v", diskNum, diskInfo)
+		idValue, found := diskInfo.Identifiers[id]
+		if found && strings.Contains(idValue, deviceName) {
```
msau42, line 157: Can you add a comment with an example format of idValue and deviceName?

`strings.Contains` is prone to substring matching. Do we have that issue here?

msau42 (commented):
Are we deploying the beta csi proxy in our CI cluster?

jingxu97:
> Are we deploying the beta csi proxy in our CI cluster?

yes, my change to install beta csi proxy is in 1.20.

pkg/mount-manager/safe-mounter_windows.go
jingxu97: so because filesystem and smb api groups do not have any changes since alpha, so we didn't move it to beta, considering they not as mature as others. We think the version of the software/code is independent of the version of the API groups.
 I discuss with @ddebroy and we will plan on moving them to beta too as post-beta action.

pkg/mount-manager/safe-mounter_windows.go
jingxu97: I added comment. From the tests, it seems working fine.

pkg/mount-manager/safe-mounter_windows.go
msau42: From an API perspective, alpha has no guarantee of backwards compatibility or upgrade support. I would be wary of using an alpha api in production.

jingxu97: right,  i agree. I think I should not say "they not as mature as others. " because there is no change. I support moving them to beta, the APIs are working fine with all the current drivers for windows. I will address it in csi-windows meetup.

deploy/kubernetes/base/node_windows/node.yaml
jingxu97: updated.

pkg/mount-manager/safe-mounter_windows.go
jingxu97: updated. thanks!

pkg/mount-manager/safe-mounter_windows.go
msau42: What if the idValue is something like "pvc-123", but the device name is "pvc-12"?

jingxu97: good point. I should use exact match

jingxu97:
> This PR updates driver to use csi-proxy beta version.
> 
> **What type of PR is this?**
> 
> > Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
> > /kind api-change
> > /kind bug
> > /kind cleanup
> > /kind design
> > /kind documentation
> > /kind failing-test
> > /kind feature
> > /kind flake
> 
> **What this PR does / why we need it**:
> 
> **Which issue(s) this PR fixes**:
> 
> Fixes #
> 
> **Special notes for your reviewer**:
> 
> **Does this PR introduce a user-facing change?**:
> 
> ```
> 
> ```

PR #610, opened 2020-09-19
Enable usages of application default credentials

mattcary:
/kind feature

This PR changes driver configuration to allow k8s-on-gce tests to be run using application default credentials, avoiding the security risks of creating service account keys.

Instructions for how to do this have been added to test/run-k8s-integration-local.sh.

```release-note
NONE
```


deploy/kubernetes/overlays/dev/kustomization.yaml
```diff
@@ -2,11 +2,12 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
-  gce-pd-csi-driver
-transformers:
-- ../../images/dev
-
+namespace: gce-pd-csi-driver
+# To change the dev image, add something like the following.
+#images:
+#- name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
+#  newName: gcr.io/mattcary-gke-dev-owned/csi/gce-pd-driver
+#  newTag: latest.
```
jingxu97, line 13: why you prefer to change image here instead of under /image dir?

mattcary: @verult 

I just figured this out this morning, it's a little complicated.... The actual change is to remove the change to the dev driver image, as currently the image in images/dev/images.yaml is not the same as the placeholder in base/.

I add a new overlay `noauth` which removes the secret volume. It uses the dev overlay as a resource. If the dev overlay uses the image resource, as far as I can tell it overrides the image before any images: transformer in the noauth overlay.

Because the images transform name: must match what's already in the yaml, that would mean for any images: transform (in overlays/noauth/kustomize.yaml) would have to match whatever dev is transformed into, that is, the newName in /images/dev/images.yaml (which is currently David's repo ;-)

When the pd csi driver k8s-integration test runs with --do-driver-build=true, the image tag to load is set by doing kustomize edit in the overlay dir, eg overlays/dev or overlays/noauth. That edit assumes the name is the placeholder (gke.gcr.io/gcp-compute-persiistent-disk-csi-driver). But, if the dev/ overlay has already changed the image name, then that kustomize edit isn't effective (since name: doesn't match anything).

Possibly one could fix this by doing a strategicMergePatch that matches the driver container rather than using an images transform? Then we wouldn't have this problem of import dependancies.

msau42: I think k8s-integration could potentially create its own image tag instead of overwriting the file. Do you think that would solve the problem?

The other way could be to move the noauth patch up one level and include it as part of the dev overlay itself.

mattcary: k8s-integration would have to know which images/ transform to change.

Moving the noauth patch up would help, I wasn't sure if we wanted to keep a use-key dev for those who aren't restricted about key downloads (because the extra cluster create parameters could break some workflows maybe)

I also looked into sharing the always-push change although the easiest way to do that means passing the no load restriction flag.

mattcary:
ping?

any other comments/suggestions, particularly on the overlay changes?

mattcary:
cmek failing, I shouldn't have changed anything that only impacted that test, so probably flake.

PR #619, opened 2020-10-01
Collect managed pd csi driver logs from node

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Collect managed pd csi driver logs from node
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -73,7 +73,8 @@ const (
 	pdImagePlaceholder        = "gke.gcr.io/gcp-compute-persistent-disk-csi-driver"
 	k8sInDockerBuildBinDir    = "_output/dockerized/bin/linux/amd64"
 	k8sOutOfDockerBuildBinDir = "_output/bin"
-	driverNamespace           = "gce-pd-csi-driver"
+	externalDriverNamespace   = "gce-pd-csi-driver"
+	managedDriverNamespace    = "kube-system"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


test/k8s-integration/main.go
```diff
@@ -74,6 +74,7 @@ const (
 	k8sInDockerBuildBinDir    = "_output/dockerized/bin/linux/amd64"
 	k8sOutOfDockerBuildBinDir = "_output/bin"
 	driverNamespace           = "gce-pd-csi-driver"
+	managedDriverNamespace    = "kube-system"
```
mattcary, line 77: can we make a function getDriverNamespace that does the *useGKEManagedDriver check? Otherwise someone will trip up in the future. And maybe change driverNamespace to externalDriverNamespace or something like that.

It looks like we only use it in a couple of places now; in installDriver we should keep using externalDriverNamespace explicitly.

test/k8s-integration/main.go
```diff
@@ -357,18 +358,18 @@ func handle() error {
...
-		if err != nil {
-			return fmt.Errorf("failed to start driver logging: %v", err)
-		}
-		defer func() {
-			if cancel != nil {
-				cancel()
-			}
-		}()
+	// Dump all driver logs to the test artifacts
+	cancel, dumpErr := dumpDriverLogs()
```
jingxu97, line 364: just to make sure, the only change here is to rename err to dumpErr?

mattcary: I think for consistency I'd prefer just naming it err, I don't think there's a shadowing problem?

saikat-royc: Done

PR #620, opened 2020-10-01
Wait on driver deployment rather than sleep

### LINKED ISSUE #614: Add pod wait function in  k8s integration test,  for pd driver start and also prepull image daemonset
See comments here https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/613#discussion_r493998132

Currently workflow, it waits for certain of of time for prepull calls driver or pr pd csi driver part.

mattcary:
/kind bug
Fixes #614 

Note that I don't think this addresses the "prepull imageset" part of the issue, I'm not sure I understand that part?

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/wait-for-driver.sh
+++ b/deploy/kubernetes/wait-for-driver.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# This script waits for the deployments from ./deploy-driver.sh to be ready.
+
+print_usage()
+{
+    printf "wait-for-driver.sh [--windows]\n"
+    printf "\t--windows: wait on the windows deployment rather than linux"
+    echo
+}
+
+node_daemonset=csi-gce-pd-node
+while [[ -n "${1-}" ]]; do
+  case $1 in
+    --windows )    shift
+                   node_daemonset=csi-gce-pd-node-win
+                   ;;
+    -h | --help )  print_usage
+                   exit 1
+                   ;;
+    * )            print_usage
+                   exit 1
+                   ;;
+  esac
+done
+
+kubectl wait -n gce-pd-csi-driver deployment csi-gce-pd-controller --for condition=available
+
+retries=90
+while [[ $retries -ge 0 ]];do
+    ready=$(kubectl -n gce-pd-csi-driver get daemonset "${node_daemonset}" -o jsonpath="{.status.numberReady}")
+    required=$(kubectl -n gce-pd-csi-driver get daemonset "${node_daemonset}" -o jsonpath="{.status.desiredNumberScheduled}")
+    if [[ $ready -eq $required ]];then
+        echo "Daemonset $node_daemonset found"
+        exit 0
+    fi
+    ((retries--))
+    sleep 10s
+done
+echo "Timeout waiting for node daemonset $node_daemonset"
+exit -1
+
--- a/test/k8s-integration/driver.go
+++ b/test/k8s-integration/driver.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"time"
 
 	"k8s.io/klog"
 	"k8s.io/kubernetes/test/e2e/framework/podlogs"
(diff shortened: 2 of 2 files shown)


deploy/kubernetes/wait-for-driver.sh
```diff
@@ -0,0 +1,42 @@
...
+    -h | --help )  print_usage
+                   exit 1
+                   ;;
+    * )            print_usage
+                   exit 1
+                   ;;
+  esac
+done
+
+kubectl wait -n gce-pd-csi-driver deployment csi-gce-pd-controller --for condition=available
```
msau42, line 27: any reason to use kubectl rather than the go client like we do in k/k e2es?

test/k8s-integration/driver.go
```diff
@@ -77,19 +76,25 @@ func installDriver(platform, goPath, pkgDir, stagingImage, stagingVersion, deplo
...
-	klog.Infof("Deploying driver")
-	// TODO (#139): wait for driver to be running
+
+	var waitCmd *exec.Cmd
+	waitScript := filepath.Join(pkgDir, "deploy", "kubernetes", "wait-for-driver.sh")
 	if platform == "windows" {
-		klog.Infof("Waiting 15 minutes for the driver to start on Windows")
-		time.Sleep(15 * time.Minute)
+		klog.Infof("will wait for windows deployment")
+		waitCmd = exec.Command(waitScript, "--windows")
```
jingxu97, line 90: Just a little clarification. for windows case, since we typically have a cluster with both linux and windows, we have daemonSets for both linux and windows deployed. 

In most of the cases, windows pods take more time linux pods. So when waiting for windows pods, linux pods should be ready. Or we can add wait for both just in case.

deploy/kubernetes/wait-for-driver.sh
```diff
@@ -0,0 +1,42 @@
...
+# This script waits for the deployments from ./deploy-driver.sh to be ready.
+
+print_usage()
+{
+    printf "wait-for-driver.sh [--windows]\n"
+    printf "\t--windows: wait on the windows deployment rather than linux"
+    echo
+}
+
+node_daemonset=csi-gce-pd-node
```
jingxu97, line 12: there are some functions like waitForDaemonSets in k8s.io/kubernetes/test/e2e/.., is it good to reuse those?

deploy/kubernetes/wait-for-driver.sh
mattcary: The driver is deployed here by shelling out to kustomize and using the deploy scripts. I wanted to be consistent with that, although since the deployment names etc have to be hard-coded into wait-for-driver.sh there's not much benefit except maybe because it lives in the same directory as the deployment scripts, if someone starts changing names around or whatever there's more chance of noticing that file.

It's also turned out to be nice to have a command-line utility that can be used manually with deploy-driver.sh.

I'm not super-attached to this method, though, so if you want to go over to client-go I could do that. It would be slightly annoying to have to set up the client since it's not used elsewhere in the driver but it's not a big deal.

test/k8s-integration/driver.go
mattcary: Good point, will do.

deploy/kubernetes/wait-for-driver.sh
mattcary: See my reply above to @msau42 on the existing driver using shell commands rather than client-go in the first place.

deploy/kubernetes/wait-for-driver.sh
msau42: Ah I forgot about that. That makes sense then. I wonder if it would be more convenient to put the wait in the deploy-driver.sh directly?

mattcary: IIUC deploy-driver.sh, whether or not we have a windows deployment depends only on if we have windows nodes in the cluster.

In driver.go we have a nice "windows" bool that's easy to check :-)

I think one could have a fancy jsonpath to get that from the node labels on the command line, but I couldn't figure it out.

mattcary:

CMEK failure. Trying again to see if it's a flake

mattcary:
Any further comments?

deploy/kubernetes/wait-for-driver.sh
```diff
@@ -0,0 +1,42 @@
...
+                   ;;
+    * )            print_usage
+                   exit 1
+                   ;;
+  esac
+done
+
+kubectl wait -n gce-pd-csi-driver deployment csi-gce-pd-controller --for condition=available
+
+retries=60
```
jingxu97, line 29: so the max waiting time is 10 mins. I have seen for windows it might take slightly longer than 10 mins to come up. How about set it to 12 mins?

mattcary: Good point updated it to 15m

PR #621, opened 2020-10-03
Add test support for regional PD

### LINKED ISSUE #408: add another storageclass test config to use Regional PDs in our external e2e
Now that we have gke regional cluster support, we could potentially add another storageclass test config to use Regional PDs in our external e2e tests.

mattcary:
/kind cleanup

Add regional PD to tests.

Fixes #408 

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -30,6 +30,10 @@ func gkeLocationArgs(gceZone, gceRegion string) (locationArg, locationVal string
 	return
 }
 
+func isRegionalGKECluster(gceZone, gceRegion string) bool {
+	return len(gceRegion) > 0
+}
+
 func clusterDownGCE(k8sDir string) error {
 	cmd := exec.Command(filepath.Join(k8sDir, "hack", "e2e-internal", "e2e-down.sh"))
 	err := runCommand("Bringing Down E2E Cluster on GCE", cmd)
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -49,6 +49,7 @@ var (
 	gkeReleaseChannel  = flag.String("gke-release-channel", "", "GKE release channel to be used for cluster deploy. One of 'rapid', 'stable' or 'regular'")
 	gkeTestClusterName = flag.String("gke-cluster-name", "gcp-pd-csi-driver-test-cluster", "GKE cluster name")
 	gkeNodeVersion     = flag.String("gke-node-version", "", "GKE cluster worker node version")
+	isRegionalCluster  = flag.Bool("is-regional-cluster", false, "tell the test that a regional cluster is being used. Should be used for running on an existing regional cluster (ie, --bringup-cluster=false). The test will fail if a zonal GKE cluster is created when this flag is true")
 
 	// Test infrastructure flags
 	boskosResourceType = flag.String("boskos-resource-type", "gce-project", "name of the boskos resource type to reserve")
@@ -74,6 +75,7 @@ const (
 	k8sOutOfDockerBuildBinDir = "_output/bin"
 	externalDriverNamespace   = "gce-pd-csi-driver"
 	managedDriverNamespace    = "kube-system"
+	regionalPDStorageClass    = "sc-regional.yaml"
 )
 
 func init() {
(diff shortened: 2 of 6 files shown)


test/k8s-integration/cluster.go
```diff
@@ -30,6 +30,10 @@ func gkeLocationArgs(gceZone, gceRegion string) (locationArg, locationVal string
 	return
 }
 
+func isRegionalGKECluster(gzeZone, gceRegion string) bool {
```
verult, line 33: typo: "gzeZone"

test/k8s-integration/main.go
```diff
@@ -270,6 +272,14 @@ func handle() error {
 		testDir = k8sDir
 	}
 
+	if *deploymentStrat == "gke" {
+		gkeRegional := isRegionalGKECluster(*gceZone, *gceRegion)
+		if *isRegionalCluster && !gkeRegional {
```
verult, line 277: What if we enforce that only one of `gceZone` and `gceRegion` could be set? Then we know the cluster is regional if `len(*gceRegion) > 0` and we don't need the `--is-regional-cluster` flag

test/k8s-integration/main.go
```diff
@@ -420,10 +430,23 @@ func handle() error {
...
 	if len(*storageClassFiles) != 0 {
-		storageClasses := strings.Split(*storageClassFiles, ",")
+		applicableStorageClassFiles := []string{}
+		for _, rawScFile := range strings.Split(*storageClassFiles, ",") {
+			scFile := strings.TrimSpace(rawScFile)
+			if len(scFile) == 0 {
+				continue
+			}
+			if scFile == regionalPDStorageClass && !*isRegionalCluster {
+				continue // Regional PDs cannot be used in a zonal cluster.
```
verult, line 440: Is there a way to log a warning?

test/run-k8s-integration-ci.sh
```diff
@@ -31,8 +31,10 @@ make -C "${PKGDIR}" test-k8s-integration
 
 base_cmd="${PKGDIR}/bin/k8s-integration-test \
             --run-in-prow=true --service-account-file=${E2E_GOOGLE_APPLICATION_CREDENTIALS} \
-            --do-driver-build=${do_driver_build} --teardown-driver=${teardown_driver} --boskos-resource-type=${boskos_resource_type} \
-            --storageclass-files=sc-standard.yaml,sc-balanced.yaml,sc-ssd.yaml --snapshotclass-file=pd-volumesnapshotclass.yaml \
+            --do-driver-build=${do_driver_build} --teardown-driver=${teardown_driver} \
+            --boskos-resource-type=${boskos_resource_type} \
+            --storageclass-files=sc-standard.yaml,sc-balanced.yaml,sc-ssd.yaml,sc-regional.yaml \
```
verult, line 36: nit: Set this flag in the `[ -z "$gce_region" ]` branch so it's more explicit that `sc-regional.yaml` is only expected to work in regional clusters

test/k8s-integration/main.go
mattcary: We still need --is-regional-cluster for when you're using a pre-existing cluster---you don't set the region or zone in that case and just use your current kubeconfig.

test/k8s-integration/cluster.go
mattcary: done

test/k8s-integration/main.go
mattcary: Note that we already check that only one of these is set in main.go:112

test/k8s-integration/main.go
mattcary: it's actually expected as we don't have a separate run-k8s-integration script for the regional job.

I'll add a warning but it will be a little noisy.

test/run-k8s-integration-ci.sh
mattcary: great point, thanks.

test/k8s-integration/main.go
verult: Gotcha sounds good

nit: maybe make this field targeted to pre-existing clusters only, i.e. only make the flag settable if `gceZone` and `gceRegion` are unset, and update the flag description. Otherwise it could get confusing for someone new to understand when `is-regional-cluster` should be set just from looking at the flag description

verult (commented):
lgtm, I could add the tag after the new commits come in

test/k8s-integration/main.go
mattcary: Good idea will do.

mattcary: (added in the bringupCluster check)

mattcary:
In the push above I changed the CI script to only use the regional SC on a regional cluster. I don't know if there's a good way to test that change?

verult:
Yeah letting it run on CI might be the easiest, altho the other changes besides the script could be tested by updating one of the local scripts

mattcary:
I ran what I could through local testing (--is-regional, etc).

mattcary:
secrets flake.

PR #623, opened 2020-10-07
Add PSP for the controller Deployment

george-angel:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Adds PSP for the controller Deployment. Without it, clusters with default PSP having basic permissions, the controller will not be allowed to run.

Fixes #

No

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/base/controller/psp.yaml
+++ b/deploy/kubernetes/base/controller/psp.yaml
@@ -1,0 +1,18 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: csi-gce-pd-controller-psp
+spec:
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  volumes:
+    - "configMap"
+    - "emptyDir"
+    - "secret"
+  hostNetwork: true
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/base/controller/psp.yaml
```diff
@@ -0,0 +1,18 @@
...
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  volumes:
+    - "configMap"
```
mattcary, line 15: Why allow configMaps if the controller isn't currently using any?

mattcary (commented):
Thanks! This is a good idea, thanks for the PR.

One small comment above.

deploy/kubernetes/base/controller/psp.yaml
george-angel: @mattcary good catch, no reason. I just copied our patch, removed.

mattcary:
/lgtm
/approve

Thanks again for your contribution!

mattcary:
I think you need to update the description as described [here](https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md#applying-a-release-note) in order to get the release note nag label removed.

PR #624, opened 2020-10-07
Wait for windows prepull images rather than sleeping

### LINKED ISSUE #614: Add pod wait function in  k8s integration test,  for pd driver start and also prepull image daemonset

mattcary:
/kind bug

Fixes #614

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/prepull-image.sh
+++ b/test/k8s-integration/prepull-image.sh
@@ -4,22 +4,40 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [[ -z "${PREPULL_IMAGE}" ]]; then
+# This is taken from prepull.yaml.
+readonly prepull_daemonset=prepull-test-containers
+
+wait_on_prepull()
+{
+    # Wait up to 15 minutes for the test images to be pulled onto the nodes.
+    retries=90
+    while [[ $retries -ge 0 ]];do
+        ready=$(kubectl get daemonset "${prepull_daemonset}" -o jsonpath="{.status.numberReady}")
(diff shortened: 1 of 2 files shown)


test/k8s-integration/prepull-image.sh
```diff
@@ -4,22 +4,40 @@ set -o nounset
...
-if [[ -z "${PREPULL_IMAGE}" ]]; then
+# This is taken from prepull.yaml.
+readonly prepull_daemonset=prepull-test-containers
+
+wait_on_prepull()
+{
+    # Wait up to 15 minutes for the test images to be pulled onto the nodes.
+    retries=90
+    while [[ $retries -ge 0 ]];do
+        ready=$(kubectl get daemonset "${prepull_daemonset}" -o jsonpath="{.status.numberReady}")
```
jingxu97, line 15: this logic is similar to the other wait script, maybe we can consolidate to a common function later?

mattcary: I think we should probably move to client-go and have it in code. That's a bigger change than I want to do right now though.

Consolidating would mean have a shared bash include directory between deploy/kubernetes and test/k8s-integration which starts to get messy.

mattcary: #625 created for the client-go work.

PR #632, opened 2020-10-13
Enable XFS selectively for GKE clusters with proper COS support

mattcary:
/kind bug

Since https://github.com/kubernetes/kubernetes/pull/95527 enables XFS on COS in the e2e tests, the PD CSI tests need to selectively disable it when the test on GKE versions without newer COS versions.

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/driver-config.go
+++ b/test/k8s-integration/driver-config.go
@@ -27,15 +27,15 @@ const (
 
 // generateDriverConfigFile loads a testdriver config template and creates a file
 // with the test-specific configuration
-func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassFile, deploymentStrat string) (string, error) {
+func generateDriverConfigFile(testParams *testParameters, storageClassFile string) (string, error) {
 	// Load template
-	t, err := template.ParseFiles(filepath.Join(pkgDir, testConfigDir, configTemplateFile))
+	t, err := template.ParseFiles(filepath.Join(testParams.pkgDir, testConfigDir, configTemplateFile))
 	if err != nil {
 		return "", err
 	}
 
 	// Create destination
-	configFilePath := filepath.Join(pkgDir, testConfigDir, configFile)
+	configFilePath := filepath.Join(testParams.pkgDir, testConfigDir, configFile)
 	f, err := os.Create(configFilePath)
 	if err != nil {
 		return "", err
@@ -55,9 +55,11 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 		"exec",
 		"multipods",
 		"topology",
+		"controllerExpansion",
+		"nodeExpansion",
 	}
 	var fsTypes []string
-	if platform == "windows" {
+	if testParams.platform == "windows" {
 		fsTypes = []string{"ntfs"}
 		caps = []string{
 			"persistence",
@@ -70,7 +72,6 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 			"ext2",
 			"ext3",
 			"ext4",
-			"xfs",
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


mattcary:
/assign @msau42 

Note: this compiles but I haven't tested it yet, I wanted to make sure this made sense to you first

test/k8s-integration/driver-config.go
```diff
@@ -124,3 +138,38 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 
 	return configFilePath, nil
 }
+
+func getGKEClusterVersion() (major, minor int, err error) {
+	if *gkeClusterVer == "" && *gkeReleaseChannel == "" {
+		return 0, 0, fmt.Errorf("No GKE version found")
+	}
+	// TODO: the release channels are hardcoded (as is latest, below), which means they will need to be updated at each GKE release. It would be better to find the version automatically, maybe through gcloud container get-server-config.
```
msau42, line 146: I thought @saikat-royc had implemented something that automatically gets the version of the release channel?

test/k8s-integration/driver-config.go
```diff
@@ -124,3 +138,38 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 
 	return configFilePath, nil
 }
+
+func getGKEClusterVersion() (major, minor int, err error) {
```
msau42, line 142: I remember @saikat-royc had already implemented some utilities that did this?

test/k8s-integration/driver-config.go
saikat-royc: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/version.go

saikat-royc: Also instead of statically mapping release channel to a cluster version(as it can change), we can do something similar to the logic of generating skip string, where we read the k8s cluster version after the cluster is deployed and use mustParseVersion() to parse the gke version (since we know the deployment strategy), and compare versions. generateDriverConfigFile comes after cluster is deployed
[here](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/12ea362bb378bb52fbbbb064407889719092ffea/test/k8s-integration/main.go#L421) and [here](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/12ea362bb378bb52fbbbb064407889719092ffea/test/k8s-integration/main.go#L427)

mattcary: Ah, thank you, I wrote this PR quickly and didn't have time to figure out exactly when this was called, it looked like it was  early.

If the cluster is up than your suggestion is good, I'll do that. Thanks for saving me the trouble of figuring out the sequencing :-)

test/k8s-integration/driver-config.go
```diff
@@ -82,22 +81,32 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
...
 	*/
 
-	// TODO: Support adding/removing capabilities based on Kubernetes version.
-	switch deploymentStrat {
+	switch testParams.deploymentStrategy {
 	case "gke":
-		fallthrough
-	case "gce":
 		caps = append(caps, "controllerExpansion", "nodeExpansion")
+		if *imageType == "cos" {
```
saikat-royc, line 87: imageType can be a part of testParams too?

saikat-royc:
The change looks good to me overlall with some minor comment

test/k8s-integration/driver-config.go
mattcary: Yup, thanks, I used them instead.

test/k8s-integration/driver-config.go
mattcary: Done

test/k8s-integration/driver-config.go
mattcary: Indeed you're right, done.

test/k8s-integration/driver-config.go
```diff
@@ -82,22 +81,32 @@ func generateDriverConfigFile(platform, pkgDir, storageClassFile, snapshotClassF
 	   dataSource
 	*/
 
-	// TODO: Support adding/removing capabilities based on Kubernetes version.
-	switch deploymentStrat {
+	switch testParams.deploymentStrategy {
 	case "gke":
-		fallthrough
-	case "gce":
 		caps = append(caps, "controllerExpansion", "nodeExpansion")
```
saikat-royc, line 86: For "GCE", looks like controllerexpansion and node expansion are getting skipped?

mattcary: Thanks, added above since they're used in all deployments now.

mattcary:
rate limit exceeded on the e2e test

/retest

test/k8s-integration/main.go
```diff
@@ -287,9 +314,9 @@ func handle() error {
 		var err error = nil
 		switch *deploymentStrat {
 		case "gce":
-			err = clusterUpGCE(k8sDir, *gceZone, *numNodes, *imageType)
+			err = clusterUpGCE(k8sDir, *gceZone, *numNodes, testParams.imageType)
 		case "gke":
-			err = clusterUpGKE(*gceZone, *gceRegion, *numNodes, *imageType, *useGKEManagedDriver)
+			err = clusterUpGKE(*gceZone, *gceRegion, *numNodes, testParams.imageType, *useGKEManagedDriver)
```
saikat-royc, line 319: Let's use testParams.useGKEManagedDriver for uniformity? similar to line 386

mattcary: Oops, thanks, done. Also updated the deploymentStrategy.

PR #635, opened 2020-10-20
Enable dump GKE node logs

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

Capturing the flow here: 
This patch uses kubetest --dump flag instead of directly setting up --report-dir flag. If dump flag is provided, kubetest will setup report-dir [here](https://github.com/kubernetes/test-infra/blob/master/kubetest/e2e.go#L40). Also using dump flag, enables kubetest to call DumpClusterLogs for both GCE and GKE deployments. For GKE, the function will setup the enviroment variables and filters [here](https://github.com/kubernetes/test-infra/blob/master/kubetest/gke.go#L409). For GKE deployment, this enables log-dump.sh to find the [node names](https://github.com/kubernetes/kubernetes/blob/master/cluster/log-dump/log-dump.sh#L400) for GKE and also gracefully skip master dump. The above dump flag works for both zonal and regional GKE clusters.

Also with this change, we do not need the change https://github.com/kubernetes/kubernetes/pull/95088. I will be closing that PR once this one is reviewed.

```release-note
Enable dump GKE node logs
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -560,15 +560,15 @@ func runTestsWithConfig(testDir, testFocus, testSkip, testConfigArg string, clou
 
 	artifactsDir, ok := os.LookupEnv("ARTIFACTS")
 	reportArg := ""
+	kubetestDumpDir := ""
 	if ok {
 		if len(reportPrefix) > 0 {
-			reportDir := filepath.Join(artifactsDir, reportPrefix)
-			if err := os.MkdirAll(reportDir, 0755); err != nil {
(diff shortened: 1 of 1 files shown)


tosi3k:
I'm not sure whether I'm an appropriate person for reviewing this PR given I don't own this code but OTOH I was the one that suggested that fix :D.

/lgtm

PR #637, opened 2020-10-21
Add Volume Resize support for Windows

jingxu97:
This PR adds the volume resize support for Window

Have tested volume expansion test cases locally. Will open another PR to enable the tests.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Add volume expansion support for Windows in GCE PD CSI driver
```


pkg/mount-manager/safe-mounter_windows.go
```diff
@@ -251,3 +251,11 @@ func (mounter *CSIProxyMounter) ExistsPath(path string) (bool, error) {
 		})
 	return isExistsResponse.Exists, err
 }
+
+func (mounter *CSIProxyMounter) GetBlockSizeBytes(diskId string) (int64, error) {
+	DiskStatsResponse, err := mounter.DiskClient.DiskStats(context.Background(),
```
msau42, line 256: You mentioned this will return a smaller size than the actual disk? Will that cause problems with resize returning success?

pkg/resizefs/resizefs_windows.go
```diff
@@ -0,0 +1,69 @@
...
+	idResponse, err := proxy.VolumeClient.GetVolumeIDFromMount(context.Background(), idRequest)
+	if err != nil {
+		return false, err
+	}
+	volumeId := idResponse.GetVolumeId()
+
+	request := &volumeapi.ResizeVolumeRequest{
+		VolumeId: volumeId,
+	}
+	_, err = proxy.VolumeClient.ResizeVolume(context.Background(), request)
```
msau42, line 64: Do you need to check filesystem type or does this imply ntfs?

pkg/mount-manager/safe-mounter_windows.go
jingxu97: Before I was using volumestats API, now I changed to DiskStats which will return the full disk size. So the number can match now.

pkg/resizefs/resizefs_windows.go
jingxu97: It implies ntfs, no need to check fs type.

jingxu97:
@msau42 could you also check this one so that when tests are triggered, resize support is available?

pkg/resizefs/resizefs_windows.go
msau42: if there ever is a need for multiple fs support, will the csi proxy be able to automatically figure out what to do?

msau42:
/hold
can you add a release note?

pkg/resizefs/resizefs_windows.go
jingxu97: for this volumeclient API, it means ntfs volume, and all functions through volume API client assumes that. 
So far there is no other type to support resize. In case, for example, smb support resize, then the API will be in smb API group.

jingxu97:
> /hold
> can you add a release note?

added. THanks!

pkg/resizefs/resizefs_windows.go
msau42: I'm thinking something like ext4 in the future, where it also belongs in the volume client. Anyway, I posed the question in slack. For this pr, this is fine.

PR #643, opened 2020-10-28
Migrate from alpha to beta compute APIs

alculquicondor:
**What type of PR is this?**

/kind cleanup

The only Disk feature that required an alpha API (multiwriter) is now in beta.
We keep the infrastructure to be able to swap API versions, but it has been simplified to reduce code duplication.

Builds on top of #642

```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/fake-gce.go
+++ b/pkg/gce-cloud-provider/compute/fake-gce.go
@@ -134,9 +134,9 @@ func (cloud *FakeCloudProvider) ListDisks(ctx context.Context, maxEntries int64,
 	}
 
 	for name, cd := range cloud.disks {
-		// Only return zonal disks for simplicity
+		// Only return v1 disks for simplicity
 		if !seen.Has(name) {
-			d = append(d, cd.ZonalDisk)
+			d = append(d, cd.disk)
 			seen.Insert(name)
 			count++
 		}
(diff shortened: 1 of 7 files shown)


pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -260,45 +260,32 @@ func (cloud *FakeCloudProvider) InsertDisk(ctx context.Context, volKey *meta.Key
 		}
 	}
 
-	var diskToCreate *CloudDisk
+	computeDisk := &computev1.Disk{
+		Name:             volKey.Name,
+		Zone:             volKey.Zone,
```
mattcary, line 265: It looks like computeDisk.Zone is set in the meta.Zonal case below, and unused in the meta.Regional case. Maybe best not to set it here?

mattcary (commented):
Thanks for cleaning up the disk organization, it was getting a little silly with all the types.

/lgtm

pkg/gce-cloud-provider/compute/fake-gce.go
alculquicondor: You are right. Removed

alculquicondor:
/assign @msau42 
for approval

mattcary:
/approve

oops

PR #646, opened 2020-10-30
Revert sidecar upgrade change for staging-rc

Jiawei0227:
**What type of PR is this?**
/kind bug

This PR revert the change to staging-rc on the sidecar upgrades as the sidecar upgrades only apply for v1.17+ and that will break current testing for v1.15 and v1.16. So until we properly set the overlay, we will not upgrade these.

Only keep csi-snapshotter to the latest version as it does not rely on k8s version and the v3 snapshotter has some test flakiness fix.

Fixes #

```release-note
NONE
```

/cc @jingxu97 
/cc @msau42

### DIFF
--- a/deploy/kubernetes/images/prow-gke-release-staging-rc/image.yaml
+++ b/deploy/kubernetes/images/prow-gke-release-staging-rc/image.yaml
@@ -4,7 +4,7 @@ metadata:
   name: imagetag-csi-provisioner-prow-rc
 imageTag:
   name: k8s.gcr.io/sig-storage/csi-provisioner
-  newTag: "v2.0.4"
+  newTag: "v1.6.0"
 ---
 
 apiVersion: builtin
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/images/prow-gke-release-staging-rc/image.yaml
```diff
@@ -25,7 +25,7 @@ metadata:
 imageTag:
   name: gke.gcr.io/csi-resizer
   newName: gcr.io/gke-release-staging/csi-resizer
-  newTag: "v1.0.1-gke.0"
+  newTag: "v0.5.0-gke.0"
```
msau42, line 28: can we keep newer resizer and snapshotter? The new versions have bug fixes that helped deal with test flakes.

Jiawei0227: Done

jingxu97:
@Jiawei0227 can you rebase?

msau42:
I think we should just create another overlay for older releases. By reverting to the older images, it means we will no longer qualify the newer sidecars for later releases.

Jiawei0227:
I think Saad has already upgraded all the images version on stable overlay... So this PR only fix the staging-rc test now.

Jiawei0227:
And yes. I agree we should create another overlay.

PR #662, opened 2020-11-09
Update gce operation timeout

jingxu97:
Update gce operation timeout to 2 mins instead of 5 mins. The 90% of
attach/detach disk is less than 15 second, and create volume is less
than 1 min.

Reduce the timeout of volume operation so that it can recover from
previous operation quicker in case operation is dropped.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Update gce volume operation timeout.
```

### DIFF
--- a/deploy/kubernetes/base/controller/controller.yaml
+++ b/deploy/kubernetes/base/controller/controller.yaml
@@ -32,7 +32,7 @@ spec:
             - "--enable-leader-election"
             - "--leader-election-type=leases"
             - "--leader-election-namespace=$(PDCSI_NAMESPACE)"
-            - "--timeout=250s"
+            - "--timeout=180s"
             - "--extra-create-metadata"
           # - "--run-controller-service=false" # disable the controller service of the CSI driver
           # - "--run-node-service=false"       # disable the node service of the CSI driver
@@ -52,7 +52,7 @@ spec:
             - "--metrics-address=:22012"
             - "--leader-election"
             - "--leader-election-namespace=$(PDCSI_NAMESPACE)"
-            - "--timeout=250s"
+            - "--timeout=600s"
           env:
             - name: PDCSI_NAMESPACE
               valueFrom:
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -46,6 +46,28 @@ const (
 	GCEAPIVersionV1 GCEAPIVersion = "v1"
 	// Alpha key type
 	GCEAPIVersionBeta GCEAPIVersion = "beta"
+	// OpPollInterval is the ineterval between polling an operation
+	OpPollInterval = 1 * time.Second
+	// OpTimeout is the default total timeout for polling on an operation
+	OpTimeout = 2 * time.Minute
+	// defaultCallTimeout is the ddefault context timeout for creating/getting on an operation
+	defaultCallTimeout = 10 * time.Minute
+	// Time interval for checking on an operation
+	AttachDiskInitialDelay = 6 * time.Second
+
+	InsertDiskInitialDelay = 3 * time.Second
+
+	// volumeAttachmentConsecutiveErrorLimit is the number of consecutive errors we will ignore when waiting for a volume to attach/detach
+	//volumeAttachmentStatusConsecutiveErrorLimit = 10
+
+	// Attach typically takes 2-5 seconds (average is 2). Asking before 2 seconds is just waste of API quota.
+	volumeAttachmentStatusInitialDelay = 2 * time.Second
+	// Detach typically takes 5-10 seconds (average is 6). Asking before 5 seconds is just waste of API quota.
+	volumeDetachmentStatusInitialDelay = 5 * time.Second
+	// After the initial delay, poll attach/detach with exponential backoff (2046 seconds total)
+	volumeAttachmentStatusPollDelay = 2 * time.Second
+	volumeAttachmentStatusFactor    = 2
+	volumeAttachmentStatusSteps     = 11
 )
 
 type GCECompute interface {
@@ -306,6 +328,11 @@ func ValidateDiskParameters(disk *CloudDisk, params common.DiskParameters) error
 	return nil
 }
 
+// ContextWithCallTimeout returns a context with a default timeout, used for generated client calls.
+func ContextWithCallTimeout() (context.Context, context.CancelFunc) {
+	return context.WithTimeout(context.Background(), defaultCallTimeout)
+}
+
 func (cloud *CloudProvider) InsertDisk(ctx context.Context, volKey *meta.Key, params common.DiskParameters, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID string, multiWriter bool) error {
 	klog.V(5).Infof("Inserting disk %v", volKey)
 
@@ -489,17 +516,20 @@ func (cloud *CloudProvider) insertZonalDisk(
 		}
 	}
 
+	callCtx, cancel := ContextWithCallTimeout()
+	defer cancel()
+
 	if gceAPIVersion == GCEAPIVersionBeta {
 		var insertOp *computebeta.Operation
 		betaDiskToCreate := convertV1DiskToBetaDisk(diskToCreate)
 		betaDiskToCreate.MultiWriter = multiWriter
-		insertOp, err = cloud.betaService.Disks.Insert(cloud.project, volKey.Zone, betaDiskToCreate).Context(ctx).Do()
+		insertOp, err = cloud.betaService.Disks.Insert(cloud.project, volKey.Zone, betaDiskToCreate).Context(callCtx).Do()
 		if insertOp != nil {
 			opName = insertOp.Name
 		}
 	} else {
 		var insertOp *computev1.Operation
-		insertOp, err = cloud.service.Disks.Insert(cloud.project, volKey.Zone, diskToCreate).Context(ctx).Do()
+		insertOp, err = cloud.service.Disks.Insert(cloud.project, volKey.Zone, diskToCreate).Context(callCtx).Do()
 		if insertOp != nil {
 			opName = insertOp.Name
 		}
@@ -523,7 +553,7 @@ func (cloud *CloudProvider) insertZonalDisk(
 		}
 		return fmt.Errorf("unknown Insert disk error: %v", err)
 	}
-
+	klog.Infof("Insert disk %s op id %s", volKey.Name, opName)
 	err = cloud.waitForZonalOp(ctx, opName, volKey.Zone)
 
 	if err != nil {
@@ -568,6 +598,7 @@ func (cloud *CloudProvider) deleteZonalDisk(ctx context.Context, zone, name stri
 		}
 		return err
 	}
+	klog.Infof("delete disk %s op id %s", name, op.Name)
 	err = cloud.waitForZonalOp(ctx, op.Name, zone)
 	if err != nil {
 		return err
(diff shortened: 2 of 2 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -610,6 +610,8 @@ func (cloud *CloudProvider) AttachDisk(ctx context.Context, volKey *meta.Key, re
 	op, err := cloud.service.Instances.AttachDisk(cloud.project, instanceZone, instanceName, attachedDiskV1).Context(ctx).Do()
 	if err != nil {
 		return fmt.Errorf("failed cloud service attach disk call: %v", err)
+	} else {
+		klog.V(5).Infof("Attaching disk %s operation id is ", volKey, op.Name)
```
mattcary, line 614: Also log op.Name at line 618? There may be lots of logging between these two line so it would be easier to debug a failed waitForZonalOp if we had the op name.

jingxu97:
cc @verult @saad-ali

pkg/gce-cloud-provider/compute/gce-compute.go
jingxu97: added, thanks!

mattcary:
Thanks!

/lgtm

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -681,7 +683,7 @@ func (cloud *CloudProvider) waitForZonalOp(ctx context.Context, opName string, z
 	// The v1 API can query for v1, alpha, or beta operations.
 	svc := cloud.service
 	project := cloud.project
-	return wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
+	return wait.Poll(3*time.Second, 2*time.Minute, func() (bool, error) {
```
saikat-royc, line 686: can we create constants for these

jingxu97: sure, updated. thanks!

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -681,7 +687,7 @@ func (cloud *CloudProvider) waitForZonalOp(ctx context.Context, opName string, z
 	// The v1 API can query for v1, alpha, or beta operations.
 	svc := cloud.service
 	project := cloud.project
-	return wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
+	return wait.Poll(OpInterval, OpTimeout, func() (bool, error) {
```
msau42, line 690: Should we honor the client context.Timeout value instead?

jingxu97: Attacher side set context timeout to 15 second. https://github.com/kubernetes-csi/external-attacher/blob/298cb965f909f76f2d2dfa2062cb5e142f14b9d5/cmd/csi-attacher/main.go#L58

my understanding is this context timeout will be used for each individual operation for calling gce api. So we need use another set up timeout for this polling. 
```
op, err := cloud.service.Instances.AttachDisk(cloud.project, instanceZone, instanceName, attachedDiskV1).Context(ctx).Do()
pollOp, err := cloud.service.RegionOperations.Get(cloud.project, region, opName).Context(ctx).Do()
```

msau42: We override the attacher timeout: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/base/controller/controller.yaml#L55

jingxu97: ah, that makes sense now. The test failure I am checking, the first operation on AttachDisk call timed out after 250 seconds. It is not created successfully at all (not sure about the reason. If it is caused by rate limit, will it error back immediately?). 

So just change here is not enough, we need to reduce the timeout defined in controller.yaml. I think 250s is too long based on normal volume operation. I propose to use 120 seconds or even less.

msau42: I think adjusting the sidecar timeout based off of the metrics we have is fine. 

But I think the waitForOp method here still needs to use the csi context's timeout instead of a hardcoded one.

jingxu97: I can change to use csi context timeout, but the poll interval will still be hardcoded for now.

jingxu97:
comments were addressed. PTAL

deploy/kubernetes/base/controller/controller.yaml
```diff
@@ -52,7 +52,7 @@ spec:
             - "--metrics-address=:22012"
             - "--leader-election"
             - "--leader-election-namespace=$(PDCSI_NAMESPACE)"
-            - "--timeout=250s"
+            - "--timeout=60s"
```
msau42, line 55: I think we should keep the value at the 99% percentile

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -681,7 +687,11 @@ func (cloud *CloudProvider) waitForZonalOp(ctx context.Context, opName string, z
 	// The v1 API can query for v1, alpha, or beta operations.
 	svc := cloud.service
 	project := cloud.project
-	return wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
+	timeout := OpTimeout
+	if deadline, ok := ctx.Deadline(); ok {
+		timeout = time.Until(deadline)
+	}
+	return wait.Poll(OpInterval, timeout, func() (bool, error) {
 		pollOp, err := svc.ZoneOperations.Get(project, zone, opName).Context(ctx).Do()
```
Jiawei0227, line 695: I think it would make sense to separate this call to another wait block. So that we can make sure the operation has been created successfully. And that can use a smaller timeout to prevent a long timeout.

deploy/kubernetes/base/controller/controller.yaml
msau42: how does a smaller timeout improve recovery if the underlying operation doesn't complete? The main thing I see is we will create an error event more frequently

jingxu97: the problem is not operation does not complete. The issue I am seeing (guessing) from logs that the first operation is not created successfully, but it does not return anything until timeout. After that, the second attach is called, and this time operation is created successfully and finished in a few second. I was suspecting it is due to rate limit and operation is dropped. But in this case, I am not sure operation creation returns an error immediately? 

Having a short timeout means we have a chance to create another operation.

jingxu97:
Some more details about the issue and proposed solution.

Sometimes csi NodeStageVolume (attach disk) timeout after 4-5 mins even through normally attach operation only takes about 4-5 seconds. The steps implemented in csi attach call includes
1. Send an attach (http) request through cloud service, and it returns an operation object. This operation has a context in which a timeout is set. The current setting for context timeout is 250s in controller yaml file.
2. Wait and check until operation is done. It periodically (3s) retrieves operation obj from cloud service and check the status until it is done. Or it will timeout after 5 minutes. Operation retrieve call also has a context in which timeout is set to 250, the same as previous step since they are using the same context.

The issue we noticed from the log, the first step to return an operation sometimes timeout after 250. THe operation is never created successfully and it times out with context setting. After that, csi issues a second NodeStageVolume call, and the operation is created and finished successfully in a few seconds. 
We are not sure yet the reason operation is failed to create without returning an error, only timeout. (might related to rate limit, but in this case should it return error instead of timeout waiting for creating operation?) To avoid waiting for too long and recover sooner, propose to adjust operation context timeout. There are actually two timeout settings involved, so let 's make a separate definition first.
1. operation timeout: each operation through cloud service has a context which has a timeout setting. The current setting is 250s in sidecar yaml file.2. watch timeout: the time to wait until operation status is done. The current setting is 5mins in hardcoded.
In order to make recovery quicker for the issue above, first propose to reduce the operation timeout. This operation timeout should not be used to represent end to end attach timeout. As the operation can be considered as nonblocking operation (does not wait until it is done), the timeout for it should be smaller. (propose 15-30s). The end-to-end attach timeout should be the watch timeout. This timeout could be minutes (5mins or less is fine for me).
If the meaning of timeout defined in csi attacher context is the end to end timeout, then we can keep timeout setting to 250s, but inside use a different timeout value for single operation context setting. This setting was set to 15 sec by default at csi attacher. But it was overwirtten by recent change https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/577

msau42:
Thanks for the detailed explanation! 

> As the operation can be considered as nonblocking operation (does not wait until it is done), the timeout for it should be smaller. 

Are you sure that this operation timeout is only for the creation of the operation and not also the duration of the operation? For example, if we set the operation timeout to something very small like under 1 second, will the operation return timeout error after 1 second when we are watching?

Jiawei0227:
I think our propose still stands regardless, right? That could fix the current issue.
To my understanding, the 250s should be the csi operation timeout. That means the sidecars will pass this to us to constraint our csi operation time.

jingxu97:
> Thanks for the detailed explanation!
> 
> > As the operation can be considered as nonblocking operation (does not wait until it is done), the timeout for it should be smaller.
> 
> Are you sure that this operation timeout is only for the creation of the operation and not also the duration of the operation? For example, if we set the operation timeout to something very small like under 1 second, will the operation return timeout error after 1 second when we are watching?

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/53e8abdbb2526655913c42452198c781956f4850/pkg/gce-cloud-provider/compute/gce-compute.go#L610

here the context.timeout is for the creation of the operation itself. To watch operation until its status is done is here https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/53e8abdbb2526655913c42452198c781956f4850/pkg/gce-cloud-provider/compute/gce-compute.go#L680:29

Again, the context timeout will be used as the Operation.get() itself.

jingxu97:
> csi operation timeout. That means the sidecars will pass this to us to constraint our csi operation time.

csi operation. you mean the end to end csi call, right? Yes, the total timeout is 250s, then we have to redefine the context passed in from csi attacher with much shorter timeout for each individual gce.service operation.

Jiawei0227:
> > csi operation timeout. That means the sidecars will pass this to us to constraint our csi operation time.
> 
> csi operation. you mean the end to end csi call, right? Yes, the total timeout is 250s, then we have to redefine the context passed in from csi attacher with much shorter timeout for each individual gce.service operation.

I am not saying the end2end call, I mean the CSI operation call timeout is 250s. So I think the context passed in from csi attacher should just be 250s.  In my opinion, for each user operation like mount/attach/etc.. normally it should just be 1 or 2 calls to the driver, right?

Jiawei0227:
I dont know why we use 250s tho. I think @saikat-royc might have an idea why 250s is needed here.

jingxu97:
> I dont know why we use 250s tho. I think @saikat-royc might have an idea why 250s is needed here.

that's my reasoning, the reason set to 250s because we use it as kind of end to end timeout for CSI calls. Here e2e I mean the beginning from the csi request until it returns. But the problem is now we also use this timeout as gce.service operation timeout, which is not correct because that operation timeout should be much smaller. We hardcoded 5mins to watch until operation is done. This 5min can be considered as whole csi operation timeout. There are other settings of watch timeout for other operations.

Jiawei0227:
Oh yeah you are right. 5min should not be used here. Instead we should use the context time to indicate the deadline. Good finding!
I dont think this has anything to do with separating the get operation call to a separate wait loop, does it? Or do I misunderstand somehing

jingxu97:
> Oh yeah you are right. 5min should not be used here. Instead we should use the context time to indicate the deadline. Good finding!
> I dont think this has anything to do with the proposal of separating the get operation call to a separate wait loop, does it? Or do I misunderstand somehing

you mean the create operation call in a separate wait loop? since the get operation call is already in a separate loop now in waitForZoneOp()..

Jiawei0227:
I think if we put 
```golang
pollOp, err := svc.ZoneOperations.Get(project, zone, opName).Context(ctx).Do()
```
this call into a separate wait loop with a smaller timeout like 30s, then we can return err early if for any reason the operation creation get dropped. So that the external-attacher can issue an attach again.

I think if we put create operation call in a wait loop util it makes sure the operation is created in gcp is also okay. I am fine with both implemention.

saikat-royc:
> this call into a separate wait loop with a smaller timeout like 30s, then we can return err early if for any reason the operation creation get dropped. So that the external-attacher can issue an attach again.

Agree.
Actually all the calls which return an operation, and the op.get() calls should have a short timeout. And the overall wait for the op to reach a done status (success/fail), should honor the csi timeout provided by the sidecars.

e.g.
wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) -> this i think should be the same as the csi timeout - some delta.
things like this 
op, err := cloud.service.Instances.AttachDisk(cloud.project, instanceZone, instanceName, attachedDiskV1).Context(ctx).Do(),
and
pollOp, err := svc.ZoneOperations.Get(project, zone, opName).Context(ctx).Do() should have a fixed short timeout in-order to react faster.

mattcary:
is this pr still relevant?

PR #666, opened 2020-11-11
Fix check volume is mounted logic during NodePublishVolume and

jingxu97:
NodeStageVolume

The logic for checking volume is mounted or not during NodePublishVolume
has issues. It might not handle failures well.

This issue was discovered by https://github.com/kubernetes/kubernetes/pull/88759 which cause windows tests fail.

**What type of PR is this?**
>
> /kind api-change
/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -69,6 +69,21 @@ func getDefaultFsType() string {
 		return defaultLinuxFsType
 	}
 }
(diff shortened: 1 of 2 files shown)


mattcary:
/lgtm

Thanks for wrapping IsLikelyNotMountPoint. The inverted logic in that function is mind-bending.

PR #679, opened 2020-12-01
disable gke deployment for clusters with installation by default if GKE deployment is not requested

verult:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

**What this PR does / why we need it**: 1.18 driver tests are failing when an overlay is used in a GKE deployment because the GKE deployment is enabled by default. This PR explicitly disables it.

Fixes #

```release-note
NONE
```

/assign @saikat-royc

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -182,9 +182,23 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 	}
 
 	cmd = exec.Command("gcloud", cmdParams...)
-	err = runCommand("Staring E2E Cluster on GKE", cmd)
+	err = runCommand("Starting E2E Cluster on GKE", cmd)
 	if err != nil {
 		return fmt.Errorf("failed to bring up kubernetes e2e cluster on gke: %v", err)
+	}
+
+	// Because gcloud cannot disable addons on cluster create, the deployment has
+	// to be disabled on update.
+	clusterVersion := mustGetKubeClusterVersion()
+	if !useManagedDriver && isGKEDeploymentInstalledByDefault(clusterVersion) {
+		cmd = exec.Command(
+			"gcloud", "beta", "container", "clusters", "update",
+			*gkeTestClusterName, locationArg, locationVal, "--quiet",
+			"--update-addons", "GcePersistentDiskCsiDriver=DISABLED")
+		err = runCommand("Updating E2E Cluster on GKE to disable driver deployment", cmd)
+		if err != nil {
+			return fmt.Errorf("failed to update kubernetes e2e cluster on gke: %v", err)
+		}
 	}
 
 	return nil
@@ -378,4 +392,11 @@ func getKubeClient() (kubernetes.Interface, error) {
 		return nil, fmt.Errorf("failed to create client: %v", err)
 	}
 	return kubeClient, nil
+}
+
+func isGKEDeploymentInstalledByDefault(clusterVersion string) bool {
+	cv := mustParseVersion(clusterVersion)
+	return cv.atLeast(mustParseVersion("1.18.10-gke.2101")) &&
+		cv.lessThan(mustParseVersion("1.19")) ||
+		cv.atLeast(mustParseVersion("1.19.3-gke.2100"))
 }
--- a/test/k8s-integration/version.go
+++ b/test/k8s-integration/version.go
@@ -105,6 +105,7 @@ func parseVersion(vs string) (*version, error) {
 	return &v, nil
 }
 
+// mustParseVersion parses a GKE cluster version.
 func mustParseVersion(version string) *version {
 	v, err := parseVersion(version)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


test/k8s-integration/cluster.go
```diff
@@ -379,3 +394,10 @@ func getKubeClient() (kubernetes.Interface, error) {
 	}
 	return kubeClient, nil
 }
+
+func isGKEDeploymentInstalledByDefault(clusterVersion string) bool {
```
msau42, line 398: Is it possible to query the cluster to see if it's already enabled, instead of encoding versions?

verult: We could parse the output of `gcloud beta container clusters describe`. It might be good to test these version bounds for defaulting tho, and it's also a teeny bit less work

verult:
Local Prow run passing

test/k8s-integration/cluster.go
```diff
@@ -182,11 +183,25 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
...
 	cmd = exec.Command("gcloud", cmdParams...)
-	err = runCommand("Staring E2E Cluster on GKE", cmd)
+	err = runCommand("Starting E2E Cluster on GKE", cmd)
 	if err != nil {
 		return fmt.Errorf("failed to bring up kubernetes e2e cluster on gke: %v", err)
 	}
 
+	// Because gcloud cannot disable addons on cluster create, the deployment has
+	// to be disabled on update.
+	clusterVersion := mustGetKubeClusterVersion()
```
saikat-royc, line 193: This update should be performed if we detect a pdcsi-node daemonset in kube-system namespace.(as this indirectly indicates that the component is enabled)

test/k8s-integration/cluster.go
saikat-royc: This function should check if we detect a pdcsi-node daemonset in kube-system namespace.(as this indirectly indicates that the component is enabled)

verult: I think it's still good to use version bounds so that defaulting is only done in the versions we expect, but not a huge benefit since other test suites already cover defaulting testing. No strong preference either way

Calling kubectl to get DaemonSet info also requires parsing, in which case I would prefer getting the `addonsConfig` from `clusters describe` which is the definitive state of whether GKE deployment should be running.

verult:
@msau42 @saikat-royc Created #680 to track refactor to use GKE go client. Updating regional clusters takes way too long through gcloud.

test/k8s-integration/cluster.go
```diff
@@ -379,3 +394,10 @@ func getKubeClient() (kubernetes.Interface, error) {
 	}
 	return kubeClient, nil
 }
+
+func isGKEDeploymentInstalledByDefault(clusterVersion string) bool {
+	cv := apimachineryversion.MustParseSemantic(clusterVersion)
```
saikat-royc, line 399: Since this is GKE version, should we use [this](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/423ae53cb07dea2ad3a9659bc92df8a303a5503b/vendor/k8s.io/apimachinery/pkg/util/version/version.go#L116)?

test/k8s-integration/cluster.go
saikat-royc: Agree that checking addonsConfig is the right way to do it.
The only issue with checking k8s version, is that we would need to revisit and update the logic again, if there is a on-by-default is enabled for 1.17 clusters for e.g and also then it may get complicated with all the patch version comparision.
For immediate relief to the test failures I am ok with the k8s version check, but eventually we should check addonsConfig.

WDYT @msau42 ?

msau42: Long term I agree checking addons config is the right way. This is fine for now

verult: Ah right yeah all the backports. Will add a TODO here.

test/k8s-integration/cluster.go
verult: That's the one used here I believe

test/k8s-integration/cluster.go
verult: Actually once the GKE go client refactor happens, we could just always explicitly enable/disable on cluster create. Defaulting logic is already tested elsewhere so not having coverage here is OK.

test/k8s-integration/cluster.go
saikat-royc: Check this https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/main.go#L530 generateGKETestSkip()

verult: ah gotcha, updated

saikat-royc:
tagging @mattcary to be in the loop

PR #682, opened 2020-12-08
Remove Probe logging

Jiawei0227:
This is especially useful when we have a livenessprobe sidecar
The log can be spamed with all the probe logs if we record them

**What type of PR is this?**
> /kind cleanup

This PR removes probe logging for pd driver. Since it can provide huge noise to the entire logging if we enable livenessprobe

Fixes #

```release-note
Remove probe logging to reduce noise
```

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -27,6 +27,8 @@ import (
 	"k8s.io/klog"
 )
 
+var ProbeCSIFullMethod = "/csi.v1.Identity/Probe"
+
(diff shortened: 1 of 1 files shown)


Jiawei0227:
It would be so much easier to debug and log searching with this PR merged.

mattcary:
This seems reasonable to me, but maybe I'm missing something.

@msau42 wdyt?

Jiawei0227:
I would strongly suggest we merge this PR to save the log space from GKE point of view. Otherwise the pd driver log will be spammed with the probe logging

PR #684, opened 2020-12-16
Round up pdcsi driver size in CreateVolume

Jiawei0227:
**What type of PR is this?**
/kind bug

Currently if we create a PVC of size 1.5Gi, the pdcsi driver will convert it to 1Gi but the provisioner will say it is 2Gi[1]. And both of them are not accurate. From the CSI spec, we know that we should provide size larger than the requested size. So if user ask for 1.5Gi, we should give them at least 1.5Gi. 

This PR changes the RoundDown behavior of CreateVolume to RoundUp. This is to align with the CreateVolume CSI spec that when the required_bytes is set, the driver should provision something bigger.

[1] https://github.com/kubernetes-csi/external-provisioner/issues/199

Fixes #

```release-note
Fix a bug that CreateVolume should round up the request_bytes.
```

/cc @msau42 
/cc @saikat-royc 
/assign @mattcary

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -49,9 +49,17 @@ const (
 	regionalDeviceNameSuffix = "_regional"
 )
 
-func BytesToGb(bytes int64) int64 {
+func BytesToGbRoundDown(bytes int64) int64 {
 	// TODO: Throw an error when div to 0
 	return bytes / (1024 * 1024 * 1024)
+}
+
+func BytesToGbRoundUp(bytes int64) int64 {
+	re := bytes / (1024 * 1024 * 1024)
+	if (bytes % (1024 * 1024 * 1024)) != 0 {
+		re++
+	}
+	return re
 }
 
 func GbToBytes(Gb int64) int64 {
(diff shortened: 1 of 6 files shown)


test/k8s-integration/test-config.yaml
```diff
@@ -0,0 +1,30 @@
+ShortName: pdtest
```
mattcary, line 1: You probably don't mean to check in this file?

mattcary (commented):
I don't know if backporting makes sense; I'm not sure this will justify a backport in GKE as +/-1G isn't a big deal.

Could you create an issue for this bug? Then we could at least have a public discussion about backporting.

mattcary:
/approved
/lgtm

Thanks!

Can you mention the issue number in this PR when you create it (or mention this PR in the issue, whatever, just so we can keep track)

Jiawei0227:
> Can you mention the issue number in this PR when you create it (or mention this PR in the issue, whatever, just so we can keep track)

Done. The bug number is 176099317. Can you help to approve this PR? Thanks a lot!

mattcary:
/approve

Sorry, I got the wrong approve robot command.

PR #685, opened 2020-12-16
Skip volume expansion test for node skew tests

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
/kind failing-test
> /kind feature
> /kind flake

volume expansion feature is beta in k8s 1.16. for the node skew tests on regular channel, we ended up with a master in 1.16 and node in 1.15. This caused the kubelet in the node to skip the NodeExpandVolume CSI calls to the driver, since the codepath is feature gated. [ExpandCSIVolumes](https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L97)

Fixes #

```release-note
NONE
```


msau42 (commented):
/lgtm

PR #688, opened 2020-12-22
Use v0.7.0 image for release-0.7

### LINKED ISSUE #687: XFS tests failing in 0.7 testgrid
https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver#Kubernetes%20Master%20Driver%20Release%200.7

It seems to be because the v0.6.2 driver image used in this release does not contain the xfs tools.

That's a little weird, because it appears the v0.6.2 release was cut after the xfs tools were added to the docker image. Inspecting the v0.6.2 image shows the tools are definitely not included, so there you have it.

I think the fix is just to use v0.7.0 in release-0.7, which I would have thought should have been done from the start?

/assign @mattcary

mattcary:
/kind bug

Fixes #687

/assign @msau42 

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/overlays/stable/kustomization.yaml
+++ b/deploy/kubernetes/overlays/stable/kustomization.yaml
@@ -7,7 +7,7 @@ images:
   # Don't change stable image without changing pdImagePlaceholder in
   # test/k8s-integration/main.go
   newName: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
-  newTag: "v0.6.2-gke.0"
+  newTag: "v0.7.0-gke.0"
 - name: gke.gcr.io/csi-provisioner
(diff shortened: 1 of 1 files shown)


mattcary:
/assign @verult 

Michelle is out for the holidays. Cheng, do you have any context on release-0.7?

verult:
/lgtm
/approve

This should've been updated as part of the release process so yeah definitely looks like a bug to me, although if no one noticed this bug, I'm wondering how important it really is to keep per-release deployment specs up-to-date.

mattcary:
Oh, I see. OWNERS on this branch is old and I'm not in it. Michelle or Saad, can you approve?

/assign @msau42 
/cc @saad-ali

Issue #469, opened 2020-02-07
Option to tag PD with cluster name


msau42:
Since the driver is usually deployed per cluster, the cluster name could be passed in as a driver arg. This can help associate PDs to clusters. Instead of being specific to cluster name, it could also just take in a set of global tags. This is what ebs driver does.

/kind feature

mattcary:
Is this effectively a duplicate of #340?

msau42:
#340 is at the StorageClass level. This is at the cluster-level and could be automatically inferred by the provider that's deploying the driver.

mattcary:
Got it, so it eg could be done at the same time that we set the PV info, https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/common/parameters.go#L86

msau42:
Yeah it can be set when you create the PV, but the actual config could be set when you start the driver (as a driver arg)

loganmzz:
I don't see interest in pushing explicitly cluster name into GCP-managed resources since cluster labels are already applied as describe in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/340#issuecomment-543221858.

Just add a `gke-cluster-id` (or similar) label(s) to your cluster. However, being able to distinct resources coming from a single cluster is still required. For the moment, I rely on `description` field which contains JSON structure. But as it's stored as raw string, filtering against is fragile.

Output example:

```
$ gcloud compute disks describe --zone gke-my-k8s-cluster-1bcec-pvc-a80c2aa5-f0c9-4a64-b3c4-a3930f037528
...
description: '{"kubernetes.io/created-for/pv/name":"pvc-a80c2aa5-f0c9-4a64-b3c4-a3930f037528","kubernetes.io/created-for/pvc/name":"my-k8s-pvc","kubernetes.io/created-for/pvc/namespace":"my-k8s-namespace"}'
...

msau42:
There is this change here which is using Disk labels, and we'll have to do a followup to set the pvc/pv tags as a Label as well: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/566

As for cluster id as a label, it's useful for non-gke users.

PR #693, opened 2021-01-11
Allow to label PD disk with k8s cluster ID

tsmetana:
/kind feature

Inspired by the similar feature of [AWS EBS CSI driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/cmd/main.go#L37): allow to optionally label the PD disk with cluster ID to denote the disk ownership. This allows e.g. better automated resource cleanup on cluster teardown.

The label was meant to be named `kubernetes.io/cluster/<cluster ID>` (again, taken from the [EBS CSI](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/constants.go#L61) driver) however GCE API does not allow '/' or '.' characters in the label, so I swapped them for underscores. I can change it to anything else, I'm more interested in comments on the general approach.

```release-note
A new `k8s-tag-cluster-id` command line option has been added. If specified, the resulting PD disk will be labelled with "kubernetes_io_cluster_<cluster ID>": "owned".
```


tsmetana:
@jsafrane: FYI...

mattcary:
Thanks Thomas. I haven't had a chance to look at this yet. Have you seen https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/566? The author We were basically happy with that PR, but the author seems to have dropped out. We want to get that in soon though.

If we had the ability of that CL (to label the PD disk via storage class parameters), would that fit your use case? That might be more flexible than a flag with a hardcoded label.

msau42:
> If we had the ability of that CL (to label the PD disk via storage class parameters), would that fit your use case? That might be more flexible than a flag with a hardcoded label.

I think the main downside is then people can't use the default storageclass, unless we had a way to generate default storageclasses and inject the cluster name into it (k8s doesn't have a first class concept of cluster name)

mattcary:
> I think the main downside is then people can't use the default storageclass, unless we had a way to generate default storageclasses and inject the cluster name into it (k8s doesn't have a first class concept of cluster name)

Since this proposal involves a command-line flag for the driver, anyone who can install the driver can also change the default storageclass to have whatever label they want.

Since this label isn't going to match the AWS EBS label anyway due to the slash-vs-underscore issue, there probably needs to be flexibility for the operator to tweak the label to what is most convenient for their tools. Doing that via the storageclass parameter seems nicer than via a driver command-line argument.

My $0.02 at least.

msau42:
previous discussion: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/469

tsmetana:
> > I think the main downside is then people can't use the default storageclass, unless we had a way to generate default storageclasses and inject the cluster name into it (k8s doesn't have a first class concept of cluster name)
> 
> Since this proposal involves a command-line flag for the driver, anyone who can install the driver can also change the default storageclass to have whatever label they want.

The users able to install drivers and those able to create storage classes might be different. The aim is to be able to create an environment where the clusters get automatically set-up (click in a browser) and torn-down (another click) and don't leave any leftovers while still not require any additional responsibilities from their admins (not forget to label StorageClasses).

mattcary:
> The users able to install drivers and those able to create storage classes might be different. The aim is to be able to create an environment where the clusters get automatically set-up (click in a browser) and torn-down (another click) and don't leave any leftovers while still not require any additional responsibilities from their admins (not forget to label StorageClasses).

Okay, that makes sense. I think fixing the label format here will be too restrictive. I'd support a change that was something like --additional-pd-label, eg --additional-pd-label=k8s-cluster-id-112233. The flag could even be given multiple times, for multiple labels.

jsafrane:
AWS has `--extra-tags=foo=bar,baz=qux`, maybe `--extra-labels`? Not sure if `pd` should be in the parameter - should it affect also snapshots?

mattcary:
> AWS has `--extra-tags=foo=bar,baz=qux`, maybe `--extra-labels`? Not sure if `pd` should be in the parameter - should it affect also snapshots?

+1 to both the name as well as labeling snapshots in addition to disks.

tsmetana:
Changed to allow generic disk labelling. PTAL.

go.sum
```diff
@@ -476,9 +476,6 @@ github.com/kr/pty v1.1.3/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/pty v1.1.5/go.mod h1:9r2w37qlBe7rQ6e1fg1S/9xpWHSnaqNdHD3WcMdbPDA=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
-github.com/kubernetes-csi/csi-proxy v0.2.2 h1:LqablYFEGw7FYBjwoh5TeXFzlcx8C+YQjKfGy6fFWJs=
```
mattcary, line 479: I don't think we want to remove the csi-proxy? It's used by windows. At any rate we should clean this up in a separate PR.

@jingxu97 We seem to have both 0.2.1 and 0.2.2 of the proxy here, not sure what's going on.

mattcary (commented):
Just a couple of nits.

Thanks!

pkg/common/parameters_test.go
```diff
@@ -25,79 +25,93 @@ func TestExtractAndDefaultParameters(t *testing.T) {
...
 		{
 			name:       "tags",
 			parameters: map[string]string{ParameterKeyPVCName: "testPVCName", ParameterKeyPVCNamespace: "testPVCNamespace", ParameterKeyPVName: "testPVName"},
+			labels:     map[string]string{},
 			expectParams: DiskParameters{
 				DiskType:             "pd-standard",
 				ReplicationType:      "none",
 				DiskEncryptionKMSKey: "",
 				Tags:                 map[string]string{tagKeyCreatedForClaimName: "testPVCName", tagKeyCreatedForClaimNamespace: "testPVCNamespace", tagKeyCreatedForVolumeName: "testPVName", tagKeyCreatedBy: "testDriver"},
+				Labels:               make(map[string]string),
```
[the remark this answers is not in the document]
tsmetana: I changed the code since `ExtractAndDefaultParameter` now takes more parameters and to have the parameteres structure complete. There's nothing much to test with the labels though -- they're only being copied.

go.sum
mattcary: At any rate this PR should not be changing go.mod (sorry I wasn't clear on that).

pkg/common/parameters_test.go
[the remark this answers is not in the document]
mattcary: If they're only we being copied we should test that---these also serve as regression tests.

go.sum
tsmetana: OK. I'll fix that.

tsmetana: The above is a result of `go mod tidy`. I use `k8s.io/component-base/cli/flag` for the CLI parsing and it was not vendored yet so I actually have to update go.mod for this patch. I think the removal of the (older) csi-proxy v0.2.1 is correct and the v0.2.2 is listed twice in go.sum.

pkg/common/parameters_test.go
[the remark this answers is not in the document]
tsmetana: I added the test in the latest commit.

go.sum
mattcary: Ok, fair enough. I'm not going to argue with go mod tidy.

mattcary (commented):
/lgtm
/approve

Thanks for working through this!

PR #698, opened 2021-01-22
Staging RC overlay: Bump all sidecar versions to latest

verult:
**What type of PR is this?**
/kind cleanup

Fixes #

Snapshotter is updated to 3.0.3 because 4.0.0 is 1.20+

Node sidecar flag changes are ignored because:
* Node sidecar liveness probes should be changed to exec-based so they don't consume ports on user nodes.
* Metrics manager in livenessprobe sidecar is only measuring probe CSI calls it seems.

```release-note
NONE
```

/assign @saikat-royc @mattcary 
/cc @msau42


mattcary:
We have a v3.0.3 version of the snapshotter, don't we want to update to use that?

The v4 is for 1.20 IIUC.

verult:
Ah thanks, updated

PR #712, opened 2021-02-17
Add a mode to run e2e tests using kubetest2.

TeweiLuo:
Engprod plans to end kubetest support. This diff adds a --use-kubetest2 flag to the
integration test driver.

Next step:
1. Add kubetest2 installation to test scripts.
2. Add a canary test config to validate kubetest2.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Migrate to the latest test-infra supported test runner. Immediate benefits include reducing ssh-related flakiness in kubetest.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -67,6 +67,8 @@ var (
 	// Test flags
 	migrationTest = flag.Bool("migration-test", false, "sets the flag on the e2e binary signalling migration")
 	testFocus     = flag.String("test-focus", "", "test focus for Kubernetes e2e")
+
+	useKubeTest2 = flag.Bool("use-kubetest2", false, "use kubetest2 to run e2e tests")
 )
 
 const (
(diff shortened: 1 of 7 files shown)


test/k8s-integration/main.go
```diff
@@ -436,10 +439,20 @@ func handle() error {
 
 	switch testParams.deploymentStrategy {
 	case "gke":
-		testParams.cloudProviderArgs, err = getGKEKubeTestArgs(*gceZone, *gceRegion, testParams.imageType)
-		if err != nil {
+		// Build params for both kubetest and kubetest2 for debugging.
+		testParams.cloudProviderArgs, err = getGKEKubeTestArgs(*gceZone, *gceRegion, testParams.imageType, false /* useKubeTest2 */)
```
mattcary, line 443: If we don't even log an error when, eg kubetest2=false and kubetest2 arg building returns an error, why are we building both sets of args?

It may simplify the logic considerably to just plumb through one set of args.

test/k8s-integration/main.go
```diff
@@ -436,10 +439,20 @@ func handle() error {
...
+		if err != nil && !*useKubeTest2 {
 			return fmt.Errorf("failed to build GKE kubetest args: %v", err)
 		}
+		testParams.cloudProviderArgsK2, err = getGKEKubeTestArgs(*gceZone, *gceRegion, testParams.imageType, true /* useKubeTest2 */)
+		if err != nil && *useKubeTest2 {
+			return fmt.Errorf("failed to build GKE kubetest2 args: %v", err)
+		}
+	case "gce":
+		testParams.cloudProviderArgsK2 = []string{
+			"--legacy-mode",
```
mattcary, line 453: could you add a comment for what legacy-mode does? I always forget, and since now kubetest1 can be considered a legacy mode it would be good to not be ambiguous.

TeweiLuo: Good call. The naming isn't very helpful.

test/k8s-integration/main.go
TeweiLuo: Agreed - changed to only generate one set of args.

PR #718, opened 2021-02-24
Add gce disk labels support via create volume parameters

### LINKED ISSUE #340: Support for disk labels parameter

mattcary:
Labels are validated to match gce requirements.

Cherry-picked from #566

Fixes #340

```release-note
Disk labels support via CreateVolume (and hence StorageClass) parameters
```


test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -292,6 +292,46 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			"key2": "value2",
+			// The label below is added as an --extra-label driver command line argument.
+			testutils.DiskLabelKey: testutils.DiskLabelValue,
+		}))
+		Expect(cloudDisk.Name).To(Equal(volName))
+
+		defer func() {
+			// Delete Disk
+			client.DeleteVolume(volID)
+			Expect(err).To(BeNil(), "DeleteVolume failed")
```
Jiawei0227, line 327: Which err?

test/k8s-integration/config/sc-standard.yaml
```diff
@@ -5,4 +5,5 @@ metadata:
 provisioner: pd.csi.storage.gke.io
 parameters:
   type: pd-standard
+  labels: key1=value1,key2=value2
```
Jiawei0227, line 8: Maybe add a comment here saying this is for tests?

test/e2e/tests/single_zone_e2e_test.go
mattcary: Good catch, thanks.

test/k8s-integration/config/sc-standard.yaml
mattcary: Good idea, done.

pkg/common/parameters_test.go
```diff
@@ -37,38 +37,41 @@ func TestExtractAndDefaultParameters(t *testing.T) {
...
 				ReplicationType:      "none",
 				DiskEncryptionKMSKey: "",
-				Tags:                 make(map[string]string),
-				Labels:               make(map[string]string),
+				Tags:                 map[string]string{},
+				Labels:               map[string]string{},
 			},
 		},
 		{
 			name:       "random keys",
```
saikat-royc, line 57: is this tc name correct?

pkg/common/parameters_test.go
```diff
@@ -37,38 +37,41 @@ func TestExtractAndDefaultParameters(t *testing.T) {
...
-			parameters: map[string]string{ParameterKeyType: "", "foo": "", ParameterKeyDiskEncryptionKmsKey: ""},
+			parameters: map[string]string{ParameterKeyType: "", "foo": "", ParameterKeyDiskEncryptionKmsKey: "", ParameterKeyLabels: ""},
 			labels:     map[string]string{},
 			expectErr:  true,
 		},
 		{
 			name:       "real values",
```
saikat-royc, line 63: maybe we should elaborate the tc name, as labels picked from params

pkg/common/utils.go
```diff
@@ -156,3 +157,60 @@ func CreateNodeID(project, zone, name string) string {
...
+func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
+	const labelsDelimiter = ","
+	const labelsKeyValueDelimiter = "="
+
+	labelsMap := make(map[string]string)
+	if labels == "" {
+		return labelsMap, nil
+	}
+
+	regexKey, _ := regexp.Compile(`^\p{Ll}[\p{Ll}0-9_-]{0,62}$`)
```
saikat-royc, line 172: Maybe we can put a pointer to the label format https://cloud.google.com/compute/docs/labeling-resources#label_format

examples/kubernetes/demo-labeled-sc.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: csi-gce-pd
+provisioner: pd.csi.storage.gke.io
+parameters:
```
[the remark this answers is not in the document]
mattcary: done

pkg/common/parameters_test.go
mattcary: done

pkg/common/utils.go
mattcary: done

PR #719, opened 2021-02-25
Emit GKE PDCSI component version metric

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR enables the gce-pd-driver container to emit 'component_version' metric. This metric will be emitted only by the controller driver gce-pd-driver container, when the env `GKE_PDCSI_VERSION` is setup in the driver manifest.
In GKE the env variable would be populated from metadata annotations (which is automatically populated for GKE pod). This metric will enable GroupBy queries on PDCSI component version
```
- name: GKE_PDCSI_VERSION
      valueFrom:
        fieldRef:
          # Component version annotation is populated by renderer
          fieldPath: metadata.annotations['components.gke.io/component-version']
```

Fixes #

Tested locally as follows:
1. deploy the driver (with a --http-endpoint=:22015 and env: -name:GKE_PDCSI_VERSION value: <some random value>)
sample patch:
 ```
@@ -138,9 +138,12 @@ spec:
           args:
             - "--v=5"
             - "--endpoint=unix:/csi/csi.sock"
+            - "--http-endpoint=:22015"
           env:
             - name: GOOGLE_APPLICATION_CREDENTIALS
               value: "/etc/cloud-sa/cloud-sa.json"
+            - name: GKE_PDCSI_VERSION
+              value: "0.9.4"
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
```

2. ssh into the node where the controller is running
3. curl to the http endpoint as follows and check the metric emitted:
```
saikatroyc@e2e-test-saikatroyc-minion-group-s7lb ~ $ curl http://localhost:22015/metrics
# HELP component_version [ALPHA] Metric to expose the version of the PDCSI GKE component.
# TYPE component_version gauge
component_version{component_version="0.9.4"} 1
```

```release-note
Emit GKE component version metric 
```


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"net/http"
+	"os"
+
+	"github.com/prometheus/client_golang/prometheus"
```
mattcary, line 23: So the sidecars use csi-lib-utils' MetricsManager, although I don't think that's appropriate here.

csi-lib-utils itself uses k8s.io/component-base/metrics/.

I have no idea if there's an advantage or good reason to using that rather than prometheus directly.

@verult ?

verult: I'm actually not sure. @Jiawei0227 do you happen to know?

mattcary: Saikat and I discussed offline and we think k8s.io/component-base/metrics is the right thing to use. ???

Jiawei0227: We should use `k8s.io/component-base/metrics` as it is the moving standard kubernetes metrics wrapper pkg

Jiawei0227: We should not use MetricsManager in csi-lib-utils I think. That might not be a good fit

Jiawei0227:
Can I ask what is benefit of adding this version through metrics? I think we have other ways to get the component version number through GKE

Jiawei0227:
I remember in GKE we have add_on_* something and that will contain the component versions.

pkg/metrics/metrics.go
saikat-royc: @Jiawei0227  I am planning to use k8s.io/component-base/metrics directly, not the csi lib utils.

mattcary:
> I remember in GKE we have add_on_* something and that will contain the component versions.

GKE is moving to a different component release pipeline which changes how versions are exported.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/klog"
+)
+
+const (
+	// envPDCSIVersion is an environment variable set in the PDCSI controller manifest
+	// with the current version of the GKE component.
+	envPDCSIVersion = "PDCSI_VERSION"
```
msau42, line 31: I wonder if this should be more gke specific like "gke component version". pdcsi version is different.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/kubernetes-csi/csi-test/v3 v3.0.0
 	github.com/onsi/ginkgo v1.11.0
 	github.com/onsi/gomega v1.7.1
+	github.com/prometheus/client_golang v1.0.0
```
msau42, line 16: I'm surprised so many packages were changed but there were no changes to go.sum?

pkg/metrics/metrics.go
saikat-royc: Ack.

go.mod
saikat-royc: ah maybe I missed adding it to the commit!

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+func (mm *metricsManager) registerToServer(s Server, metricsPath string) {
+	s.Handle(metricsPath, metrics.HandlerFor(
+		mm.GetRegistry(),
+		metrics.HandlerOpts{
+			ErrorHandling: metrics.ContinueOnError}))
+}
+
+// InitializeHttpHandler sets up a server and creates a handler for metrics.
+func (mm *metricsManager) InitializeHttpHandler(address, path string) {
+	mux := http.NewServeMux()
```
verult, line 98: Just a note that a mux is only necessary if there are multiple handlers, so here either this or the previous version of this code would work

saikat-royc: Keeping mux for future proofing?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+const (
+	// envGKEPDCSIVersion is an environment variable set in the PDCSI controller manifest
+	// with the current version of the GKE component.
+	envGKEPDCSIVersion = "GKE_PDCSI_VERSION"
+)
+
+var (
+	// This metric is exposed only from the controller driver component when PDCSI_VERSION env variable is set.
```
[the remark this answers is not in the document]
saikat-royc: Done

PR #720, opened 2021-03-12
Fix cluster list parsing for latest gcloud version

verult:
**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**: Prow tests are failing with the latest kubekins image because in the latest gcloud version, `gcloud container clusters list --filter name=<cluster_name>` prints one line of warning instead of returning empty, which makes the k8s-integration binary to proceed to cluster deletion, and fail when the cluster doesn't actually exist.

This PR is a short term solution. Long term solution would be moving to the cloud SDK (#680).

```release-note
None
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -145,8 +145,9 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


mattcary:
/lgtm

This hit the filestore driver, I should have seen the problem...

PR #722, opened 2021-03-14
Set major & minor version in docker version defs

### LINKED ISSUE #721: GCE cluster up is failing for k8s-integration tests on k/k master.
The symptom is that the e2e-up called from the k8s-integration test on k/k master will hang on waiting for nodes to be ready. Node kubelet logs contain an error like the following. 

```
cannot create certificate signing request: certificatesigningrequests.certificates.k8s.io is forbidden: User "kubelet" cannot create resource "certificatesigningrequests" in API group "certificates.k8s.io" at the cluster scope.
```

And there are additional errors in the kube-addon-manager.log as well.

It appears that the 888.888 git version used in test/k8s-integration/.dockerized-kube-version-defs causes something critical to get removed or not activated, although I haven't been able to figure out what.

This can be temporarily fixed by setting the version there to 1.21+. It would be better to automatically infer the kube version, maybe there is some trickery to be done, or maybe it would be easiest to clone from git rather than just downloading the archive.

mattcary:
This should get kube-up used by the integration tests temporarily working. It appears the old fake 888.888 version was causing something---I haven't figured out what yet---to not get installed or activated, and node RBAC couldn't get installed.  See the issue for further details.

/kind bug

Fixes #721

```release-note
None
```

### DIFF
--- a/test/k8s-integration/.dockerized-kube-version-defs
+++ b/test/k8s-integration/.dockerized-kube-version-defs
@@ -1,5 +1,5 @@
 KUBE_GIT_COMMIT='somefakecommit'
 KUBE_GIT_TREE_STATE='dirty'
 KUBE_GIT_VERSION='v888.888.888-fake-testing-master.version'
-KUBE_GIT_MAJOR='888'
-KUBE_GIT_MINOR='888+'
+KUBE_GIT_MAJOR='1'
+KUBE_GIT_MINOR='21+'


mattcary:
See #723 for an alternative fix. This one can be pushed quickly to stop the current bleeding.

mattcary:
Obsolete by #723 
/close

PR #723, opened 2021-03-14
Make a clone of k8s master instead of using archive

### LINKED ISSUE #721: GCE cluster up is failing for k8s-integration tests on k/k master.

mattcary:
/kind bug

Fixes #721 

This is a more complete change compare with #722.

See the comment in the change for the details of how the git clone is done for the master branch.

Pulling a specific version is unchanged (the k/k release archive is still used).

```release-note

```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -206,57 +206,46 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
 
 func downloadKubernetesSource(pkgDir, k8sIoDir, kubeVersion string) error {
 	k8sDir := filepath.Join(k8sIoDir, "kubernetes")
-	/*
-		// TODO: Download a fresh copy every time until mutate manifests hardcoding existing image is solved.
-		if _, err := os.Stat(k8sDir); !os.IsNotExist(err) {
-			klog.Infof("Staging Kubernetes already found at %s, skipping download", k8sDir)
-			return nil
-		}
-	*/
+	klog.V(4).Infof("Downloading Kubernetes source")
 
-	klog.V(4).Infof("Staging Kubernetes folder not found, downloading now")
-
-	err := os.MkdirAll(k8sIoDir, 0777)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


test/k8s-integration/cluster.go
```diff
@@ -206,57 +206,46 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, imageType string, use
...
 	}
 
-	kubeTarDir := filepath.Join(k8sIoDir, fmt.Sprintf("kubernetes-%s.tar.gz", kubeVersion))
-
-	var vKubeVersion string
 	if kubeVersion == "master" {
-		vKubeVersion = kubeVersion
-		// A hack to be able to build Kubernetes in this nested place
-		// KUBE_GIT_VERSION_FILE set to file to load kube version from
-		err = os.Setenv("KUBE_GIT_VERSION_FILE", filepath.Join(pkgDir, "test", "k8s-integration", ".dockerized-kube-version-defs"))
```
saikat-royc, line 231: Should we also remove dockerized-kube-version-defs file if its not needed now?

saikat-royc:
the change looks good to me, with some minor comment

test/k8s-integration/cluster.go
mattcary: I did remove it :-)

saikat-royc: oops missed it

PR #724, opened 2021-03-15
Add cloudbuild config to build gcp-compute-persistent-disk-csi-driver image

cpanato:
**What type of PR is this?**

/kind feature

As part of the effort described in this [issue](https://github.com/kubernetes/k8s.io/issues/1525) we are deprecating the `gcr.io/gke-release` and the image `gcr.io/gke-release/gcp-compute-persistent-disk-csi-driver` does not exist in other registries. So we need to build that and push to a new registry.

We will use `gcr.io/k8s-staging-cloud-provider-gcp` to push the images and This PR add the cloudbuild to generate and push the image.

for more context see the slack 🧵  https://kubernetes.slack.com/archives/C09QZFCE5/p1615563067038600bump

- Also bump the go version to the lastest 1.13 available

`k/test-infra` PR to add the prow job: TBC

/assign @spiffxp @msau42 @mattcary 

```release-note
Add cloudbuild config to build gcp-compute-persistent-disk-csi-driver image
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,26 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+# For more information about Image pushing refer to https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md
+timeout: 3600s
+
+options:
+  substitution_option: ALLOW_LOOSE
+
+steps:
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20201130-750d12f'
+    entrypoint: make
+    env:
+    - GCE_PD_CSI_STAGING_IMAGE=gcr.io/${_STAGING_PROJECT}/gcp-compute-persistent-disk-csi-driver
+    - GCE_PD_CSI_STAGING_VERSION=${_PULL_BASE_REF}
+    # default cloudbuild has HOME=/builder/home and docker buildx is in /root/.docker/cli-plugins/docker-buildx
+    # set the home to /root explicitly to if using docker buildx
+    - HOME=/root
+    args:
+      - build-and-push-multi-arch
+
+substitutions:
+  _STAGING_PROJECT: 'k8s-staging-cloud-provider-gcp'
+  _PULL_BASE_REF: 'master'
+
+tags:
+- 'gcp-compute-persistent-disk-csi-driver'
+- ${_PULL_BASE_REF}
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.13.4 as builder
+FROM golang:1.13.15 as builder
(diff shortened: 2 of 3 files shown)


mattcary:
Note the integration failure is due to #721 for which a fix is on its way.

cloudbuild.yaml
```diff
@@ -0,0 +1,28 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 3600s
+
+options:
+  substitution_option: ALLOW_LOOSE
+
+steps:
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20201130-750d12f'
+    entrypoint: 'bash'
```
mattcary, line 9: Can't we just use make as the entrypoint? See eg https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/cloudbuild.yaml.

cloudbuild.yaml
```diff
@@ -0,0 +1,28 @@
...
+    # default cloudbuild has HOME=/builder/home and docker buildx is in /root/.docker/cli-plugins/docker-buildx
+    # set the home to /root explicitly to if using docker buildx
+    - HOME=/root
+    args:
+    - -c
+    - |
+      make build-and-push-multi-arch
+
+substitutions:
+  _REGISTRY: 'fake.repository/registry-name'
```
mattcary, line 23: How does this registry work?

In particular I think in the filestore driver we seem to just get it from the project (q.v.)

cpanato: this is set in the prow job, to configure the registry name like `gcr.io/my-registry` we did that in the k/release to support other registries in the cloudbuild for any other users that are using that. But we can make it static in this case. I don't have a strong opinion on that.

cloudbuild.yaml
cpanato: sounds good, will update

cloudbuild.yaml
cpanato: reverted and make similar in what we have for gcp-filestore-csi-driver

cpanato:
testing this cloudbuild on my GCP account, and everything worked fine

mattcary:
/lgtm

thanks tons for helping with this!

spiffxp (approved):
/approve
/lgtm

spiffxp:
/assign @mattcary @msau42 
For the /approve, I don't have rights here

mattcary:
Oops I can never keep track of when I have to lgtm and when I have to approve. Thanks @spiffxp 

/lgtm
/approve

cpanato:
job ran successfully (https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-gcp-compute-persistent-disk-csi-driver-push-images/1371915861887029248)  and the master tag is available: `docker pull gcr.io/k8s-staging-cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:master`

PR #747, opened 2021-04-16
Sanitize GRPC logging

tsmetana:
**What type of PR is this?**
/kind cleanup

We should use the protosanitizer on the GRPC logs in order not to leak some sensitive data to the logs.

This PR adds dependency on protosanitizer from csi-lib-utils, hence the changes in vendor, go.mod, go.sum, etc.

```release-note
NONE
```


msau42:
We originally added protosanitizer and removed it because we found it had signficant performance degradation. Our driver does not use secrets so I don't think we need it. https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/356#issuecomment-550529004

PR #741, opened 2021-04-13
Build multi-arch images for ARM64 and AMD64 at the same time

sarandia:
We can use buildx, which is already set up in the Makefile, to
simultaneously build a multi-arch Docker image for this component for
ARM64 and AMD64, such that the image will work for both architectures.

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
We can publish an image that is multi-arch across X86 and ARM64.
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #751, opened 2021-04-18
Configure QEMU userspace emulation for Linux build

sarandia:
The Dockerfile uses commands in Linux, which require the correct binfmt
for the architecture we build for. Therefore we need to configure Docker
to use QEMU userspace emulation to build images whose arches are
different from the host machine.

**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This is a follow-up to #741. We need to configure Docker buildx correctly for the changes in #741 to work.

Fixes #

```release-note

```


sarandia:
@jingxu97 @mattcary PTAL, thanks!

PR #759, opened 2021-04-20
Fix label test used in staging

mattcary:
/kind bug

The staging overlay has a bad command-line patch which tests the label command line.

```release-note
None
```


mattcary:
The subpath failure I think is a flake unrelated to this cl (https://github.com/kubernetes/kubernetes/issues/101331).

The ephemeral problem looks like infrastructure (but it seems to be happening more often these days).

/retest

PR #760, opened 2021-04-21
fix missing variable when building the image on cloudbuild

cpanato:
**What type of PR is this?**
/kind bug

When we built the image `k8s.gcr.io/cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:v1.2.1` we missed one parameter that needs to be set during the compilation process

This PR fix that, making it similar in how we build the image for windows.

built on my GCP account and now we can run the image without the error we saw before

```
$ docker run -it k8s.gcr.io/cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:v1.2.1
F0421 13:50:23.955027       1 main.go:71] version must be set at compile time

# after the fix
$ docker run -it gcr.io/cpanato-general/gcp-compute-persistent-disk-csi-driver:v99.99.99
W0421 13:50:31.576613       1 gce.go:137] GOOGLE_APPLICATION_CREDENTIALS env var not set
F0421 13:50:31.576650       1 main.go:103] Failed to get cloud provider: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

We will need to release a version `v1.2.2` to fix the image issue

this PR is needed to unblock https://github.com/kubernetes/kubernetes/pull/100747

/assign @msau42 @mattcary @saikat-royc 

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,16 +13,19 @@
 # limitations under the License.
 
 FROM golang:1.13.15 as builder
+
+ARG STAGINGVERSION
+
 WORKDIR /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 ADD . .
-RUN make
+RUN GCE_PD_CSI_STAGING_VERSION=${STAGINGVERSION} make gce-pd-driver
 
 # MAD HACKS: Build a version first so we can take the scsi_id bin and put it somewhere else in our real build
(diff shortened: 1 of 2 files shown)


mattcary:
We should update this to buster 1.6.0 too.

cpanato:
@mattcary done. ptal

I also build the image on my cloudbuild and works fine

mattcary:
/lgtm

The presubmit is still not very healthy, we're working on it :-/

PR #766, opened 2021-04-23
Change mount utils to "k8s.io/mount-utils"

jingxu97:
k8s.io/utils/mount is deprecating. Change to use k8s.io/mount-utils

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```


jingxu97:
cc @andyzhangx

go.mod
```diff
@@ -28,8 +27,9 @@ require (
 	k8s.io/component-base v0.19.0
 	k8s.io/klog v1.0.0
 	k8s.io/kubernetes v1.18.0
+	k8s.io/mount-utils v0.20.6
 	k8s.io/test-infra v0.0.0-20200115230622-70a5174aa78d
-	k8s.io/utils v0.0.0-20200324210504-a9aa75ae1b89
```
msau42, line 32: This line should be removed.

jingxu97: I see there are places like using k8s.io/utils/exec etc. 
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/search?q=%22k8s.io%2Futils%2Fexec%22

jingxu97: So I think we should leave this line here? @msau42

msau42: yup thanks for checking!

PR #770, opened 2021-04-26
Move ARG decls to top of Dockerfile

mattcary:
This appears to be necessary to get docker build to work, when BUILDPLATFORM is manually passed in rather than through the magic of buildx.

```release-note
None
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,17 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+ARG BUILDPLATFORM
+
 FROM --platform=$BUILDPLATFORM golang:1.13.15 as builder
 
 ARG STAGINGVERSION
-ARG BUILDPLATFORM
 ARG TARGETPLATFORM
 
 WORKDIR /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 ADD . .
-# RUN mkdir -p bin \
-#   && GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') go build -mod=vendor -ldflags "-X main.version=${STAGINGVERSION}" -o bin/gce-pd-csi-driver ./cmd/gce-pd-csi-driver/
-RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=${STAGINGVERSION} make gce-pd-driver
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,17 +12,16 @@
...
 ARG STAGINGVERSION
-ARG BUILDPLATFORM
 ARG TARGETPLATFORM
 
 WORKDIR /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 ADD . .
-# RUN mkdir -p bin \
-#   && GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') go build -mod=vendor -ldflags "-X main.version=${STAGINGVERSION}" -o bin/gce-pd-csi-driver ./cmd/gce-pd-csi-driver/
-RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=${STAGINGVERSION} make gce-pd-driver
+RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=$STAGINGVERSION make gce-pd-driver
```
jingxu97, line 24: just wondering removing {} for STAGINGVERSION is needed?

Makefile
```diff
@@ -33,18 +33,20 @@ WINDOWS_IMAGE_TAGS=ltsc2019 1909 2004 20H2
...
 
-gce-pd-driver-windows:
+gce-pd-driver-windows: require-GCE_PD_CSI_STAGING_VERSION
 	mkdir -p bin
 	GOOS=windows go build -mod=vendor -ldflags -X=main.version=$(STAGINGVERSION) -o bin/${DRIVERWINDOWSBINARY} ./cmd/gce-pd-csi-driver/
 
-build-container: require-GCE_PD_CSI_STAGING_IMAGE init-buildx
-	$(DOCKER) buildx build --platform=linux \
+build-container: require-GCE_PD_CSI_STAGING_IMAGE require-GCE_PD_CSI_STAGING_VERSION init-buildx
+	$(DOCKER) buildx build --platform=linux --progress=plain \
```
jingxu97, line 45: what is progress here?

mattcary: It makes the output much cleaner on some terminal types. Otherwise docker gets very noisy.

Also on some platforms it seems that autodetecting the terminal works badly and can cause docker to crash :-(

Dockerfile
mattcary: I changed it to be consistent with the use of TARGETPLATFORM.

Also, I wasn't sure how the RUN command was actually parsed. I think the problem is actually the ordering of the ARGS in Dockerfile, and not their use here, so it doesn't matter.

PR #774, opened 2021-04-28
Update parallel setting for in integration test

jingxu97:
Current parallel setting to 25 is too high for Windows. Every node will
have about 8-10 disks are being attached and fomatted around the same
time. For Windows Disk format time could reach about 4 mins which makes
the whole pod start time exceeds 5 mins and timeout.

Set parallel to 6 for Windows

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -70,6 +70,7 @@ var (
 	testFocus     = flag.String("test-focus", "", "test focus for Kubernetes e2e")
 
 	useKubeTest2 = flag.Bool("use-kubetest2", false, "use kubetest2 to run e2e tests")
+	parallel     = flag.Int("parallel", 10, "the number of parallel tests setting for ginkgo parallelism")
 )
 
 const (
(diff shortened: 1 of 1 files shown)


mattcary:
Can we just plumb through an argument so that we can set it in the run scripts?

I was about to look into this for linux too, as a reason for why tests have gotten flakier. We could change the default to 10 for any platform.

jingxu97:
> Can we just plumb through an argument so that we can set it in the run scripts?
> 
> I was about to look into this for linux too, as a reason for why tests have gotten flakier. We could change the default to 10 for any platform.

Sure, that's better. I updated the code. PTAL. Thanks!

mattcary:
/lgtm 

Thanks!

TeweiLuo:
Thanks for adding the parameter! Maybe set these values via an environment variable in the test scripts so we can adjust them more easily in test configs?

PR #777, opened 2021-04-30
 put GCFLAGS back in Makefile for debug builds

mattcary:
/kind bug
```release-note
None
```


mauriciopoppe:
/ltgm
/approve

Don't know if I can execute these commands

PR #779, opened 2021-05-10
Update csi sidecar versions on staging rc

Jiawei0227:
Upgrade csi sidecar to the latest version except csi-snapshotter.


deploy/kubernetes/base/controller/csidriver_info.yaml
```diff
@@ -1,5 +1,4 @@
-#### TODO: This need to be updated to v1 once we official support 1.18+
-apiVersion: storage.k8s.io/v1beta1
```
mattcary, line 2: We say that dev supports 1.17+, so we can't move to v1 csi driver yet.

Jiawei0227: sure let me revert it.

Jiawei0227: Done.

mattcary:
/lgtm

thanks!

PR #780, opened 2021-05-12
Update csi sidecar versions on stable

Jiawei0227:
Upgrade csi sidecar version to latest on stable-master

/release-note-none
/assign @mattcary


mattcary:
confirmed tests passing on internal testgrid with staging-rc.

/lgtm

Issue #781, opened 2021-05-19
Snapshots cannot be used imported projects


christian-roggia:
This issue is related to the topic discussed here:  https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/224.

According to the documentation, it should be possible to import pre-existing snapshots across projects: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/docs/kubernetes/user-guides/snapshots.md

But this doesn't seem to be the case:

> 'Failed to check and update snapshot content: failed to list snapshot for content restored-snapshot-content: "can not find snapshot for snapshotID projects/PROJECT_ID/global/snapshots/SNAPSHOT_ID"'

My first guess was that the CSI driver did not have the right IAM permissions in the cluser.

I attempted to grant the following permissions:
```
    "compute.disks.get",
    "compute.disks.create",
    "compute.disks.createSnapshot",
    "compute.snapshots.list",
    "compute.snapshots.get",
    "compute.snapshots.create",
    "compute.snapshots.useReadOnly",
    "compute.snapshots.delete",
    "compute.zones.get",
    "storage.objects.create",
    "storage.objects.delete",
    "storage.objects.get",
    "storage.objects.list",
    "iam.serviceAccounts.signBlob"
```

to the following service accounts:
```
    "serviceAccount:PROJECT_NUMBER-compute@developer.gserviceaccount.com",
    "serviceAccount:service-PROJECT_NUMBER@compute-system.iam.gserviceaccount.com",
    "serviceAccount:service-PROJECT_NUMBER@container-engine-robot.iam.gserviceaccount.com",
    "serviceAccount:pdcsi-workload@PROJECT_ID.iam.gserviceaccount.com",
```

where

```
pdcsi-workload@PROJECT_ID.iam.gserviceaccount.com
```

is an identity workload with service account
```
PROJECT_ID.svc.id.goog[kube-system/pdcsi-node]
```

---
I am assuming that the CSI driver is not able to import snapshots from another project because of the way it was implemented.

If this was not the case and the CSI driver is supposed (and has been observed) to work cross-projects then any clue would be helpful to understand why it is failing to locate the snapshot in my case (is it because of IAM? did I misconfigure something? is my CSI driver version too old?).

Thank you!

christian-roggia:
I confirm that the project is currently hardcoded here:

[...]

PR #782, opened 2021-05-19
Allow cross project snapshots and volumes

christian-roggia:
**What type of PR is this?**

/kind feature

This PR allows using CSI snapshots across projects.

This feature enables GKE users to create VolumeSnapshots in a project FOO and import them in a project BAR.
This enhancement simplifies the migration process from between projects and improves the user experience with tools like Velero.

Fixes https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/781

```release-note
- It is now possible to access snapshots and volumes across different projects.
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -72,7 +72,7 @@ type GCECompute interface {
 	// Zone Methods
 	ListZones(ctx context.Context, region string) ([]string, error)
 	ListSnapshots(ctx context.Context, filter string, maxEntries int64, pageToken string) ([]*computev1.Snapshot, string, error)
-	GetSnapshot(ctx context.Context, snapshotName string) (*computev1.Snapshot, error)
+	GetSnapshot(ctx context.Context, project, snapshotName string) (*computev1.Snapshot, error)
 	CreateSnapshot(ctx context.Context, volKey *meta.Key, snapshotName string) (*computev1.Snapshot, error)
 	DeleteSnapshot(ctx context.Context, snapshotName string) error
 }
@@ -765,10 +765,12 @@ func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone
 	return instance, nil
 }
 
-func (cloud *CloudProvider) GetSnapshot(ctx context.Context, snapshotName string) (*computev1.Snapshot, error) {
+func (cloud *CloudProvider) GetSnapshot(ctx context.Context, project, snapshotName string) (*computev1.Snapshot, error) {
 	klog.V(5).Infof("Getting snapshot %v", snapshotName)
 	svc := cloud.service
-	project := cloud.project
+	if project == "" {
+		project = cloud.project
+	}
 	snapshot, err := svc.Snapshots.Get(project, snapshotName).Context(ctx).Do()
 	if err != nil {
 		return nil, err
@@ -908,7 +910,7 @@ func (cloud *CloudProvider) waitForSnapshotCreation(ctx context.Context, snapsho
 		select {
 		case <-ticker.C:
 			klog.V(6).Infof("Checking GCE Snapshot %s.", snapshotName)
-			snapshot, err := cloud.GetSnapshot(ctx, snapshotName)
+			snapshot, err := cloud.GetSnapshot(ctx, "", snapshotName)
 			if err != nil {
 				klog.Warningf("Error in getting snapshot %s, %v", snapshotName, err)
 			} else if snapshot != nil {
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -54,7 +54,7 @@ type GCECompute interface {
 	GetDefaultZone() string
 	// Disk Methods
 	GetDisk(ctx context.Context, project string, volumeKey *meta.Key, gceAPIVersion GCEAPIVersion) (*CloudDisk, error)
-	RepairUnderspecifiedVolumeKey(ctx context.Context, project string, volumeKey *meta.Key) (*meta.Key, error)
+	RepairUnderspecifiedVolumeKey(ctx context.Context, project string, volumeKey *meta.Key) (string, *meta.Key, error)
 	ValidateExistingDisk(ctx context.Context, disk *CloudDisk, params common.DiskParameters, reqBytes, limBytes int64, multiWriter bool) error
 	InsertDisk(ctx context.Context, project string, volKey *meta.Key, params common.DiskParameters, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID string, multiWriter bool) error
 	DeleteDisk(ctx context.Context, project string, volumeKey *meta.Key) error
@@ -106,11 +106,14 @@ func (cloud *CloudProvider) ListDisks(ctx context.Context, maxEntries int64, pag
 
 // RepairUnderspecifiedVolumeKey will query the cloud provider and check each zone for the disk specified
 // by the volume key and return a volume key with a correct zone
-func (cloud *CloudProvider) RepairUnderspecifiedVolumeKey(ctx context.Context, project string, volumeKey *meta.Key) (*meta.Key, error) {
+func (cloud *CloudProvider) RepairUnderspecifiedVolumeKey(ctx context.Context, project string, volumeKey *meta.Key) (string, *meta.Key, error) {
 	klog.V(5).Infof("Repairing potentially underspecified volume key %v", volumeKey)
+	if project == common.UnspecifiedValue {
+		project = cloud.project
+	}
 	region, err := common.GetRegionFromZones([]string{cloud.zone})
 	if err != nil {
-		return nil, fmt.Errorf("failed to get region from zones: %v", err)
+		return "", nil, fmt.Errorf("failed to get region from zones: %v", err)
 	}
 	switch volumeKey.Type() {
 	case meta.Zonal:
@@ -119,7 +122,7 @@ func (cloud *CloudProvider) RepairUnderspecifiedVolumeKey(ctx context.Context, p
 			// list all zones, try to get disk in each zone
 			zones, err := cloud.ListZones(ctx, region)
 			if err != nil {
-				return nil, err
+				return "", nil, err
 			}
 			for _, zone := range zones {
 				_, err := cloud.getZonalDiskOrError(ctx, project, zone, volumeKey.Name)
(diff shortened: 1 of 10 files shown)


msau42:
/ok-to-test
/assign @mattcary 
This is an interesting idea but I'm not sure we'll be able to support it in GKE. The service account that we use the run the driver in GKE is limited to a single project. I'm not even sure a GCP service account can have cross project permissions?

mattcary:
At any rate it is confusing that as written we will silently change the project given in a snapshot id.

@jingxu97 it looks like you implemented this, do you have any background?

christian-roggia:
Service accounts can normally have cross-project permissions. This is actually a very common scenario.

I believe that the CSI driver has a dedicated "Service Agent" service account, but it should behave like any other service account.

Additionally, the CSI driver in GKE is mounting a Kubernetes service account. That Kubernetes-native SA can be used as a [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) binding it to a GCP IAM service account managed by the project owner instead of GCP (i.e. this allows to avoid granting permissions directly to the "Service Agent" service account).

In both cases allowing the right IAM permissions cross-project should be possible.

christian-roggia:
@mattcary it is worth mentioning that the proposed PR solves the issue only for the lookup operation (i.e. GetSnapshot), but actually all operations currently performed by the driver have an hardcoded project.

I believe that in all cases where a fully qualified path is provided to the driver it should attempt to perform the operation with the project given in the path instead of silently replacing it.

I can update the PR with changes to all operations that should be provided with the project name.

mattcary:
> I can update the PR with changes to all operations that should be provided with the project name.

+1 let's do that. Even if for some reason the cross-project roles don't work out (I think you're right but I want to double-check), we should not be changing the project name. eg a user may have different resources in two different projects sharing the same name.  It would be much better to fail in that situation than silently switch projects....

mattcary:
> /ok-to-test
> /assign @mattcary
> This is an interesting idea but I'm not sure we'll be able to support it in GKE. The service account that we use the run the driver in GKE is limited to a single project. I'm not even sure a GCP service account can have cross project permissions?

Cross-project snapshots are a thing :-) https://cloud.google.com/compute/docs/disks/create-snapshots#sharing_snapshots

It works without any additional roles being necessary when run as a user that is owner of two different projects. I don't see why a service account couldn't be given the same roles on two different projects either. However, in GKE, as Michelle said, which service account is used by master pods to authenticate to GCP is a little obscure. I'm still digging into that.

However, since cross-project snapshots work so smoothly I don't see any reason for the pd csi driver not to support them:

```
/clank/go/src/k8s.io/kubernetes$ gcloud compute instances list
NAME                                 ZONE           MACHINE_TYPE   PREEMPTIBLE  INTERNAL_IP  EXTERNAL_IP      STATUS
e2e-test-mattcary-master             us-central1-b  n1-standard-1               10.128.0.2   104.198.179.125  RUNNING
e2e-test-mattcary-minion-group-017w  us-central1-b  n1-standard-2               10.128.0.5   35.239.144.136   RUNNING
e2e-test-mattcary-minion-group-g4rb  us-central1-b  n1-standard-2               10.128.0.4   34.70.139.106    RUNNING
e2e-test-mattcary-minion-group-qtm1  us-central1-b  n1-standard-2               10.128.0.3   35.193.112.53    RUNNING
/clank/go/src/k8s.io/kubernetes$ gcloud compute disks create --zone us-central1-b --size 10G cross-test --project mattcary-e2e
WARNING: You have selected a disk size of under [200GB]. This may result in poor I/O performance. For more information, see: https://developers.google.com/compute/docs/disks#performance.
Created [https://www.googleapis.com/compute/v1/projects/mattcary-e2e/zones/us-central1-b/disks/cross-test].
NAME        ZONE           SIZE_GB  TYPE         STATUS
cross-test  us-central1-b  10       pd-standard  READY

New disks are unformatted. You must format and mount a disk before it
can be used. You can find instructions on how to do this at:

https://cloud.google.com/compute/docs/disks/add-persistent-disk#formatting

/clank/go/src/k8s.io/kubernetes$ gcloud compute instances attach-disk e2e-test-mattcary-minion-group-qtm1 --disk cross-test
Updated [https://www.googleapis.com/compute/v1/projects/mattcary-e2e/zones/us-central1-b/instances/e2e-test-mattcary-minion-group-qtm1].
/clank/go/src/k8s.io/kubernetes$ gcloud compute ssh e2e-test-mattcary-minion-group-qtm1
/clank/go/src/k8s.io/kubernetes$ gcloud compute instances describe e2e-test-mattcary-minion-group-qtm1
...
- autoDelete: false
  boot: false
  deviceName: persistent-disk-2
  diskSizeGb: '10'
  index: 2
  interface: SCSI
  kind: compute#attachedDisk
  mode: READ_WRITE
  source: https://www.googleapis.com/compute/v1/projects/mattcary-e2e/zones/us-central1-b/disks/cross-test
  type: PERSISTENT
...
/clank/go/src/k8s.io/kubernetes$ gcloud compute ssh e2e-test-mattcary-minion-group-qtm1
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ ls -l /dev/disk/by-id/google-persistent-disk-2
lrwxrwxrwx 1 root root 9 May 21 16:15 /dev/disk/by-id/google-persistent-disk-2 -> ../../sdc
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ ls /dev/sdc*
/dev/sdc
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ sudo mkfs.ext4 -m 0 /dev/sdc 
...
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ sudo mkdir /mnt/disks/cross-test
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ sudo mount /dev/sdc /mnt/disks/cross-test
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ ls /mnt/disks/cross-test
lost+found
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ echo cross-test | sudo tee -a /mnt/disks/cross-test/hello
cross-test
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ ls /mnt/disks/cross-test
hello  lost+found
mattcary@e2e-test-mattcary-minion-group-qtm1 ~ $ exit
logout
Connection to 35.193.112.53 closed.
/clank/go/src/k8s.io/kubernetes$ gcloud compute disks snapshot cross-test
Creating snapshot(s) ystif0358wgz...done.
/clank/go/src/k8s.io/kubernetes$ gcloud compute instances create snapshot-test --project mattcary-snapshot --machine-type e2-small
/clank/go/src/k8s.io/kubernetes$ gcloud compute disks create duplicate-cross-test --project mattcary-snapshot --source-snapshot projects/mattcary-e2e/global/snapshots/ystif0358wgz
Created [https://www.googleapis.com/compute/v1/projects/mattcary-snapshot/zones/us-central1-b/disks/duplicate-cross-test].
NAME                  ZONE           SIZE_GB  TYPE         STATUS
duplicate-cross-test  us-central1-b  10       pd-standard  READY
/clank/go/src/k8s.io/kubernetes$ gcloud compute instances attach-disk --project mattcary-snapshot snapshot-test --disk duplicate-cross-test
Updated [https://www.googleapis.com/compute/v1/projects/mattcary-snapshot/zones/us-central1-b/instances/snapshot-test].
/clank/go/src/k8s.io/kubernetes$ gcloud compute instances describe snapshot-test --project mattcary-snapshot
...
- autoDelete: false
  boot: false
  deviceName: persistent-disk-1
  diskSizeGb: '10'
  index: 1
  interface: SCSI
  kind: compute#attachedDisk
  mode: READ_WRITE
  source: https://www.googleapis.com/compute/v1/projects/mattcary-snapshot/zones/us-central1-b/disks/duplicate-cross-test
  type: PERSISTENT
...
/clank/go/src/k8s.io/kubernetes$ gcloud compute ssh --project mattcary-snapshot snapshot-test
mattcary@snapshot-test:~$ ls -l /dev/disk/by-id | grep disk-1
lrwxrwxrwx 1 root root  9 May 21 16:53 google-persistent-disk-1 -> ../../sdb
lrwxrwxrwx 1 root root  9 May 21 16:53 scsi-0Google_PersistentDisk_persistent-disk-1 -> ../../sdb
mattcary@snapshot-test:~$ sudo mount /dev/sdb /mnt
mattcary@snapshot-test:~$ ls /mnt
hello  lost+found
mattcary@snapshot-test:~$ cat /mnt/hello
cross-test
```

christian-roggia:
I should be able to find some time to work on it this week, do you have any clue why integration tests failed?

mattcary:
Yes, this looks like the same issue as #102077, but on the pd csi driver
side. I'll take a look. It's not your fault :-)
```
W0521 02:25:46.569] F0521 02:25:46.566914   91994 helpers.go:116] error:
unable to recognize
"/tmp/gcp-compute-persistent-disk-csi-driver-specs-generated.yaml": no
matches for kind "CSIDriver" in version "storage.k8s.io/v1beta1"
```

mattcary:

I think (hope) that #783 fixed this.

christian-roggia:
Even though the tests are all green, I feel like new tests should be added (especially e2e) to verify that the CSI driver can import snapshots and volumes from another project.

It is also important to verify which service account requires additional permission when performing cross-project operations, and which IAM permissions should be granted.

As described in my original issue, the candidate service accounts are the following:
```
    "serviceAccount:PROJECT_NUMBER-compute@developer.gserviceaccount.com",
    "serviceAccount:service-PROJECT_NUMBER@compute-system.iam.gserviceaccount.com",
    "serviceAccount:service-PROJECT_NUMBER@container-engine-robot.iam.gserviceaccount.com",
    "serviceAccount:pdcsi-workload@PROJECT_ID.iam.gserviceaccount.com",
```

and maybe workload identity might be required for the CSI SA:
```
pdcsi-workload@PROJECT_ID.iam.gserviceaccount.com
```
```
PROJECT_ID.svc.id.goog[kube-system/pdcsi-node-sa]
```

finally, I would expect the SA to require at least the following IAM permissions for snapshots:
```
compute.snapshots.list
compute.snapshots.get
compute.snapshots.useReadOnly
```

with additional permissions on disks that might be necessary:
```
compute.disks.get
compute.disks.list
compute.disks.use
compute.disks.useReadOnly
```

mattcary:
> Even though the tests are all green, I feel like new tests should be added (especially e2e) to verify that the CSI driver can import snapshots and volumes from another project.
> 
Thanks! I think that's a great point. However, the logistics of that get complicated. I think it should be possible to do it by renting another project, but then that would have to be plumbed through to the e2e test which is breaking new ground and is therefore complicated?

So I'd be okay both with holding this change until those e2e tests are done, or pushing this out and relying on the VolumeIdToKey unit tests for now.

It's my understanding that on GKE, it will be enough to give the container-engine-robot service accounts roles in the cross-project---that's the account plumbed to the pd csi cloud provider. That of course requires running on GKE, which we do in our internal tests, and so that would also require more plumbing in the k8s-integration tests.

For GCE (ie, kube-up) we'd have to check in the /etc/gke.conf that's installed on the master node. We're in the middle of making the pd csi driver on by default for kube-up which might clarify this issue a bit.

WDYT? In any case thanks again for noticing & clearing up this cross-project problem.

christian-roggia:
> However, the logistics of that get complicated. I think it should be possible to do it by renting another project, but then that would have to be plumbed through to the e2e test which is breaking new ground and is therefore complicated?

I was expecting that this setup wouldn't be simple, I would suggest pushing into master then and manually verify that cross-project VolumeSnapshots work without major issues, the Velero community might help to test this feature out too.

> WDYT? In any case thanks again for noticing & clearing up this cross-project problem.

You are more than welcome! The CSI driver is a very powerful tool and I am happy to be able to help out.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -306,7 +303,7 @@ func ValidateDiskParameters(disk *CloudDisk, params common.DiskParameters) error
 	return nil
 }
 
-func (cloud *CloudProvider) InsertDisk(ctx context.Context, volKey *meta.Key, params common.DiskParameters, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID string, multiWriter bool) error {
+func (cloud *CloudProvider) InsertDisk(ctx context.Context, project string, volKey *meta.Key, params common.DiskParameters, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID string, multiWriter bool) error {
```
mattcary, line 306: Since disks cannot be used cross-project, why are we plumbing the project around everywhere? Wouldn't it be easier & less toil to use cloud.project for everything except the snapshot api?

christian-roggia: My understanding was that we didn't want to override the original project provided in the resource's names.

I can revert these changes if they don't make sense to you as they provide no additional functionalities.

mattcary: Ah, good point. Thank you, I've spent too long between reviewing sessions.

/lgtm
/approve

husunal:
Thanks for the PR. In order to consume snapshots between GCP projects, should we use `VolumeSnapshotContent` and specify the cross-project snapshot resource in the `snapshotHandle`?

christian-roggia:
@husunal yes, that would be the expected use of cross-projects snapshots.

PR #785, opened 2021-06-08
Revert "Use kubetest instead of kubetest2 for CSI migration test"

pacoxu:
Reverts kubernetes-sigs/gcp-compute-persistent-disk-csi-driver#775

image


mattcary:
/hold

We had to revert back to kubetest in order to fix some problems. We're figuring out what the right thing to do here is.

TeweiLuo:
We haven't figured out how to make kubetest2 work for migration tests yet so let's hold on to this PR.

pacoxu:
https://prow.k8s.io/?job=ci-gce-pd-csi-driver-latest-k8s-master-windows-2019-migration
green now

PR #786, opened 2021-06-08
Disable default pd csi install

leiyiz:
/kind feature
because pd-csi driver is installed by default in k/k, we want to disable that when launching test from this repo

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -108,6 +108,13 @@ func clusterUpGCE(k8sDir, gceZone string, numNodes int, imageType string) error 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mattcary:
/ok-to-test
/hold

This looks about right. We'll probably end up changing the env variable to something like ENABLE_GCE_PD_CSI_DRIVER or something like that, so we can sit on this PR until the k/k one gets closer.

leiyiz:
#788 seems to have the same testing failure related to #789

mattcary:
/remove-lifecycle stale

We'll need to coordinate this with turning migration on by default / using
the default driver install (in cloud-provider-gcp).

mattcary:
This is outdated and can be closed

PR #793, opened 2021-06-21
Allow users to specify "storage-locations" for snapshots.

### LINKED ISSUE #791: Allow users to specify "Storage Locations" in snapshot class
Current we don't set any storage locations and a default location (https://cloud.google.com/compute/docs/disks/snapshots#default_location) is used. This can be problematic for some users.

TeweiLuo:
Use a comma separated string, for example "us-central1,us-west1"
as the "snapshot-locations" field of snapshot class parameters.
See the PR for detailed examples.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind feature

Allow users to set storage locations[1] for their PD snapshots by specifying them in the snapshot class. For example:

```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  name: snapshot-class-with-storage-locations
parameters:
  storage-locations: us-central1
driver: pd.csi.storage.gke.io
deletionPolicy: Delete
```

Current the default location[2] will be used for all snapshots. GKE customers may have organization policies restricting the placement of snapshots that conflicts with the default location. This features will help customers to consistently apply their policies.

This parameter is described in the API document[3] for snapshots.

Fixes #791

```release-note
Users will be able to set the storage locations for snapshots by specifying them in the snapshot class.
```
[1] https://cloud.google.com/compute/docs/disks/snapshots#custom_location
[2] https://cloud.google.com/compute/docs/disks/snapshots#default_location
[3] https://cloud.google.com/compute/docs/reference/rest/v1/snapshots

### DIFF
--- a/pkg/common/parameters.go
+++ b/pkg/common/parameters.go
@@ -22,10 +22,14 @@ import (
 )
 
 const (
+	// Parameters for StorageClass
 	ParameterKeyType                 = "type"
 	ParameterKeyReplicationType      = "replication-type"
 	ParameterKeyDiskEncryptionKmsKey = "disk-encryption-kms-key"
 	ParameterKeyLabels               = "labels"
+
+	// Parameters for VolumeSnapshotClass
+	ParameterKeyStorageLocations = "storage-locations"
 
 	replicationTypeNone = "none"
 
@@ -58,6 +62,11 @@ type DiskParameters struct {
 	// Values: {map[string]string}
 	// Default: ""
 	Labels map[string]string
+}
+
+// SnapshotParameters contains normalized and defaulted parameters for snapshots
+type SnapshotParameters struct {
+	StorageLocations []string
 }
 
 // ExtractAndDefaultParameters will take the relevant parameters from a map and
@@ -115,6 +124,25 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 	}
 	if len(p.Tags) > 0 {
 		p.Tags[tagKeyCreatedBy] = driverName
+	}
+	return p, nil
+}
+
+func ExtractAndDefaultSnapshotParameters(parameters map[string]string) (SnapshotParameters, error) {
+	p := SnapshotParameters{
+		StorageLocations: []string{},
+	}
+	for k, v := range parameters {
+		switch strings.ToLower(k) {
+		case ParameterKeyStorageLocations:
+			normalizedStorageLocations, err := ProcessStorageLocations(v)
+			if err != nil {
+				return p, err
+			}
+			p.StorageLocations = normalizedStorageLocations
+		default:
+			return p, fmt.Errorf("parameters contains invalid option %q", k)
+		}
 	}
 	return p, nil
 }
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -50,6 +50,18 @@ const (
 	nodeIDTotalElements = 6
 
 	regionalDeviceNameSuffix = "_regional"
+
+	// Snapshot storage location format
+	// Reference: https://cloud.google.com/storage/docs/locations
+	// Example: us
+	multiRegionalLocationFmt = "^[a-z]+$"
+	// Example: us-east1
+	regionalLocationFmt = "^[a-z]+-[a-z]+[0-9]$"
+)
+
+var (
+	multiRegionalPattern = regexp.MustCompile(multiRegionalLocationFmt)
+	regionalPattern      = regexp.MustCompile(regionalLocationFmt)
 )
 
 func BytesToGbRoundDown(bytes int64) int64 {
@@ -216,4 +228,13 @@ func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
 	}
 
 	return labelsMap, nil
+}
+
+// ProcessStorageLocations trims and normalizes storage location to lower letters.
+func ProcessStorageLocations(storageLocations string) ([]string, error) {
+	normalizedLoc := strings.ToLower(strings.TrimSpace(storageLocations))
+	if !multiRegionalPattern.MatchString(normalizedLoc) && !regionalPattern.MatchString(normalizedLoc) {
+		return []string{}, fmt.Errorf("invalid location for snapshot: %q", storageLocations)
+	}
+	return []string{normalizedLoc}, nil
 }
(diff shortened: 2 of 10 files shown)


pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -845,7 +845,7 @@ func TestCreateVolumeWithVolumeSource(t *testing.T) {
 		}
 
 		if tc.snapshotOnCloud {
-			gceDriver.cs.CloudProvider.CreateSnapshot(context.Background(), tc.volKey, name)
+			gceDriver.cs.CloudProvider.CreateSnapshot(context.Background(), tc.volKey, name, req.GetParameters()[common.ParameterStorageLocations])
```
mauriciopoppe, line 848: is it possible for `req.GetParameters()` to be nil?

TeweiLuo: I think it could - let me check

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -73,7 +73,7 @@ type GCECompute interface {
 	ListZones(ctx context.Context, region string) ([]string, error)
 	ListSnapshots(ctx context.Context, filter string, maxEntries int64, pageToken string) ([]*computev1.Snapshot, string, error)
 	GetSnapshot(ctx context.Context, project, snapshotName string) (*computev1.Snapshot, error)
-	CreateSnapshot(ctx context.Context, project string, volKey *meta.Key, snapshotName string) (*computev1.Snapshot, error)
+	CreateSnapshot(ctx context.Context, project string, volKey *meta.Key, snapshotName string, storageLocations string) (*computev1.Snapshot, error)
```
mauriciopoppe, line 76: I had a question about using an array but I checked that `parameters` is a `map[string]string` https://github.com/kubernetes-csi/external-snapshotter/blob/f255fe7b46bd5410370537319879d06f2bf1a156/client/apis/volumesnapshot/v1/types.go#L215, I don't know if users would do this though (NOTE the space between the comma and the other region):

```
parameters:
  storage-locations: us-central1, us-central2
```

pkg/common/parameters.go
```diff
@@ -22,11 +22,15 @@ import (
...
 
 const (
+	// Parameters for StorageClass
 	ParameterKeyType                 = "type"
 	ParameterKeyReplicationType      = "replication-type"
 	ParameterKeyDiskEncryptionKmsKey = "disk-encryption-kms-key"
 	ParameterKeyLabels               = "labels"
 
+	// Parameters for VolumeSnapshotClass
+	ParameterStorageLocations = "storage-locations"
```
mauriciopoppe, line 32: For consistency this could be `ParameterKeyStorageLocations` but it's just a nit

mauriciopoppe:
Probably need to add an e2e test too

TeweiLuo:
> Probably need to add an e2e test too

I'm not sure if this should be a k/k e2e test because this param is pd specific. Maybe a GKE internal test is better?

mauriciopoppe:
got it, I saw a few tests in `test/e2e` but they seem to be integration tests, sure having a test either here or in GKE internal would be nice

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -609,7 +609,11 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 			return nil, status.Error(codes.Internal, fmt.Sprintf("Unknown get snapshot error: %v", err))
 		}
 		// If we could not find the snapshot, we create a new one
-		snapshot, err = gceCS.CloudProvider.CreateSnapshot(ctx, project, volKey, req.Name)
+		parameters := make(map[string]string)
+		if req.GetParameters() != nil {
+			parameters = req.GetParameters()
+		}
+		snapshot, err = gceCS.CloudProvider.CreateSnapshot(ctx, project, volKey, req.Name, parameters[common.ParameterStorageLocations])
```
mattcary, line 616: We should split, trim and validate the storage locations here (via a helper ftn of course). See for example common.ExtractAndDefaultParameters and its usages as a model.

TeweiLuo: I was thinking about the same thing initially but 

1. It's debatable whether we should validate the regions themselves - the PD backend may change.
2.  If (1) is true, then there is no other validation, because every string can be treated as a comma-separated string.

mattcary: discussed offline. The API takes a list of strings, so we definitely need to split on commas, so we should do that early. If we're splitting on commas, we need to trim whitespace, which means we should do a regexp validation of the location to make sure it's plausible. Things like validating the regions will be done over in the PD service anyway and not in the driver, even in the cloud provider.

TeweiLuo:
Refactored to use a generalized `SnapshotParameters` struct and added normalization/validation of storage locations. Will add an e2e test.

pkg/common/utils.go
```diff
@@ -217,3 +224,26 @@ func ConvertLabelsStringToMap(labels string) (map[string]string, error) {
...
+// Transform storage locations from a comma separated string to an array,
+// and validate the format of each location. Note that empty string does
+// not need specail treatment because Split() turns it into an empty
+// array, which is the desired value.
+func ProcessStorageLocations(storageLocations string) ([]string, error) {
+	locations := strings.Split(storageLocations, ",")
+
+	var normalizedLocations []string
+
+	multiRegionalPattern, _ := regexp.Compile(multiRegionalLocationFmt)
```
mauriciopoppe, line 237: I don't know how often this is called and if the patterns are known in advance and never changed probably these two calls to `regexp.Compile` could be const and declared at the top

mauriciopoppe:
lgtm :)

mattcary:
Can you add an example to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/docs/kubernetes/user-guides/snapshots.md?

pkg/common/utils.go
TeweiLuo: Ack. I'll pre-compute the pattern.

TeweiLuo:
Update: while the `storageLocations` argument for the snapshot REST API is of type []string but it must be one location only. The only documentation I found is this sentence

```
Determine your snapshot storage location. You can use the default storage location, or a custom storage location.
```

from this [page](https://cloud.google.com/compute/docs/disks/create-snapshots#create_zonal_snapshot). I'll amend the PR accordingly.

pkg/common/parameters.go
TeweiLuo: Done.

pkg/common/utils.go
TeweiLuo: Done.

TeweiLuo:
> Can you

Doc updated.

mattcary:
thanks!

/lgtm

PR #796, opened 2021-07-01
Rename master to main

ikarldasan:
**What type of PR is this?**
> /kind documentation

Needed for https://github.com/kubernetes/k8s.io/issues/1597

Needed for https://github.com/kubernetes/k8s.io/issues/1597


mattcary:
@ikarldasan It looks like this only effects the promotion process, and not our deployment.

/lgtm

mattcary:
force merging as failures are for known flakes.

pohly:
Hmm, isn't this repo also using the process described in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/release-tools/README.md#sharing-and-updating?

Direct modifications to `release-tools` then should be avoided.

PR #803, opened 2021-07-20
fix image repo

mattcary:
I neglected to update the image repo :-(

/kind bug
/assign @saikat-royc 

```release-note
None
```


mattcary:
integration test flaking, force-merged so we can get the image tested.

PR #805, opened 2021-07-22
Fix kubectl parsing

mattcary:
/kind bug
/assign @saikat-royc 

kubectl output has changed and our version parsing needs to be updated.

```release-note
None
```


mattcary:
Just the multivolume flake. Hopefully we'll get to this in our fixit in August!

Issue #809, opened 2021-07-28
Implement ControllerGetVolume RPC in CSI Spec


chrishenzie:
CSI spec v1.3.0 introduces the [`ControllerGetVolume` RPC](https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#controllergetvolume).

This task will involve implementing the logic for `ControllerGetVolume` and advertising related capabilities to inform clients the RPC is supported.

PR #810, opened 2021-07-28
Update container-storage-interface dependency to v1.5.0

chrishenzie:
**What type of PR is this?**
/kind feature

Updates the CSI spec dependency to v1.5.0.

This is required for adding support for new CSI access modes added as part of [KEP-2485](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md) (ReadWriteOncePod access mode).

This is also required for updating Kubernetes dependencies (some local testing showed it required v1.3.0 which introduces the `ControllerGetVolume` RPC). For now I've filed #809 to implement this. My main focus for this PR is just to bump the CSI spec dependency version.

```release-note
NONE
```

/assign @mattcary


chrishenzie:
/assign @msau42
for approval

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -514,6 +515,11 @@ func (gceCS *GCEControllerServer) ValidateVolumeCapabilities(ctx context.Context
 	}, nil
 }
 
+// TODO(#809): Implement logic and advertise related capabilities.
+func (gceCS *GCEControllerServer) ControllerGetVolume(ctx context.Context, req *csi.ControllerGetVolumeRequest) (*csi.ControllerGetVolumeResponse, error) {
+	return nil, errors.New("ControllerGetVolume is not implemented")
```
msau42, line 520: Is there a NOT_IMPL error code we're supposed to return?

saikat-royc: Ah yes. codes.Unimplemented

PR #811, opened 2021-07-28
Update k8s dependencies to v1.22.0-rc.0

chrishenzie:
**What type of PR is this?**
/kind feature

This PR updates k8s dependencies to v1.22.0-rc.0. This will require bumping again to v1.22.0 once it is released.

There isn't any particular feature in v1.22.0 that this driver is going to depend on; just bumping to make sure we stay up to date with latest changes.

This change is rebased on #810 so they can be reviewed independently.

```release-note
NONE
```

/assign @mattcary 
/hold


chrishenzie:
The verify job failure appears to be related to the go version (1.13) that we are using. I cannot repro this failure with 1.16. It's probably a good idea that we update this as well.

chrishenzie:
Opened https://github.com/kubernetes/test-infra/pull/23067 to update the image version to one that uses go1.16.

PR #813, opened 2021-08-04
list volumes in cluster's region instead of zone

leiyiz:
**What type of PR is this?**

/kind bug

when controller is not in the same zone as nodes, DiskList don't list out disks attached in other zones, thus reconciler will force controllerPublishVolume to reprocess them, causing excessive api calls breaking customer clusters' quotas.

Fixes https://github.com/kubernetes-csi/external-attacher/issues/291

No comment still from arcus team so chose this implementation over `aggregatedList` due to concerns over it being potentially slow for projects with PVs in lots of regions.

```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -64,7 +64,7 @@ type GCECompute interface {
 	GetDiskTypeURI(project string, volKey *meta.Key, diskType string) string
 	WaitForAttach(ctx context.Context, project string, volKey *meta.Key, instanceZone, instanceName string) error
 	ResizeDisk(ctx context.Context, project string, volKey *meta.Key, requestBytes int64) (int64, error)
-	ListDisks(ctx context.Context, maxEntries int64, pageToken string) ([]*computev1.Disk, string, error)
+	ListDisks(ctx context.Context) ([]*computev1.Disk, string, error)
 	// Regional Disk Methods
 	GetReplicaZoneURI(project string, zone string) string
 	// Instance Methods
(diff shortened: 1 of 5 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -90,18 +90,33 @@ func (cloud *CloudProvider) GetDefaultZone() string {
 // ListDisks lists disks based on maxEntries and pageToken only in the project
 // and zone that the driver is running in.
 func (cloud *CloudProvider) ListDisks(ctx context.Context, maxEntries int64, pageToken string) ([]*computev1.Disk, string, error) {
-	lCall := cloud.service.Disks.List(cloud.project, cloud.zone)
```
mattcary, line 93: I think we want to drop the maxEntries & pageToken parameters if we're just going to drop them. 

It also looks like we may need to handle paging a step up in ListVolumes, as the caller of that is going to only expect maxEntries results.

leiyiz: agreed that we might just drop maxEntries and pageToken
we can handle the paging in ListVolume by maintaining a list in memory and refresh it whenever there's no pageToken passed in, but I think we wouldn't need to deal with that as long as [attacher](https://github.com/kubernetes-csi/external-attacher/blob/0691388e9b94c43ac0aea6840fa81662b5741722/pkg/attacher/lister.go#L39) remains to be the only caller of this function.

mattcary: Technically some other container orchestrator may be using this driver so we don't want to break the API.

leiyiz: done

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -531,15 +535,30 @@ func (gceCS *GCEControllerServer) ListVolumes(ctx context.Context, req *csi.List
 		klog.Warningf("ListVolumes requested max entries of %v, GCE only supports values <=500 so defaulting value back to 500", maxEntries)
 		maxEntries = 500
 	}
-	diskList, nextToken, err := gceCS.CloudProvider.ListDisks(ctx, maxEntries, req.StartingToken)
-	if err != nil {
-		if gce.IsGCEInvalidError(err) {
-			return nil, status.Error(codes.Aborted, fmt.Sprintf("ListVolumes error with invalid request: %v", err))
+	if maxEntries == 0 {
+		maxEntries = 500
```
mattcary, line 539: I don't think there's any reason to keep the 500 max entries, that was a GCE thing. I think we should we return all entries in a single page if maxEntries == 0.

mattcary (commented):
Just a little nit, otherwise looks good.

pkg/gce-pd-csi-driver/controller.go
leiyiz: done. also removed the limit of 500 if maxEntries is >500

mattcary:
Can you squash to a single commit?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -89,19 +89,31 @@ func (cloud *CloudProvider) GetDefaultZone() string {
...
+		nextPageToken := "pageToken"
+		for nextPageToken != "" {
+			diskList, err := lCall.Do()
+			if err != nil {
+				return nil, "", err
+			}
+			items = append(items, diskList.Items...)
+			nextPageToken = diskList.NextPageToken
+			lCall.PageToken(nextPageToken)
+		}
```
saikat-royc, line 114: nit: this may be more concise and readable.	
```
    var nextToken string
	for {
		resp, err := cloud.service.Disks.List(cloud.project, zone).PageToken(nextToken).Do()
		if err != nil {
			// return error
		}
         items = append(items, diskList.Items...)
		if resp.NextPageToken == "" {
			break
		}
		nextToken = resp.NextPageToken
	}
```

PR #847, opened 2021-09-28
Attach/Detach back off

lizhuqi:
**What type of PR is this?**
/kind feature

In GCE, Attach/Detach operations have a 32-deep queue on VM. Added a workqueue for Attach/Detach calls with a rate limiter.

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -30,6 +30,8 @@ import (
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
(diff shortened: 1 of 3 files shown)


jingxu97:
Some scalability tests would be good to understand performance impact about this change

mattcary:
/ok-to-test

Offline discussion: the problems we've been seeing in practice occur when a particular node starts to error (ie, the per-vm operation queue in GCE fills up causing attach requests to immediately return with an error until the queue drains). So Julie's looking into making the rate limiting be more fine-grained, at least to the node level and maybe only on error as well.

We don't want, eg, scale-ups for large clusters to be unnecessarily rate-limited when the requests per node are not high.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -280,25 +296,107 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
...
+		return nil, err
+	}
+
+	// If the node is not marked, proceed the request
+	if _, found := gceCS.nodes[req.NodeId]; !found {
+		return gceCS.executeControllerPublishVolume(ctx, req)
+	}
+
+	// Node is marked so queue up the request
+	gceCS.queue.AddRateLimited(&workItem{
```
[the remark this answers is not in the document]
mattcary: Offline discussion: if we did something like one limiter per erroring node, then the number of potentially erroring API calls can increase like the number of erroring nodes, meaning that API quota can get exhausted faster. So having a single, global rate limiter actually makes the most sense from this perspective.

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1716,3 +1718,145 @@ func TestCreateVolumeDiskReady(t *testing.T) {
...
+
+		if tc.errorSeenOnNode {
+			if err := wait.Poll(10*time.Nanosecond, 10*time.Millisecond, func() (bool, error) {
+				t.Logf("%v items queueed", gceDriver.cs.queue.Len())
+				return gceDriver.cs.queue.Len() == requestCount, nil
+			}); err != nil {
+				t.Fatalf("Only %v requests queued up for node has seen error", gceDriver.cs.queue.Len())
+			}
+		}
+
```
mattcary, line 1790: If we wait a little longer, shouldn't the queue empty (because the fake provider will succeed the test?)

It's also possible that this test will be flaky. It would be more rigorous to use a fake rate limiter that could be controlled explicitly. I don't know if that's straightforward to do, so maybe we can see how this goes (although you may be creating more work for yourself in the long run...)

lizhuqi: The queue will be not empty as the request doesn't succeed so it will be re-queued. The reason why the request doesn't succeed is because I am using a real cloud provider so it will fail at getting volume lock step. The test case is to make sure when the request doesn't go through, all the requests failed and new requests will be queued. 

Maybe I can add more test cases using a fake cloud provider to have the request succeed but if that is the case, it is quite dynamic and async so nothing can be for sure.

lizhuqi: Added more tests

PR #865, opened 2021-11-06
Use correct images for alpha and dev builds

mattcary:
/kind failing-test

See failing migration tests, eg https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-gcp-compute-persistent-disk-csi-driver-latest-k8s-master-migration/1456750077849636864.

The migration tests use the dev overlay, which get images from alpha.

The alpha image overlay included some images apparently targeted at windows. However, metadata name doesn't mean anything, and the transformers changed the images on all versions. The csi driver image ends up being a no-op, because the tag is set to the latest build, but the node-driver-registrar change is effective. When the registrar was updated to 2.3.0, the liveness probe was added (and maybe other things); this is incompatible with the 2.1.0 version.

I don't think it's necessary to use special versions for windows at this point anyway. If we do, we'll have to change the image names in the kustomize base so that we can target windows nodes anyway.

/assign @jingxu97 
/assign @saikat-royc 

```release-note
None
```


saikat-royc:
the change looks good to me. Will let @jingxu97  lgtm it

PR #866, opened 2021-11-14
add CSI liveness probe container

zetaab:
**What type of PR is this?**

/kind bug

`-mode` parameter is not valid anymore to `csi-node-driver-registrar`. Also nowadays upstream uses separate container to run liveness probes using `k8s.gcr.io/sig-storage/livenessprobe` image

Fixes #

```release-note
Add CSI liveness probe container for CSI daemonset
```

### DIFF
--- a/deploy/kubernetes/base/node_linux/node.yaml
+++ b/deploy/kubernetes/base/node_linux/node.yaml
@@ -38,13 +38,13 @@ spec:
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
-          livenessProbe:
-            initialDelaySeconds: 3
-            exec:
-              command:
-                - /csi-node-driver-registrar
-                - --kubelet-registration-path=/var/lib/kubelet/plugins/pd.csi.storage.gke.io/csi.sock
-                - --mode=kubelet-registration-probe
+        - name: liveness-probe
+          image: k8s.gcr.io/sig-storage/livenessprobe
+          args:
+            - --csi-address=/csi/csi.sock
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: /csi
         - name: gce-pd-driver
           # Don't change base image without changing pdImagePlaceholder in
           # test/k8s-integration/main.go
--- a/deploy/kubernetes/images/stable-master/image.yaml
+++ b/deploy/kubernetes/images/stable-master/image.yaml
@@ -46,6 +46,16 @@ imageTag:
 apiVersion: builtin
 kind: ImageTagTransformer
 metadata:
+  name: imagetag-csi-liveness-probe
+imageTag:
+  name: k8s.gcr.io/sig-storage/livenessprobe
+  newTag: "v2.2.0"
+
+---
+
+apiVersion: builtin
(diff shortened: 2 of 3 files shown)


deploy/kubernetes/images/stable-master/image.yaml
```diff
@@ -43,6 +43,16 @@ imageTag:
   newTag: "v2.3.0"
 ---
 
+apiVersion: builtin
```
zetaab, line 46: I have no idea should this be added to different files as well in `images/` folder?

mattcary:
The mode for the node-driver-registrar was added [this summer](https://github.com/kubernetes-csi/node-driver-registrar/commit/58bd237058e91d74c20b3a2e1ee5ffcfb488aa5e#diff-8bb18ba5722e957285edbbd745096cf05020af17a0411ffea3c6cc1abc573e1d) and are in version [2.3.0](https://github.com/kubernetes-csi/node-driver-registrar/blob/master/CHANGELOG/CHANGELOG-2.3.md#bug-or-regression) and later. Are you using an older version of the registrar?

I don't know what our plans are for the csi liveness probe, @mauriciopoppe I feel like we've discussed this but I can't remember.

Also, changes like this should got through rc-staging first before stable-master.

deploy/kubernetes/base/node_linux/node.yaml
```diff
@@ -38,13 +38,13 @@ spec:
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
-          livenessProbe:
```
mattcary, line 41: I think we would want to leave this liveness probe, to check that registration is still good (the CSI liveness probe won't know about that).

deploy/kubernetes/base/node_linux/node.yaml
```diff
@@ -38,13 +38,13 @@ spec:
...
-            initialDelaySeconds: 3
-            exec:
-              command:
-                - /csi-node-driver-registrar
-                - --kubelet-registration-path=/var/lib/kubelet/plugins/pd.csi.storage.gke.io/csi.sock
-                - --mode=kubelet-registration-probe
+        - name: liveness-probe
```
mattcary, line 41: This needs an (http) liveness probe on the driver. However, since the pod is running hostNetwork, that consumes a host port which I don't think we want to do by default.

deploy/kubernetes/base/node_linux/node.yaml
mauriciopoppe: the `mode` flag is pretty recent with more background in https://github.com/kubernetes-csi/node-driver-registrar/issues/143, there are two checks that need to be done: checking that node-driver-registrar successfully registered the driver with the kubelet (done with a different running mode of node-driver-registrar) and once the CSI Driver is registered successfully check that the CSI Driver is healthy (done through https://github.com/kubernetes-csi/livenessprobe)

Q: isn't kubernetes-csi/livenessprobe enough?
Unfortunately livenessprobe only restarts the CSI Driver when there might be another issue in node-driver-registrar as described in the issue.

PR #869, opened 2021-11-19
pdcsi rox workflow fix

### LINKED ISSUE #872: Can't use ReadOnlyMany volume restored from Snapshot with controller-publish-readonly provisioner option
When using the new [`controller-publish-readonly` command-line option](https://github.com/kubernetes-csi/external-provisioner/pull/469) in the external provisioner sidecar, the ReadOnlyMany volumes seem to be still attempted to be mounted read-write. Moreover it looks like when the ROX volume is restored from snapshot, the kubelet still attempts to even create a filesystem on it.

Reproducer:
1. Create a CSI volume:
```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: myclaim1
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: standard-csi
  resources:
    requests:
      storage: 1Gi
```
2. Start a pod that writes some data on it (optionally, just to have something for verification).
3. Create a snapshot of the volume:
```
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshot
metadata:
  name: mysnap-1
spec:
  volumeSnapshotClassName: csi-gce-pd-vsc
  source:
    persistentVolumeClaimName: myclaim1
```
4. Restore the snapshot to ReadOnlyMany volume
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc1-restore
spec:
  accessModes:
  - ReadOnlyMany
  dataSource:
    apiGroup: snapshot.storage.k8s.io
    kind: VolumeSnapshot
    name: mysnap-1
  resources:
    requests:
      storage: 1Gi
  storageClassName: standard-csi
```
5. Try to use the restored volume in a pod:
```
kind: Pod
apiVersion: v1
metadata:
  name: pod-restore
  labels:
    name: pod-restore
spec:
  containers:
    - name: busybox
      image: gcr.io/google_containers/busybox
      command:
        - "/bin/sh"
        - "-c"
        - "while true; do sleep 1; done"
      volumeMounts:
      - mountPath: "/mnt/test"
        name: vol
        readOnly: true
  volumes:
    - name: vol
      persistentVolumeClaim:
        claimName: pvc1-restore
        readOnly: true
```

Result:
Pod fails to start with message similar to this one in the log:
```
Warning  FailedMount             16s (x7 over 51s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-e4269d3a-3a2d-4920-8917-a30c0a0773e7" : rpc error: code = Internal desc = Failed to format and mount device from ("/dev/disk/by-id/google-pvc-e4269d3a-3a2d-4920-8917-a30c0a0773e7") to ("/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-e4269d3a-3a2d-4920-8917-a30c0a0773e7/globalmount") with fstype ("ext4") and options ([]): format of disk "/dev/disk/by-id/google-pvc-e4269d3a-3a2d-4920-8917-a30c0a0773e7" failed: type:("ext4") target:("/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-e4269d3a-3a2d-4920-8917-a30c0a0773e7/globalmount") options:("defaults") errcode:(exit status 1) output:(mke2fs 1.45.6 (20-Mar-2020)
/dev/disk/by-id/google-pvc-e4269d3a-3a2d-4920-8917-a30c0a0773e7: Read-only file system while setting up superblock
```

This looks to be incorrect: the volume should not be attempted to be mounted read-write and the attempt to create a filesystem on a formatted drive pre-populated with data looks quite dangerous. I will try to tinker with this but if someone reading this had an idea, I'm interested. I'm also not sure who exactly is at fault here: might be also kubelet.

leiyiz:
turn on controller-publish-readonly flag  and add validation in pd-csi driver for when readonly is on

if mounting a read-only ext3 or ext4 disk, use `noload` to prevent mount failure caused by journal replay 

**What type of PR is this?**
/kind feature

when ROX accessMode is specified, csi.Readonly is currently not set. this PR fix that, and at the same time validate that a content source is set when readonly modes are specified at provision time.

Fixes #872

```release-note
Dynamically provisioning for Read only PVs will fail now unless a volume source (either restore from snapshot or volume cloning) is specified. 
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -60,6 +60,7 @@ const (
 	volumeLimitBig       int64 = 127
 	defaultLinuxFsType         = "ext4"
 	defaultWindowsFsType       = "ntfs"
+	fsTypeExt3                 = "ext3"
 )
 
 func getDefaultFsType() string {
@@ -311,8 +312,29 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
+	readonly, _ := getReadOnlyFromCapability(volumeCapability)
+	if readonly {
+		options = append(options, "ro")
+		klog.V(4).Infof("CSI volume is read-only, mounting with extra option ro")
+	}
+
 	err = formatAndMount(devicePath, stagingTargetPath, fstype, options, ns.Mounter)
 	if err != nil {
+		// If a volume is created from a content source like snapshot or cloning, the filesystem might get marked
+		// as "dirty" even if it is otherwise consistent and ext3/4 will try to restore to a consistent state by replaying
+		// the journal which is not possible in read-only mode. So we'll try again with noload option to skip it. This may
+		// allow mounting of an actually inconsistent filesystem, but because the mount is read-only no further damage should
+		// be caused.
+		if readonly && (fstype == defaultLinuxFsType || fstype == fsTypeExt3) {
+			klog.V(4).Infof("Failed to mount CSI volume read-only, retry mounting with extra option noload")
+
+			options = append(options, "noload")
+			err = formatAndMount(devicePath, stagingTargetPath, fstype, options, ns.Mounter)
+			if err == nil {
+				klog.V(4).Infof("NodeStageVolume succeeded with \"noload\" option on %v to %s", volumeID, stagingTargetPath)
+				return &csi.NodeStageVolumeResponse{}, nil
+			}
+		}
 		return nil, status.Error(codes.Internal,
 			fmt.Sprintf("Failed to format and mount device from (%q) to (%q) with fstype (%q) and options (%q): %v",
 				devicePath, stagingTargetPath, fstype, options, err))
--- a/deploy/kubernetes/images/alpha/image.yaml
+++ b/deploy/kubernetes/images/alpha/image.yaml
@@ -1,0 +1,7 @@
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-csi-provisioner-alpha
+imageTag:
+  name: k8s.gcr.io/sig-storage/csi-provisioner
+  newTag: "v3.0.0"
--- a/deploy/kubernetes/images/alpha/kustomization.yaml
+++ b/deploy/kubernetes/images/alpha/kustomization.yaml
@@ -2,3 +2,4 @@ namespace:
   gce-pd-csi-driver
 resources:
 - ../stable-master/
+- image.yaml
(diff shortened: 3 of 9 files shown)


deploy/kubernetes/images/stable-master/image.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: imagetag-csi-provisioner
 imageTag:
   name: k8s.gcr.io/sig-storage/csi-provisioner
-  newTag: "v2.2.1"
```
mattcary, line 7: Update the images in the alpha or dev overlay first, and let it run through the postsubmits, we'll update stablemaster only after it's validated.

leiyiz: done

leiyiz:
> either don't have filesystem copied over, or filesystem is not correctly recognized by mounter. hold until we figure things out more

actually it's not issue with filesystem not there, mounter is not trying to setup a new filesystem. the mount failed at mount step with `cannot mount /dev/sdb read-only.`

pkg/gce-pd-csi-driver/node.go
```diff
@@ -311,8 +312,25 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
 
+	readonly, _ := getReadOnlyFromCapability(volumeCapability)
+	if readonly {
+		options = append(options, "ro")
+		klog.V(4).Infof("CSI volume is read-only, mounting with extra option ro")
+	}
+
 	err = formatAndMount(devicePath, stagingTargetPath, fstype, options, ns.Mounter)
 	if err != nil {
+		// ext4 and ext3 replays the journal which needs write permission. So noload option stops that
```
mattcary, line 323: Let's be a bit more precise in the comment. Isn't it more precise to say that if the volume was created by cloning or from a snapshot, the journal might not be in a consistent state even if the fs is otherwise fine. ext3/4 will try to fix that by reloading the journal, which is not possible if the fs is read-only, and which is skipped by noload. We should also mention that this may allow mounting of a corrupted filesystem, but because the mount is read-only no (further) harm will happen.

leiyiz: totally agreed

mattcary:
Can you enumerate what overlays we use in upstream testing (both presubmit & periodic) & internally?

I'm wondering if we want to change stable-master now, or wait until this bakes, but I forget which test jobs use which overlay.

leiyiz:
Internally most prow tests are using `stable-master`, a few are using `prow-gke-release-staging-head` and `prow-gke-release-staging-rc-master`.

Upstream postsubmits are using `stable-master` and `dev` overlays

mattcary:
/lgtm

Okay. Let's let this run through the postsubmits using `dev`, then in a day or two when we see things are green you can push a PR to update `stable-master`?

leiyiz:
migration testgrid is red but the [driver latest](https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver#Kubernetes%20Master%20Driver%20Latest) grid looks green with csi-provisioner v3.0.0 picked up.

PR #870, opened 2021-11-22
Change to distroless base image

amacaskill:
**What type of PR is this?**
/kind feature

We are changing pdcsi driver to be built from distroless base image instead of debian. Distroless images are much smaller as they only contain your application and its runtime dependencies. Restricting what's in your runtime container is a best practice, and it improves the signal to noise of scanners (e.g. CVE). See [distroless](https://github.com/GoogleContainerTools/distroless) for more information. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,15 +23,44 @@ WORKDIR /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 ADD . .
 RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=$STAGINGVERSION make gce-pd-driver
 
-# MAD HACKS: Build a version first so we can take the scsi_id bin and put it somewhere else in our real build
-FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as mad-hack
-RUN clean-install udev
-
-# Start from Kubernetes Debian base
-FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0
-COPY --from=builder /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/bin/gce-pd-csi-driver /gce-pd-csi-driver
+# Start from Kubernetes Debian base.
+FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as debian
 # Install necessary dependencies
 RUN clean-install util-linux e2fsprogs mount ca-certificates udev xfsprogs
-COPY --from=mad-hack /lib/udev/scsi_id /lib/udev_containerized/scsi_id
+# Since we're leveraging apt to pull in dependencies, we use `gcr.io/distroless/base` because it includes glibc.
+FROM gcr.io/distroless/base-debian11
+# Copy necessary dependencies into distroless base.
+COPY --from=builder /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/bin/gce-pd-csi-driver /gce-pd-csi-driver
+COPY --from=debian /etc/mke2fs.conf /etc/mke2fs.conf
+COPY --from=debian /lib/udev/scsi_id /lib/udev_containerized/scsi_id
+COPY --from=debian /bin/mount /bin/mount
+COPY --from=debian /bin/umount /bin/umount
+COPY --from=debian /sbin/blkid /sbin/blkid
+COPY --from=debian /sbin/blockdev /sbin/blockdev
+COPY --from=debian /sbin/dumpe2fs /sbin/dumpe2fs
+COPY --from=debian /sbin/e* /sbin/
+COPY --from=debian /sbin/e2fsck /sbin/e2fsck
+COPY --from=debian /sbin/fsck /sbin/fsck
+COPY --from=debian /sbin/fsck* /sbin/
+COPY --from=debian /sbin/fsck.xfs /sbin/fsck.xfs
+COPY --from=debian /sbin/mke2fs /sbin/mke2fs
+COPY --from=debian /sbin/mkfs* /sbin/
+COPY --from=debian /sbin/resize2fs /sbin/resize2fs
+COPY --from=debian /sbin/xfs_repair /sbin/xfs_repair
+COPY --from=debian /usr/include/xfs /usr/include/xfs
+COPY --from=debian /usr/lib/xfsprogs/xfs* /usr/lib/xfsprogs/
+COPY --from=debian /usr/sbin/xfs* /usr/sbin/
+
+# Copy x86 shared libraries into distroless base.
+COPY --from=debian /lib/x86_64-linux-gnu/libblkid.so.1 /lib/x86_64-linux-gnu/libblkid.so.1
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -25,13 +26,52 @@ RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=$
...
+COPY --from=debian /sbin/fsck* /sbin/
+COPY --from=debian /sbin/e2fsck /sbin/e2fsck
+COPY --from=debian /sbin/xfs_repair /sbin/xfs_repair
+COPY --from=debian /usr/sbin/xfs* /usr/sbin/
+COPY --from=debian /usr/include/xfs /usr/include/xfs
+COPY --from=debian /usr/lib/xfsprogs/xfs* /usr/lib/xfsprogs/
+COPY --from=debian /usr/include/xfs /usr/include/xfs
+COPY --from=debian /sbin/blkid /sbin/blkid
+# Copy shared libraries into distoless base.
+COPY --from=debian /lib/x86_64-linux-gnu/libselinux.so.1 /lib/x86_64-linux-gnu/libselinux.so.1
```
mauriciopoppe, line 59: these shared libraries are compiled for x86 from the prefix and I think that the multidistro build also builds for arm

ref: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/bc88549c3cdc1f5881af5992d04a1e13ae314def/Makefile#L98-L102

mattcary: Interesting point. Maybe we can use the docker $BUILDPLATFORM or something like that?

I wonder if this sort of thing has been done before, maybe there's some common advice for dealing with these issues? eg, the debian base that we're pulling from presumably would be arm when such a build.

Or we just use a different dockerfile for arm, like we do for windows.

amacaskill: I decided to go with a separate dockerfile for arm as opposed to checking if the environment variable TARGETPLATFORM = "arm64" because conditionals in docker (e.g. `RUN if [ "$TARGETPLATFORM" = "arm64"]; then ...`) uses bash. Since our distroless image does not have shell, we would have to use a distroless base image that has shell, but the bash image contains a much larger potential attack surface than the other distroless images. I thought about doing something like making a new folder in the debian image x86_64-linux-gnu-containerized then moving all the content in /lib/x86_64-linux-gnu/ to that folder if the TARGETPLATFORM is "arm64" since the debian image has bash. Then in the distroless image we always copy from the x86_64-linux-gnu-containerized  folder, but the docker build will fail if the COPY doesn't COPY over at least one file, which would be the case for the arm64.

Dockerfile
```diff
@@ -23,15 +23,46 @@ WORKDIR /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 ADD . .
 RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=$STAGINGVERSION make gce-pd-driver
 
-# MAD HACKS: Build a version first so we can take the scsi_id bin and put it somewhere else in our real build
+# MAD HACKS: Build a version first so we can take the scsi_id bin and put it somewhere else in our real build.
 FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as mad-hack
```
mauriciopoppe, line 27: I think these two can be merged now:

```
FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as mad-hack
RUN clean-install udev
FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as debian
RUN clean-install util-linux e2fsprogs mount ca-certificates udev xfsprogs
```

Dockerfile.arm64
```diff
@@ -0,0 +1,54 @@
+# Copyright 2018 The Kubernetes Authors.
```
mauriciopoppe, line 1: should be 2021

Dockerfile.arm64
```diff
@@ -0,0 +1,54 @@
...
+COPY --from=debian /sbin/fsck* /sbin/
+COPY --from=debian /sbin/fsck.xfs /sbin/fsck.xfs
+COPY --from=debian /sbin/mke2fs /sbin/mke2fs
+COPY --from=debian /sbin/mkfs* /sbin/
+COPY --from=debian /sbin/resize2fs /sbin/resize2fs
+COPY --from=debian /sbin/xfs_repair /sbin/xfs_repair
+COPY --from=debian /usr/include/xfs /usr/include/xfs
+COPY --from=debian /usr/lib/xfsprogs/xfs* /usr/lib/xfsprogs/
+COPY --from=debian /usr/sbin/xfs* /usr/sbin/
+
```
mattcary, line 53: why does arm not need to copy libraries---is everything static?

If so, let's put a comment in here about that.

amacaskill: Discussed offline, there isn't really a way to test distroless on arm until GCE has arm hardware, so until distroless on arm can be tested, we will leave the Dockerfile.arm64 to use the debian base image.

Dockerfile.arm64
```diff
@@ -0,0 +1,37 @@
...
+ARG STAGINGVERSION
+ARG TARGETPLATFORM
+
+WORKDIR /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
+ADD . .
+RUN GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=$STAGINGVERSION make gce-pd-driver
+
+# MAD HACKS: Build a version first so we can take the scsi_id bin and put it somewhere else in our real build
+FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as mad-hack
```
mauriciopoppe, line 27: mad-hack and `k8s.gcr.io/build-image/debian-base:buster-v1.9.0` below could be merged here too

amacaskill: Why can we merge it? Was this logic only for x86?

mauriciopoppe: oh, I meant the sequence:

```
FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0 as mad-hack
RUN clean-install udev
FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0
COPY --from=builder /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/bin/gce-pd-csi-driver /gce-pd-csi-driver
RUN clean-install util-linux e2fsprogs mount ca-certificates udev xfsprogs
COPY --from=mad-hack /lib/udev/scsi_id /lib/udev_containerized/scsi_id
```

is the same as:

```
FROM k8s.gcr.io/build-image/debian-base:buster-v1.9.0
RUN clean-install util-linux e2fsprogs mount ca-certificates udev xfsprogs

COPY --from=builder /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/bin/gce-pd-csi-driver /gce-pd-csi-driver
COPY /lib/udev/scsi_id /lib/udev_containerized/scsi_id
```

mauriciopoppe: Discussed offline, not a blocker

PR #871, opened 2021-11-23
Log operation id for attach and detach

amacaskill:
**What type of PR is this?**
/kind bug

This PR logs operation id for each PD disk operations attempt in order to help debug attach/detach errors by making it more clear which retry attempt of the driver the error was part of. Logging the operation id in the driver helps map the audit, and the driver CSI call execution.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -441,6 +441,7 @@ func (cloud *CloudProvider) insertRegionalDisk(
 		}
 		return status.Error(codes.Internal, fmt.Sprintf("unknown Insert disk error: %v", err))
 	}
+	klog.V(5).Infof("InsertDisk operation %s for disk %s", opName, diskToCreate.Name)
 
 	err = cloud.waitForRegionalOp(ctx, project, opName, volKey.Region)
 	if err != nil {
@@ -536,6 +537,7 @@ func (cloud *CloudProvider) insertZonalDisk(
 		}
 		return fmt.Errorf("unknown Insert disk error: %v", err)
 	}
+	klog.V(5).Infof("InsertDisk operation %s for disk %s", opName, diskToCreate.Name)
 
 	err = cloud.waitForZonalOp(ctx, project, opName, volKey.Zone)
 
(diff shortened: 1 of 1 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -621,6 +631,9 @@ func (cloud *CloudProvider) AttachDisk(ctx context.Context, project string, volK
 	}
 
 	op, err := cloud.service.Instances.AttachDisk(project, instanceZone, instanceName, attachedDiskV1).Context(ctx).Do()
+	if op != nil {
```
saikat-royc, line 634: let's move this block after the err != nil checks. same everywhere else.

amacaskill: By moving the logging statement to after the err != nil checks, the if op != nil is no longer necessary, so I removed that.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -461,6 +461,7 @@ func (cloud *CloudProvider) insertRegionalDisk(
 		}
 		return fmt.Errorf("unknown Insert disk operation error: %v", err)
 	}
+	klog.V(5).Infof("InsertDisk operation %s for disk %s", opName, diskToCreate.Name)
```
saikat-royc, line 464: sorry if I was not clear earlier. the error I meant was in line 421. So basically before we start waiting for the op in line 445. i.e put this log in line 444. That way we know what op was initiated, and check in cases of scenarios like op taking long time

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -557,6 +558,7 @@ func (cloud *CloudProvider) insertZonalDisk(
...
```
saikat-royc, line 561: same here. move this to line line 540.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -585,6 +587,7 @@ func (cloud *CloudProvider) deleteZonalDisk(ctx context.Context, project, zone,
 	if err != nil {
 		return err
 	}
+	klog.V(5).Infof("DeleteDisk operation %s for disk %s", op.Name, name)
```
saikat-royc, line 590: move to line 586.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -601,6 +604,7 @@ func (cloud *CloudProvider) deleteRegionalDisk(ctx context.Context, project, reg
...
```
saikat-royc, line 607: same for other places

pkg/gce-cloud-provider/compute/gce-compute.go
amacaskill: Ah ok got it.

PR #876, opened 2021-12-10
Add guest-configs to go.mod

lizhuqi:
**What type of PR is this?**
/kind feature

To support symlink validation for NVMe, we need the google_nvme_id scripts and rules from GCP guest configs repro. This PR is to add the guest configs repro as dependency. Will upload the script and rules to pd csi driver image in next PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 )
 
 require (
+	github.com/GoogleCloudPlatform/guest-configs v0.0.0-20211028171426-a0dacef88470 // indirect
 	github.com/Microsoft/go-winio v0.4.16 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/blang/semver v3.5.1+incompatible // indirect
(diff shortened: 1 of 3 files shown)


mauriciopoppe:
hey @lizhuqi, how is this dependency used in the driver?

lizhuqi:
> hey @lizhuqi, how is this dependency used in the driver?

Hi @mauriciopoppe , I plan to use the google_nvme_id script in the guest config repro to fetch unique ID of a NVMe disk for the udev issue validation.

mauriciopoppe:
I see, if you want this file https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/google_nvme_id I think it should be added in this PR after running `go mod vendor` with `git add --force vendor`

lizhuqi:
> I see, if you want this file https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/google_nvme_id I think it should be added in this PR after running `go mod vendor` with `git add --force vendor`

I just ran 'go mod vendor' and 'git add --force vendor' but no changes happens.

mauriciopoppe:
> I just ran 'go mod vendor' and 'git add --force vendor' but no changes happens.

I read that `go mod vendor` might not be pulling non-go files https://github.com/golang/go/issues/26366#issuecomment-405683150, if you need the script in the image then you might need to add it to the image as a static file copied from the Dockerfile

lizhuqi:
> > I just ran 'go mod vendor' and 'git add --force vendor' but no changes happens.
> 
> I read that `go mod vendor` might not be pulling non-go files [golang/go#26366 (comment)](https://github.com/golang/go/issues/26366#issuecomment-405683150), if you need the script in the image then you might need to add it to the image as a static file copied from the Dockerfile

I see. Thanks!

PR #879, opened 2021-12-15
Add CREATE_SA_KEY flag to deploy/setup-project.sh

### LINKED ISSUE #639: setup-project.sh should skip keys creation step for noauth overlay
In https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/setup-project.sh,
gcloud iam service-accounts keys create step should be skipped when using a noauth overlay.

pwschuurman:
**What type of PR is this?**
/kind cleanup

This adds a flag to deploy/setup-project.sh which skips creating SA keys. This is useful when setting up a project when using the noauth overlay, and creating SA keys is not necessary.

Fixes #639

N/A

```release-note
NONE
```


mattcary:
thanks!

PR #890, opened 2022-01-11
add create regional clone from zonal disk and improve tests

amacaskill:
**What type of PR is this?**
> /kind feature

GCE recently added [creating a regional disk from a zonal disk](https://cloud.google.com/compute/docs/disks/create-disk-from-source#create-regional-clone). This PR adds this feature to volume cloning for the pd csi driver. 
```release-note
Lets users clone a regional disk from a zonal disk if one of the replica zones of the clone matchesthe zone of the source disk.
```


test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1011,6 +1011,64 @@ var _ = Describe("GCE PD CSI Driver", func() {
 			Expect(gce.IsGCEError(err, "notFound")).To(BeTrue(), "Expected snapshot to not be found")
 		}()
 	})
+
+	It("Should successfully create RePD from a zonal PD", func() {
```
saikat-royc, line 1015: I think we should also add a zonal to zonal disk clone case?

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -199,3 +199,13 @@ func collectMountOptions(fsType string, mntFlags []string) []string {
 	}
 	return options
 }
+
+func containsZone(s []string, str string) bool {
```
saikat-royc, line 203: zones []string, zone string

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -243,10 +243,34 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
+
+			// Verify the disk type and encryption key of the clone are the same as that of the source disk.
+			if diskFromSourceVolume.GetPDType() != params.DiskType || !gce.KmsKeyEqual(diskFromSourceVolume.GetKMSKeyName(), params.DiskEncryptionKMSKey) {
+				return nil, status.Errorf(codes.InvalidArgument, `CreateVolume source volume Parameters do not match CreateVolumeRequest Parameters %v`, params)
+			}
+
+			if params.ReplicationType == replicationTypeNone {
+				// For zonal->zonal disk clones, verify the zone is the same as that of the source disk.
+				if sourceVolKey.Zone != volKey.Zone {
+					return nil, status.Errorf(codes.InvalidArgument, `CreateVolume source volume zone %s does not match CreateVolume disk zone %s`, sourceVolKey.Zone, volKey.Zone)
```
saikat-royc, line 255: "CreateVolume disk zone %s does not match source volume zone %s" may be a better wording.
same for other places

test/e2e/tests/single_zone_e2e_test.go
amacaskill: Done

pkg/gce-pd-csi-driver/controller.go
amacaskill: Done

Issue #892, opened 2022-01-11
CI tests aren't using the e2e.test binary compiled from the latest kubernetes master codebase


mauriciopoppe:
We're trying to fix a flaky VolumeSnapshot test, after we reordered some of the test statements in https://github.com/kubernetes/kubernetes/pull/107173 we checked [testgrid](https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver#Kubernetes%20Master%20Driver%20Latest) and saw that the test was still flaky (for example [this run](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-gcp-compute-persistent-disk-csi-driver-latest-k8s-master-integration/1480761349897195520)), moreover, after analyzing the logs we saw that it was logging the statements in a different order than the latest in Kubernetes master, we found out the following:

- the `k8s-integration` binary downloads the latest code of Kubernetes master and calls `make quick-release` [ref](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/35ab61ad92908178cf02c111936278b2c33d25f8/test/k8s-integration/main.go#L278), in the logs we saw that `e2e.test` was generated.
- kube-up is called with these binaries
- after the CSI Driver is installed it looks like another release of kubernetes is downloaded again, we suspect that the `e2e.test` binary taken is from this release and not the one that was recently compiled

Logs for the last point:

```
I0111 05:09:49.697] Running Tests
I0111 05:09:49.697] [kubetest2 gce --test=ginkgo --legacy-mode --repo-root=/tmp/gcp-pd-driver-tmp2860208485/kubernetes --artifacts=/workspace/_artifacts/sc-balanced -- --focus-regex=External.Storage --skip-regex=\[Disruptive\]|\[Serial\] --parallel=4 --test-args=--storage.testdriver=/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/test/k8s-integration/config/test-config.yaml ]
W0111 05:09:51.890] Copying gs://kubernetes-release/release/v1.24.0-alpha.1/kubernetes-test-linux-amd64.tar.gz...
W0111 05:09:51.892] / [0 files][    0.0 B/250.5 MiB]                                                
==> NOTE: You are downloading one or more large file(s), which would
W0111 05:09:51.892] run significantly faster if you enabled sliced object downloads. This
[...]

PR #893, opened 2022-01-12
Use the compiled kubernetes binaries in the e2e tests in CI

mauriciopoppe:
**What type of PR is this?**
>
/kind bug

When the kubetest2 flags are created it'll detect if we're trying to run the e2e test from master, if so then it'll assume that the cluster was created by compiling a kubernetes release or kubernetes master and that the binaries to use in the e2e tests should be the ones generated during the compilation.

Might help https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/892

```release-note
NONE
```

/cc @mattcary @jingxu97

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"syscall"
 
@@ -662,15 +663,19 @@ func runTestsWithConfig(testParams *testParameters, testConfigArg, reportPrefix 
 
 	testArgs := fmt.Sprintf("%s %s", ginkgoArgs, testConfigArg)
 
-	kubeTestArgs := []string{
-		"--test",
-		"--ginkgo-parallel",
-		"--check-version-skew=false",
-		fmt.Sprintf("--test_args=%s", testArgs),
+	// kubetest2 flags
+
+	var runID string
+	if uid, exists := os.LookupEnv("PROW_JOB_ID"); exists && uid != "" {
+		// reuse uid for CI use cases
+		runID = uid
+	} else {
+		runID = string(uuid.NewUUID())
 	}
 
 	kubeTest2Args := []string{
 		*deploymentStrat,
+		fmt.Sprintf("--run-id=%s", runID),
 		"--test=ginkgo",
 	}
 	kubeTest2Args = append(kubeTest2Args, testParams.cloudProviderArgs...)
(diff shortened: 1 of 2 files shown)


test/k8s-integration/main.go
```diff
@@ -678,14 +673,33 @@ func runTestsWithConfig(testParams *testParameters, testConfigArg, reportPrefix
...
 	}
 	kubeTest2Args = append(kubeTest2Args, "--")
-	if len(*testVersion) != 0 && *testVersion != "master" {
-		kubeTest2Args = append(kubeTest2Args, fmt.Sprintf("--test-package-marker=latest-%s.txt", *testVersion))
+	if len(*testVersion) != 0 {
+		if *testVersion == "master" {
+			// the kubernetes binaries should've already been built above
+			// or by the user if --localK8sDir was set, these binaries should be copied to the
+			// path sent to kubetest2 through its --artifacts path
+			klog.Infof("Copying kubernetes binaries to path=%s to run the tests", kubetestDumpDir)
```
mattcary, line 681: kubetestdumpDir may be empty, in that case maybe it would be better to set it to ./? (and will it work in this case, or have we implicitly assumed an artifactsDir if we're testing master?)

mattcary:
Looks like something is awry with the directories or copying? The relevant integration test error I think is this:

```
W0112 01:36:19.231] I0112 01:36:19.053936   91423 app.go:61] RunDir for this run: "/workspace/_artifacts/sc-standard/ba403430-7343-11ec-94ac-7eb5bc905a97"
W0112 01:36:19.231] I0112 01:36:19.058854   91423 app.go:120] ID for this run: "ba403430-7343-11ec-94ac-7eb5bc905a97"
W0112 01:36:19.232] I0112 01:36:19.107699   91437 ginkgo.go:120] Using kubeconfig at /root/.kube/config
W0112 01:36:19.232] F0112 01:36:19.107813   91437 ginkgo.go:205] failed to run ginkgo tester: failed to validate kubectl:stat /workspace/_artifacts/sc-standard/ba403430-7343-11ec-94ac-7eb5bc905a97/kubectl: no such file or directory
W0112 01:36:19.232] Error: exit status 255
```

mauriciopoppe:
I see that the problem is that `kubetest2` appends a run id, so while the binaries are copied to `W0112 01:36:17.362] I0112 01:36:17.361583    6663 main.go:736] copying /tmp/gcp-pd-driver-tmp2296392321/kubernetes/_output/dockerized/bin/linux/amd64/kubectl to /workspace/_artifacts/sc-standard/kubectl
` the binary that it runs is `/workspace/_artifacts/sc-standard/ba403430-7343-11ec-94ac-7eb5bc905a97/kubectl`

test/k8s-integration/main.go
mauriciopoppe: thanks for catching this, I saw that if we don't set `--artifacts` then kubetest2 assumes that it's `./_artifacts`, from its help:

```
Flags:
      --artifacts string         top-level directory to put artifacts under for each kubetest2 run, defaulting to "${ARTIFACTS:-./_artifacts}". If using the ginkgo tester, this must be an absolute path. (default "/usr/local/google/home/mauriciopoppe/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/_artifacts")
```

I modified it to be that path if `kubetestdumpDir` is empty

mauriciopoppe:
@mattcary I'm using the `--run-id` flag that is now set from PD CSI if it's not already set by the CI job, with it I could also write the test binaries to the right location

mattcary:
> @mattcary I'm using the `--run-id` flag that is now set from PD CSI if it's not already set by the CI job, with it I could also write the test binaries to the right location

When in doubt of kubetest2 add more flags I guess... :-P

/lgtm
/approve

PR #900, opened 2022-01-18
add nvme script and rules as static files

lizhuqi:
**What type of PR is this?**
/kind feature

To support NVME, we need the script and rules uploaded to docker files. Add those files as static files for now before we figure out git subtree.

Fixes #

```release-note
None
```


mattcary (commented):
Can you add a section to the readme that explains these files and (most importantly) which repo they came from, with the repo version, so we know how to update them?

mauriciopoppe (commented):
These files aren't used in the image, would adding these lines in the Dockerfile work?

```
# or from raw github.com
ADD https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/google_nvme_id /path/inside/the/image
```

mattcary:
> These files aren't used in the image, would adding these lines in the Dockerfile work?
> 
> ```
> # or from raw github.com
> ADD https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/google_nvme_id /path/inside/the/image
> ```

We can't pull stuff off of the internet during our build process--it violates new security guidelines.

We need to have everything hermetic inside the repo at build time (eg, vendered).

lizhuqi:
> These files aren't used in the image, would adding these lines in the Dockerfile work?
> 
> ```
> # or from raw github.com
> ADD https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/lib/udev/google_nvme_id /path/inside/the/image
> ```

Yeah, as Matt mentioned, if we download raw file or git clone, it goes against the security guidelines. We eventually will use git subtree but before that is set up, we will checkin the files as the static first.

lizhuqi:
> Can you add a section to the readme that explains these files and (most importantly) which repo they came from, with the repo version, so we know how to update them?

sure. I will check in a script for downloading and copying the files

deploy/kubernetes/google_nvme_id
```diff
@@ -0,0 +1,245 @@
+#!/bin/bash
```
mauriciopoppe, line 1: Understood thanks, are these files going to be added in the image? I didn't see any changes to the Dockefile files

mattcary: @lizhuqi I think it would be fine to add the dockerfile change to this PR too unless you think there's a reason to keep it separate? Having it in the same change does self-document how these are used.

lizhuqi: oh I prefer to have another PR for the dockerfile change.

lizhuqi: I am also fine to add the dockerfile change as well. I will add the dockerfile change to this PR.

mattcary: 👍🏾

mauriciopoppe (commented):
/lgtm

Thanks for `deploy/kubernetes/update-nvme.sh`, makes it clear what's downloaded. Out of curiousity how is google_nvme_id going to be used? It's a distroless image so it doesn't have a shell, are you going to do something like `kubectl run NAME --image=<pdcsi image> --command -- <cmd> <arg1> ... <argN>` and use `google_nvme_id` as the command?

mattcary:
/hold

Waiting on removing the udev rules which aren't needed.

lizhuqi:
> /lgtm
> 
> Thanks for `deploy/kubernetes/update-nvme.sh`, makes it clear what's downloaded. Out of curiousity how is google_nvme_id going to be used? It's a distroless image so it doesn't have a shell, are you going to do something like `kubectl run NAME --image=<pdcsi image> --command -- <cmd> <arg1> ... <argN>` and use `google_nvme_id` as the command?

I plan to have it trigger by executeCommand following the existing scsi_id one:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/82070d808371d6cca9d18fdd6cbacf4b559cf698/pkg/mount-manager/device-utils.go#L121

deploy/kubernetes/update-nvme.sh
```diff
@@ -0,0 +1,26 @@
...
+# and copy script and rules required for NVMe support to be under deploy/kubernetes 
+#
+# Args:
+# GUEST_CONFIGS_VERSION: The version of the guest configs release
+
+set -o nounset
+set -o errexit
+
+readonly PKGDIR="${GOPATH}/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver"
+readonly INSTALL_DIR="${PKGDIR}/tmp"
```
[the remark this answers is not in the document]
mattcary: As written has the advantage of of only creating one directory in case it's not cleaned up, not so much clutter.

But I'm fine either way.

PR #903, opened 2022-01-26
Run the k8s-integration-test binary without eval

mauriciopoppe:
**What type of PR is this?**
>
/kind bug

The CI scripts for Windows integration passed! However I saw that the migration job failed, I think we should run the k8s-integration-test binary without eval, hopefully this is the last PR about making both Linux and Windows build strategies to be the same.

```release-note
NONE
```

/cc @mattcary @jingxu97


mattcary:
/lgtm
/approve

I'm hoping too :-)

PR #904, opened 2022-01-27
Use double quotes to expand variables in the windows migration e2e script

mauriciopoppe:
**What type of PR is this?**
>
/kind bug

`--kube-feature-gates=CSIMigration=true,CSIMigrationGCE=true,ExpandCSIVolumes=true` was stripped and understood by the binary as `KUBE_FEATURE_GATES=CSIMigration=true`, using double quotes to expand variables in the windows migration e2e script

Sorry to keep coming back with more PRs

```release-note
NONE
```

/cc @mattcary


mattcary:
bash strikes again!

/lgtm
/approve

PR #908, opened 2022-02-01
Promote release 1.4.0 to prow staging test

mattcary:
/kind cleanup

/assign @saikat-royc 

```release-note
None
```


mattcary:
@saikat-royc Given the time between cutting the release and promoting it, I wonder if in the future we should just promote the prow image directly to the prod one and save ourselves a PR (at the risk of having a failing staging prow job, which would probably disappear off the test grid by the team we pushed the next release).

PR #910, opened 2022-02-02
Update staging with new sidecar versions

mattcary:
/kind cleanup

Update staging with new sidecar versions. After the [staging prow](https://testgrid.corp.google.com/gcp-compute-persistent-disk-csi-driver#pd-master-sidecars-staging-k8s-master-on-gce) (internal google only) is green, I'll push this to stable-master with a release note.

/assign @saikat-royc 

```release-note
None
```


leiyiz:
the sidecar versions in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/images/prow-stable-sidecar-rc-master/image.yaml are all equal or newer than those mentioned in this PR. should be closed.

PR #929, opened 2022-03-16
Add e2e/integration tests for image snapshot

### LINKED ISSUE #924: Support disk image snapshots
I believe the right documentation is  https://cloud.google.com/compute/docs/images/create-delete-deprecate-private-images for creating, and https://cloud.google.com/compute/docs/disks/create-disk-from-source.

To create the disk image, a parameter could be added to the VolumeSnapshotClass. For restoring, the existing volume cloning support (create from VolumeSource) may just work, or additional plumbing may be required.

It's also necessary to look into the zonal vs regional behavior of disk images and see if anything needs to be done around that.

luohao:
**What type of PR is this?**

/kind feature

Fixes #924 

```release-note
N/A
```

### DIFF
--- a/test/e2e/tests/single_zone_e2e_test.go
+++ b/test/e2e/tests/single_zone_e2e_test.go
@@ -946,6 +946,65 @@ var _ = Describe("GCE PD CSI Driver", func() {
 			Expect(gce.IsGCEError(err, "notFound")).To(BeTrue(), "Expected snapshot to not be found")
 		}()
 	})
+
+	// Use the region of the test location.
+	It("Should successfully create snapshot backed by disk image", func() {
+		testContext := getRandomTestContext()
+
+		p, z, _ := testContext.Instance.GetIdentity()
+		client := testContext.Client
+
+		// Create Disk
+		volName, volID := createAndValidateUniqueZonalDisk(client, p, z)
+
+		// Create Snapshot
+		snapshotName := testNamePrefix + string(uuid.NewUUID())
+		testImageFamily := "test-family"
+
+		snapshotParams := map[string]string{common.ParameterKeySnapshotType: common.DiskImageType, common.ParameterKeyImageFamily: testImageFamily}
+		snapshotID, err := client.CreateSnapshot(snapshotName, volID, snapshotParams)
+		Expect(err).To(BeNil(), "CreateSnapshot failed with error: %v", err)
+
+		// Validate Snapshot Created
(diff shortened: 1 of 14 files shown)


test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -876,6 +876,65 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+		snapshotParams := map[string]string{common.ParameterKeySnapshotType: common.DiskImageType, common.ParameterKeyImageFamily: testImageFamily}
+		snapshotID, err := client.CreateSnapshot(snapshotName, volID, snapshotParams)
+		Expect(err).To(BeNil(), "CreateSnapshot failed with error: %v", err)
+
+		// Validate Snapshot Created
+		snapshot, err := computeService.Images.Get(p, snapshotName).Do()
+		Expect(err).To(BeNil(), "Could not get snapshot from cloud directly")
+		Expect(snapshot.Name).To(Equal(snapshotName))
+
+		err = wait.Poll(10*time.Second, 3*time.Minute, func() (bool, error) {
```
mattcary, line 903: I think you've seen disk image creation take up to 4 minutes? Maybe should set timeout to 5 min to decrease test flakiness.

test/run-k8s-integration-ci.sh
```diff
@@ -29,7 +29,7 @@ readonly run_intree_plugin_tests=${RUN_INTREE_PLUGIN_TESTS:-false}
 readonly use_kubetest2=${USE_KUBETEST2:-true}
 readonly test_pd_labels=${TEST_PD_LABELS:-true}
 readonly migration_test=${MIGRATION_TEST:-false}
-readonly test_disk_image_snapshot=${TEST_DISK_IMAGE_SNAPSHOT:-false}
+readonly test_disk_image_snapshot=${TEST_DISK_IMAGE_SNAPSHOT:-true}
```
mattcary, line 32: Can we use multiple snapshot class files with the test flag? If so, we should just set the CI to run both the conventional & disk image snapshots.

luohao:
@mattcary  I updated the driver name by appending the snapshotclass name to it(when no snapshot class is specified, it appends `no-volumesnapshotclass` to it). 

The driver name(`csi-gcepd-sc-standard--pd-volumesnapshotclass`) looks ok to me(originally I thought the name might be too long).

Can you take a look, please? Thanks!

PR #935, opened 2022-03-22
Add support for extra-create-metadata when creating snapshots

### LINKED ISSUE #695: Implement extra-create-metadata in CreateSnapshot
`--extra-create-metadata` was introduced in external-snapshotter 4.0.0.

francoispqt:
**What type of PR is this?**
/kind feature

This PR adds metadata in the description field for Snapshots and Images created from the CreateSnapshot CSI method when the external-snapshotter is run with the `--extra-create-metadata` flag.

Fixes #695 

```release-note
Description field for Images and Snapshots will be filled with JSON encoded metadata containing the name and namespace of the VolumeSnapshot and the name of the VolumeSnapshot content.
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -324,7 +324,7 @@ func ValidateDiskParameters(disk *CloudDisk, params common.DiskParameters) error
 func (cloud *CloudProvider) InsertDisk(ctx context.Context, project string, volKey *meta.Key, params common.DiskParameters, capBytes int64, capacityRange *csi.CapacityRange, replicaZones []string, snapshotID string, volumeContentSourceVolumeID string, multiWriter bool) error {
 	klog.V(5).Infof("Inserting disk %v", volKey)
 
-	description, err := encodeDiskTags(params.Tags)
+	description, err := encodeTags(params.Tags)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


mattcary:
/ok-to-test

Thanks for working on this! Basically looks okay to me, I'll wait for it to run through tests.

Remember to squash your commits before checking in.

francoispqt:
Thanks for reviewing!

Sure I will squash my commit before merging.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -1088,17 +1113,17 @@ func removeCryptoKeyVersion(kmsKey string) string {
 	return kmsKey
 }
 
-// encodeDiskTags encodes requested volume tags into JSON string, as GCE does
-// not support tags on GCE PDs and we use Description field as fallback.
-func encodeDiskTags(tags map[string]string) (string, error) {
+// encodeTags encodes requested volume or snapshot tags into JSON string, as GCE does
```
mattcary, line 1116: This comment is a little out of date, as PD does support labels. To reduce confusion, could you reword this something like the following?

```
enocodeTags encodes requested volume or snapshot tags into a JSON string, suitable for putting into 
the PD Description field.
```

francoispqt: You're right. I have updated the comment. 

I think using labels would be more convenient than using the description field (for cost monitoring for example) but even though Snapshot and PD now support labels, the constraint on label values (max length 63 char) make it unsuitable.

mattcary: Thanks. Let's wait for the tests to pass, squash your commits, and I'll lgtm.

francoispqt: I have just squashed commits

mattcary:
/lgtm
/approve

Thanks!

PR #956, opened 2022-04-01
allow to specify wait time for attach disk operation

sagor999:
**What type of PR is this?**

/kind feature

Currently poll time for attach disk or global\regional\zonal operation is hard coded:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/9681ab1fa79bce8e7a33ad308e2ae96dd388a088/pkg/gce-cloud-provider/compute/gce-compute.go#L781

For example, for attach disk it is hard coded to wait at least 5 seconds before checking if disk was attached.

For our [project](https://gitpod.io) we need to be able to attach disk as fast as possible, and usually it is attached extremely fast (less than 1 second). So we need to be able to poll faster.

This PR allows to do that via providing parameters as command line arguments that allow you to change how often we poll for changes.

Default parameters remain the same (for example AttachDisk is still polled every 5 seconds with timeout of 2 minutes as before), so existing installations will not notice any change.
 

Fixes #

```release-note
Allow to specify how frequently to poll for AttachDisk operation status, or any other global\regional\zonal operation.
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -33,14 +33,24 @@ import (
 )
 
 var (
-	cloudConfigFilePath  = flag.String("cloud-config", "", "Path to GCE cloud provider config")
-	endpoint             = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
-	runControllerService = flag.Bool("run-controller-service", true, "If set to false then the CSI driver does not activate its controller service (default: true)")
-	runNodeService       = flag.Bool("run-node-service", true, "If set to false then the CSI driver does not activate its node service (default: true)")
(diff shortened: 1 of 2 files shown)


rzetelskik:
> usually it is attached extremely fast (less than 1 second)

@sagor999 I hope you don't mind me digging this out, but were you actually able to achieve a subsecond atachment? Experimentally I wasn't able to go below roughly a 3 second threshold, even outside of GKE, i.e. with just a VM, PD and a CLI attach-disk command.

rzetelskik:
> I don't think you're going to achieve reliable sub-second attach time, at least according to our fleet numbers. P50 is still seconds.

@mattcary thanks for the input. Are these metrics you're referring to available for the broader public anywhere?

PR #960, opened 2022-04-06
Simplify the node backoff logic

saikat-royc:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR aims to simply the node backoff logic by removing the worker based backoff queue, to a simple map which tracks the active backoff condition per node. Removing the worker, removes dual sources of controller publish/unpublish calls (one from side car, one from the queue worker), and potential out of order execution of attach/detach disks operations.

Fixes #

Tested manually by running multiple cycles of attach/detach operations of 100 disks for a single pod.

```release-note
Simplify the node backoff logic
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -258,8 +258,17 @@ func IsGCENotFoundError(err error) bool {
 	return IsGCEError(err, "notFound")
 }
 
-// IsInvalidError returns true if the error is a googleapi.Error with
+// IsGCEInvalidError returns true if the error is a googleapi.Error with
 // invalid reason
 func IsGCEInvalidError(err error) bool {
 	return IsGCEError(err, "invalid")
+}
+
+// IsTooManyRequestError returns true if the error is a googleapi.Error with
+// resource exhausted error code.
+func IsTooManyRequestError(err error) bool {
+	if apierr, ok := err.(*googleapi.Error); ok && apierr.Code == http.StatusTooManyRequests {
+		return true
+	}
+	return false
 }
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -30,12 +30,16 @@ import (
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/util/workqueue"
+	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
+)
+
+const (
+	nodeBackoffInitialDuration = 200 * time.Millisecond
+	nodeBackoffMaxDuration     = 5 * time.Minute
 )
 
 type GCEControllerServer struct {
@@ -50,17 +54,9 @@ type GCEControllerServer struct {
 	// Aborted error
 	volumeLocks *common.VolumeLocks
 
-	// queue is a rate limited work queue for Controller Publish/Unpublish
-	// Volume calls
-	queue workqueue.RateLimitingInterface
-
-	// publishErrorsSeenOnNode is a list of nodes with attach/detach
-	// operation failures so those nodes shall be rate limited for all
-	// the attach/detach operations until there is an attach / detach
-	// operation succeeds
-	publishErrorsSeenOnNode map[string]bool
-
-	opsManager *OpsManager
+	// nodeBackoff keeps track of any active backoff condition on a given node, and the time when retry of controller publish/unpublish is permissible.
+	nodeBackoff *flowcontrol.Backoff
+	opsManager  *OpsManager
 }
 
 type workItem struct {
@@ -337,46 +333,6 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 // Run starts the GCEControllerServer.
 func (gceCS *GCEControllerServer) Run() {
 	go gceCS.opsManager.HydrateOpsCache()
-	go wait.Until(gceCS.worker, 1*time.Second, wait.NeverStop)
-}
-
-func (gceCS *GCEControllerServer) worker() {
-	// Runs until workqueue is shut down
-	for gceCS.processNextWorkItem() {
-	}
-}
-
-func (gceCS *GCEControllerServer) processNextWorkItem() bool {
-	item, quit := gceCS.queue.Get()
-	if quit {
-		return false
-	}
-	defer gceCS.queue.Done(item)
-
-	workItem, ok := item.(*workItem)
-	if !ok {
-		gceCS.queue.AddRateLimited(item)
-		return true
-	}
-
-	if workItem.publishReq != nil {
-		_, err := gceCS.executeControllerPublishVolume(workItem.ctx, workItem.publishReq)
-
-		if err != nil {
-			klog.Errorf("ControllerPublishVolume failed with error: %v", err)
-		}
-	}
-
-	if workItem.unpublishReq != nil {
-		_, err := gceCS.executeControllerUnpublishVolume(workItem.ctx, workItem.unpublishReq)
-
-		if err != nil {
-			klog.Errorf("ControllerUnpublishVolume failed with error: %v", err)
-		}
-	}
-
-	gceCS.queue.Forget(item)
-	return true
 }
 
 func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
@@ -384,28 +340,27 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 		return nil, status.Errorf(codes.Aborted, "Cache not ready")
 	}
 
-	// Only valid requests will be queued
+	// Only valid requests will be accepted
 	_, _, err := gceCS.validateControllerPublishVolumeRequest(ctx, req)
-
 	if err != nil {
 		return nil, err
 	}
 
-	// If the node is not marked, proceed the request
-	if _, found := gceCS.publishErrorsSeenOnNode[req.NodeId]; !found {
-		return gceCS.executeControllerPublishVolume(ctx, req)
+	if gceCS.nodeBackoff.IsInBackOffSinceUpdate(req.NodeId, gceCS.nodeBackoff.Clock.Now()) {
+		return nil, status.Errorf(codes.Unavailable, "ControllerPublish not permitted on node %q due to backoff", req.NodeId)
 	}
 
-	// Node is marked so queue up the request. Note the original gRPC context may get canceled,
-	// so a new one is created here.
-	//
-	// Note that the original context probably has a timeout (see csiAttach in external-attacher),
-	// which is ignored.
-	gceCS.queue.AddRateLimited(&workItem{
-		ctx:        context.Background(),
-		publishReq: req,
-	})
-	return nil, status.Error(codes.Unavailable, "Request queued due to error condition on node")
+	resp, err := gceCS.executeControllerPublishVolume(ctx, req)
+	backoff := isResourceExhaustedError(err)
+	if backoff && !gceCS.nodeBackoff.IsInBackOffSinceUpdate(req.NodeId, gceCS.nodeBackoff.Clock.Now()) {
+		klog.V(5).Infof("For node %s adding backoff due to error for volume %s", req.NodeId, req.VolumeId)
+		gceCS.nodeBackoff.Next(req.NodeId, gceCS.nodeBackoff.Clock.Now())
+	} else if err == nil {
+		klog.V(5).Infof("For node %s clear backoff due to successful publish of volume %v", req.NodeId, req.VolumeId)
+		gceCS.nodeBackoff.Reset(req.NodeId)
+	}
+
+	return resp, err
 }
 
 func (gceCS *GCEControllerServer) validateControllerPublishVolumeRequest(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (string, *meta.Key, error) {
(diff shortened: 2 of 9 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -20,18 +20,18 @@ import (
 	"math/rand"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	b "github.com/cenkalti/backoff/v4"
```
mattcary, line 27: Can't you use Backoff from k8s apimachinery/.../wait? The ExponentialBackoff bit won't work as it's based on timers, but Backoff.Step() seems about the same as cenkalti.NextBackoff that you use below.

Note this will save us doing an open source review before deploying in GKE.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1353,3 +1304,64 @@ func pickRandAndConsecutive(slice []string, n int) ([]string, error) {
...
+func (n *NodeBackoffInfo) DeleteNodeBackoffEntry(nodeId string) {
+	n.Lock()
+	defer n.Unlock()
+	delete(n.backoff, nodeId)
+}
+
+func (n *NodeBackoffInfo) AddOrUpdateNodeBackoffEntry(nodeId string) {
+	n.Lock()
+	defer n.Unlock()
+	boffInfo, ok := n.backoff[nodeId]
```
mattcary, line 1328: nit: bkoffInfo reads a little nicer than boffInfo.

pkg/gce-pd-csi-driver/controller.go
saikat-royc: Ack

saikat-royc: using the client-go flowcontrol lib now

pkg/gce-pd-csi-driver/controller.go
saikat-royc: stale comment now

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -30,14 +30,18 @@ import (
...
-	"k8s.io/client-go/util/workqueue"
+	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
 )
 
+const (
+	nodeBackoffInitialDuration = 200 * time.Millisecond
```
pwschuurman, line 41: How were these numbers selected? We used to wait up to 16 minutes per item: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/11e2389ad189eed794c16239d68c97c64c4d0d70/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L41

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -436,11 +388,12 @@ func (gceCS *GCEControllerServer) validateControllerPublishVolumeRequest(ctx con
 	return project, volKey, nil
 }
 
-func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
+// executeControllerPublishVolume returns the controllerPublishVolumeResponse object. If there is an error, the function provides additional boolean hint on whether to set a backoff condition on the node.
+func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, bool, error) {
```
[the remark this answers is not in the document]
pwschuurman: Another thought: If the goal of the exponential backoff is to prevent/mitigate quota exhaustion, can we encode quota exhaustion errors from the GCE PD API and only backoff on these errors?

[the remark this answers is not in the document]
saikat-royc: Thats a good point. I think we ideally want to backoff only on resource quota exceeded errors i.e http 429 errors.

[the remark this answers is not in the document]
saikat-royc: also about the RepairUnderspecifiedVolumeKey I think I missed it. its a bug on my end

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -645,12 +595,17 @@ func (gceCS *GCEControllerServer) executeControllerUnpublishVolume(ctx context.C
...
+		backoff := false
+		// Errors during API calls are reported as Internal
+		if isInternalError(err) {
+			backoff = true
+		}
+		return nil, backoff, err
 	}
 
-	delete(gceCS.publishErrorsSeenOnNode, nodeID)
+	// delete(gceCS.publishErrorsSeenOnNode, nodeID)
```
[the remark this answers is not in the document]
saikat-royc: done

pkg/gce-pd-csi-driver/gce-pd-driver.go
```diff
@@ -151,13 +151,12 @@ func NewNodeServer(gceDriver *GCEDriver, mounter *mount.SafeFormatAndMount, devi
...
-		seen:                    map[string]int{},
-		volumeLocks:             common.NewVolumeLocks(),
-		queue:                   workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "controllerserver"),
-		publishErrorsSeenOnNode: map[string]bool{},
-		opsManager:              NewOpsManager(cloudProvider),
+		Driver:        gceDriver,
+		CloudProvider: cloudProvider,
+		seen:          map[string]int{},
+		volumeLocks:   common.NewVolumeLocks(),
+		nodeBackoff:   flowcontrol.NewBackOff(nodeBackoffInitialDuration, nodeBackoffMaxDuration),
```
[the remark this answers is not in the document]
saikat-royc: done

pkg/gce-pd-csi-driver/controller.go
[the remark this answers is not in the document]
saikat-royc: done

pkg/gce-pd-csi-driver/controller.go
saikat-royc: I have added some scenarios in the [doc](https://docs.google.com/document/d/138UQqyIbuWQ3xc_3fCXaI7MzKX_veNJe1bdWZ1ZZIcg/edit?resourcekey=0-f3uGOrMCZcmsd-v6meapFQ). 

PS. I have changed the backoff implementation slightly such that if the failure happens within the backoff condition true, then no new backoff delay added. The previous implemetation was adding a delay on every failed attempt.

saikat-royc: FYI @pwschuurman

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -586,56 +548,62 @@ func (gceCS *GCEControllerServer) validateControllerUnpublishVolumeRequest(ctx c
...
 	volumeID := req.GetVolumeId()
 	nodeID := req.GetNodeId()
 	project, volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, project, volKey)
 	if err != nil {
 		if gce.IsGCENotFoundError(err) {
-			return nil, status.Errorf(codes.NotFound, "ControllerUnpublishVolume could not find volume with ID %v: %v", volumeID, err)
+			return nil, false, status.Errorf(codes.NotFound, "ControllerUnpublishVolume could not find volume with ID %v: %v", volumeID, err)
+		}
+		if gce.IsTooManyRequestError(err) {
+			return nil, true, status.Errorf(codes.Internal, "ControllerUnpublishVolume error repairing underspecified volume key: %v", err)
```
pwschuurman, line 566: Do you want to return codes.ResourceExhausted here?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -586,56 +548,62 @@ func (gceCS *GCEControllerServer) validateControllerUnpublishVolumeRequest(ctx c
...
 	if err != nil {
 		if gce.IsGCENotFoundError(err) {
 			// Node not existing on GCE means that disk has been detached
 			klog.Warningf("Treating volume %v as unpublished because node %v could not be found", volKey.String(), instanceName)
-			return &csi.ControllerUnpublishVolumeResponse{}, nil
+			return &csi.ControllerUnpublishVolumeResponse{}, false, nil
 		}
-		return nil, status.Error(codes.Internal, fmt.Sprintf("error getting instance: %v", err))
+		if gce.IsTooManyRequestError(err) {
+			return nil, true, status.Errorf(codes.Internal, "ControllerUnpublishVolume error repairing underspecified volume key: %v", err)
```
pwschuurman, line 591: codes.ResourceExhausted

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -337,75 +333,33 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
...
-
 	if err != nil {
 		return nil, err
 	}
 
-	// If the node is not marked, proceed the request
-	if _, found := gceCS.publishErrorsSeenOnNode[req.NodeId]; !found {
-		return gceCS.executeControllerPublishVolume(ctx, req)
+	if gceCS.nodeBackoff.IsInBackOffSinceUpdate(req.NodeId, gceCS.nodeBackoff.Clock.Now()) {
+		return nil, status.Errorf(codes.Unavailable, "ControllerPublish not permitted on node %q due to backoff", req.NodeId)
```
pwschuurman, line 350: Worth it to log when the backoff expires here (and maybe current time if we don't have that in the log already)? That might help with debugging to see when requests were expected to become unavailable

saikat-royc: Unfortunately the flowcontrol implementation does not expose the last update time. Also the Get() which gives the backoff duration is a protected access i.e have to acquire lock again to read the entry. The best info we have is the logging while calling Next(). Thoughts? any suggestion

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -436,11 +390,11 @@ func (gceCS *GCEControllerServer) validateControllerPublishVolumeRequest(ctx con
 	return project, volKey, nil
 }
 
-func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
+// executeControllerPublishVolume returns the controllerPublishVolumeResponse object. If there is an error, the function provides additional boolean hint on whether to set a backoff condition on the node.
+func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, bool, error) {
```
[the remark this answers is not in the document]
saikat-royc: removed the bool for backoff.

The error should now be logged in the cloud provider core api calls. also any error would also show up in attacher sidecar logs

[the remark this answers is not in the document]
saikat-royc: In general I agree we should revisit the logging to fix things in uniform way across the code

pkg/gce-pd-csi-driver/controller.go
saikat-royc: fixed it

pkg/gce-pd-csi-driver/controller.go
saikat-royc: fixed

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -263,3 +263,12 @@ func IsGCENotFoundError(err error) bool {
 func IsGCEInvalidError(err error) bool {
 	return IsGCEError(err, "invalid")
 }
+
+// IsGCENotFoundError returns true if the error is a googleapi.Error with
```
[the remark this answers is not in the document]
saikat-royc: fixed

pkg/gce-pd-csi-driver/controller.go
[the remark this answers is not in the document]
saikat-royc: We can take this up in a separate PR

pkg/gce-pd-csi-driver/controller.go
saikat-royc: As discussed keeping the logs as the source of truth

pwschuurman:
LGTM

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -523,20 +486,20 @@ func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Con
...
+		}
 		return nil, err
 	}
 
 	err = gceCS.CloudProvider.WaitForAttach(ctx, project, volKey, instanceZone, instanceName)
 	if err != nil {
-		// Mark the node and rate limit all the following attach/detach
-		// operations for this node
-		gceCS.publishErrorsSeenOnNode[nodeID] = true
+		if gce.IsTooManyRequestError(err) {
```
mattcary, line 497: I think we want to back off on any error. With this logic, we'll let the attacher retry until api quota gets exhausted. I think we would rather not even reach that point.

I don't think there's a case where immediate retries are useful?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -546,25 +509,25 @@ func (gceCS *GCEControllerServer) ControllerUnpublishVolume(ctx context.Context,
...
-
-	// Node is marked so queue up the request
-	gceCS.queue.AddRateLimited(&workItem{
-		ctx:          ctx,
-		unpublishReq: req,
-	})
-
-	return &csi.ControllerUnpublishVolumeResponse{}, nil
+	resp, err := gceCS.executeControllerUnpublishVolume(ctx, req)
+	backoff := isResourceExhaustedError(err)
```
amacaskill, line 522: I agree. I think we should backoff for all errors from  executeControllerUnpublishVolume, but we have also seen a  customer case where executeControllerUnpublishVolume returns true: `ControllerUnpublishVolume succeeded for disk. Already not attached.` They still ended up with the same issue where the finalizer external-attacher/pd-csi-storage-gke-io was still set on the VA. I think the CSI attacher was the one that kept retrying because the csi-attacher logs repeated
`VA csi-xx for volume projects/xx/zones/us-east4-a/disks/pvc-xx has attached status true but actual state false. Adding back to VA queue for forced reprocessing`
Which made the kubecontroller manager keep trying to force detach but failing.

amacaskill: To elaborate,  in some cases, 
instance, err := gceCS.CloudProvider.GetInstanceOrError(ctx, instanceZone, instanceName)
will still return the instance , but the device will not show up under devices, so then controllerUnpublishVolume returns true, but then we never called detachDisk (which I am assuming is what makes VA attached status to be false).  So then the csi-attacher keeps queuing the VA. `VA csi-xx has attached status true but actual state false. Adding back to VA queue for forced reprocessing`

PR #963, opened 2022-04-19
Add functionality to make architecture specific builds

### LINKED ISSUE #944: Add E2E test flags to specify machine type, zone, OS image, architecture
Currently the E2E tests spin up two GCE VM instances by default. These install:
 * Test Driver
 * PD CSI Driver that accepts CSI calls

In order to test the PD CSI driver on different machine types that may have different configuration, or operating systems that have different dependencies, we should have flags that control:
 * OS Image: Currently this is hard coded
 * Machine Type: Current default is n1-standard-1
 * Architecture: Dependent on machine-type, built binaries need to match the architecture platform that the driver is running on
 * Zone: If machine types are available in specific zones, this may need to be specified

These flags can likely live in the remote package: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/remote

In addition to these flags, the built architecture of the Test Driver and the PD CSI Driver should match. These are built for the e2e test here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/remote/archiver.go#L58

DW2022511:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #944 

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,10 @@ gce-pd-driver: require-GCE_PD_CSI_STAGING_VERSION
 	go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
 
 gce-pd-driver-windows: require-GCE_PD_CSI_STAGING_VERSION
+ifeq ($(ARCH), amd64)
 	mkdir -p bin
 	GOOS=windows go build -mod=vendor -ldflags -X=main.version=$(STAGINGVERSION) -o bin/${DRIVERWINDOWSBINARY} ./cmd/gce-pd-csi-driver/
+endif
 
 build-container: require-GCE_PD_CSI_STAGING_IMAGE require-GCE_PD_CSI_STAGING_VERSION init-buildx
 	$(DOCKER) buildx build --platform=linux --progress=plain \
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,9 @@ gce-pd-driver: require-GCE_PD_CSI_STAGING_VERSION
 gce-pd-driver-windows: require-GCE_PD_CSI_STAGING_VERSION
 ifeq ($(ARCH), amd64)
 	mkdir -p bin
-	GOOS=windows go build -mod=vendor -ldflags -X=main.version=$(STAGINGVERSION) -o bin/${DRIVERWINDOWSBINARY} ./cmd/gce-pd-csi-driver/
+	GOOS=windows GOARCH=$(ARCH) go build -mod=vendor -ldflags -X=main.version=$(STAGINGVERSION) -o bin/${DRIVERWINDOWSBINARY} ./cmd/gce-pd-csi-driver/
+else
+  echo "Warning: gcp-pd-driver-windows only supports ARCH=amd64."
 endif
 
 build-container: require-GCE_PD_CSI_STAGING_IMAGE require-GCE_PD_CSI_STAGING_VERSION init-buildx
--- a/Makefile
+++ b/Makefile
@@ -41,11 +41,11 @@ gce-pd-driver: require-GCE_PD_CSI_STAGING_VERSION
 	go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
 
 gce-pd-driver-windows: require-GCE_PD_CSI_STAGING_VERSION
-ifeq ($(ARCH), amd64)
+ifeq (GOARCH, amd64)
 	mkdir -p bin
-	GOOS=windows GOARCH=$(ARCH) go build -mod=vendor -ldflags -X=main.version=$(STAGINGVERSION) -o bin/${DRIVERWINDOWSBINARY} ./cmd/gce-pd-csi-driver/
+	GOOS=windows go build -mod=vendor -ldflags -X=main.version=$(STAGINGVERSION) -o bin/${DRIVERWINDOWSBINARY} ./cmd/gce-pd-csi-driver/
 else
-  echo "Warning: gcp-pd-driver-windows only supports ARCH=amd64."
+  $(warning gcp-pd-driver-windows only supports amd64.)
 endif
 
 build-container: require-GCE_PD_CSI_STAGING_IMAGE require-GCE_PD_CSI_STAGING_VERSION init-buildx
--- a/test/e2e/tests/setup_e2e_test.go
+++ b/test/e2e/tests/setup_e2e_test.go
@@ -35,6 +35,8 @@ import (
 
 var (
 	project         = flag.String("project", "", "Project to run tests in")
+	architecture    = flag.String("arch", "amd64", "Architecture pd csi driver build on")
+	vmName          = flag.String("vm-name", "", "Name of the VM to run tests on")
 	serviceAccount  = flag.String("service-account", "", "Service account to bring up instance with")
 	zones           = flag.String("zones", "us-central1-c,us-central1-b", "Zones to run tests in. If there are multiple zones, separate each by comma")
 	machineType     = flag.String("machine-type", "n1-standard-1", "Type of machine to provision instance on")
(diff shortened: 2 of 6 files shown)


DW2022511:
@pwschuurman

test/remote/archiver.go
```diff
@@ -53,8 +53,9 @@ func CreateDriverArchive(archiveName, pkgPath, binPath string) (string, error) {
 	return filepath.Join(dir, archiveName), nil
 }
 
-func setupBinaries(tarDir, pkgPath, binPath string) error {
+func setupBinaries(architecture, tarDir, pkgPath, binPath string) error {
 	klog.V(4).Infof("Making binaries and copying to temp dir...")
+	os.Setenv("ARCH", architecture)
 	out, err := exec.Command("make", "-C", pkgPath).CombinedOutput()
```
pwschuurman, line 59: I think you can also just call "make -C pkgPath ARCH=<architecture>", rather than setting an environment variable?

Makefile
```diff
@@ -35,14 +35,18 @@ ifdef GCE_PD_CSI_DEBUG
...
+ARCH ?= amd64
+
 all: gce-pd-driver gce-pd-driver-windows
 gce-pd-driver: require-GCE_PD_CSI_STAGING_VERSION
 	mkdir -p bin
-	go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
+	GOARCH=$(ARCH) go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
 
 gce-pd-driver-windows: require-GCE_PD_CSI_STAGING_VERSION
+ifeq ($(ARCH), amd64)
```
pwschuurman, line 46: Can you add an else block that emits a warning?

```
ifeq ($(ARCH), amd64)
  ...
else
	echo "Warning: gcp-pd-driver-windows only supports ARCH=amd64."
endif
```
And might as well set GOARCH for windows too. Right now we're just building for whatever the default host architecture is.

test/remote/archiver.go
DW2022511: done

Makefile
DW2022511: done

test/e2e/tests/setup_e2e_test.go
```diff
@@ -34,6 +34,8 @@ import (
 
 var (
 	project         = flag.String("project", "", "Project to run tests in")
+	architecture    = flag.String("arch", "amd64", "Architecture pd csi driver build on")
+	vmName          = flag.String("vm-name", "", "Name of the VM to run tests on")
```
DW2022511, line 38: @pwschuurman I run the t2a test, and we need to make vmName as flag as well.

Makefile
```diff
@@ -35,14 +35,20 @@ ifdef GCE_PD_CSI_DEBUG
 	GCFLAGS="all=-N -l"
 endif
 
+ARCH ?= amd64
+
 all: gce-pd-driver gce-pd-driver-windows
 gce-pd-driver: require-GCE_PD_CSI_STAGING_VERSION
 	mkdir -p bin
-	go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
+	GOARCH=$(ARCH) go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
```
leiyiz, line 43: I suspect this might break the multi-arch build process because the `GOARCH` is set in the `Dockerfile`. Have you tried and make sure that `build-and-push-container-linux-arm64` still works as intended?

DW2022511: It works. I think we pass the info through this arg --platform=linux/arm64

leiyiz: So if you look into `Dockerfile` you can see eventually it calls this with GOARCH set https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/Dockerfile#L24

pwschuurman: Good call @leiyiz. @DW2022511, can we just set GOARCH= variable when we invoke the make command, similar to what is done in the Dockerfile?

DW2022511: set GOARCH= when invoke the make command works.

test/remote/instance.go
```diff
@@ -121,31 +123,47 @@ func (i *InstanceInfo) CreateOrGetInstance(imageURL, serviceAccount string) erro
...
-	if _, err := i.computeService.Instances.Get(i.project, i.zone, inst.Name).Do(); err != nil {
-		op, err := i.computeService.Instances.Insert(i.project, i.zone, inst).Do()
-		klog.V(4).Infof("Inserted instance %v in project: %v, zone: %v", inst.Name, i.project, i.zone)
-		if err != nil {
-			ret := fmt.Sprintf("could not create instance %s: API error: %v", i.name, err)
-			if op != nil {
-				ret = fmt.Sprintf("%s. op error: %v", ret, op.Error)
+	curInst, _ := i.computeService.Instances.Get(i.project, i.zone, newInst.Name).Do()
+	if curInst != nil {
+		if !strings.Contains(curInst.MachineType, newInst.MachineType) {
```
pwschuurman, line 139: Can you just compare these as equal?

test/remote/instance.go
```diff
@@ -121,31 +123,47 @@ func (i *InstanceInfo) CreateOrGetInstance(imageURL, serviceAccount string) erro
...
+				return true, nil
+			})
+			if err != nil {
+				return err
 			}
-			return errors.New(ret)
-		} else if op.Error != nil {
-			return fmt.Errorf("could not create instance %s: %+v", i.name, op.Error)
+
+			if err := insertInstance(i, newInst); err != nil {
```
pwschuurman, line 157: Can you refactor this so your function only calls insertInsert 1x instead of 2x? This if statement is 3 levels deep at one point and I think the logic can be simplified to escape out of some error conditions prior to returning success. IMO the main flow of a function should be the happy-path, with if statements reserved for exceptional scenarios and error handling. I think the general flow can be:
 * Check if instance exists. If instance exists and not matching machine-type: Delete and verify instance does not exist.
 * If instance not exists: Create

test/remote/instance.go
DW2022511: The machine type string get from service has a prefix [https://www.googleapis.com/compute/v1/projects/${PROJECT}](https://www.googleapis.com/compute/v1/projects/$%7BPROJECT%7D). If we want to use equal I need to parse the string. I feel parse prefix is similar to using Contains?

pwschuurman: Ok, let's keep as is then.

test/remote/instance.go
DW2022511: Make sense. I remove the insertInstance method and add the original code block back. Check the machine-type before insertion.

test/remote/instance.go
```diff
@@ -121,20 +123,43 @@ func (i *InstanceInfo) CreateOrGetInstance(imageURL, serviceAccount string) erro
...
+				}
+				return true, nil
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	if _, err := i.computeService.Instances.Get(i.project, i.zone, newInst.Name).Do(); err != nil {
```
pwschuurman, line 160: Can you just check of curInst (or instance) is nil here? I don't think you need an extra Get call. This will also make your API response handling more consistent (right now you're using the first and second parameter in the return call to indicate the same semantics)

pwschuurman (approved):
LGTM

test/remote/instance.go
DW2022511: done.

DW2022511:
/assign @saikat-royc
/assign @mattcary 
This PR should be ready to merge can you take a look?

Makefile
```diff
@@ -41,8 +41,12 @@ gce-pd-driver: require-GCE_PD_CSI_STAGING_VERSION
 	go build -mod=vendor -gcflags=$(GCFLAGS) -ldflags "-X main.version=$(STAGINGVERSION)" -o bin/${DRIVERBINARY} ./cmd/gce-pd-csi-driver/
 
 gce-pd-driver-windows: require-GCE_PD_CSI_STAGING_VERSION
+ifeq (GOARCH, amd64)
```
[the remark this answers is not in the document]
mauriciopoppe: I'm attempting to fix this in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/973

PR #966, opened 2022-04-27
Run snapshot class tests separately from other tests

mattcary:
Running all snapshot classes with all storage class is causing timeouts. Running the snapshot tests with just one storageclass does not substantially reduce the test coverage.

/kind failing-test

```release-note
None
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -93,6 +93,7 @@ type testParameters struct {
 	k8sSourceDir         string
 	testFocus            string
 	testSkip             string
+	storageClassFile     string
 	snapshotClassFile    string
 	cloudProviderArgs    []string
 	deploymentStrategy   string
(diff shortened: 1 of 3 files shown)


mattcary:
/hold

I'll want to examine the CI test run to make sure it's working as expected.

mattcary:
Still timing out, will have to see what's getting run

mattcary:
I don't understand ginkgo focus very well I guess; I think we have to basically ignore the passed in focus if there is a snapshot class given and use the volume snapshot feature. I'll have to double-check that's what we want here.

mattcary:
/hold cancel

finally it looks right! @leiyiz can you take a look & lgtm?

test/k8s-integration/main.go
```diff
@@ -668,7 +675,20 @@ func runTestsWithConfig(testParams *testParameters, testConfigArg, reportPrefix
 			kubetestDumpDir = artifactsDir
 		}
 	}
-	ginkgoArgs := fmt.Sprintf("--ginkgo.focus=%s --ginkgo.skip=%s", testParams.testFocus, testParams.testSkip)
+
+	focus := testParams.testFocus
+	skip := testParams.testSkip
+	if testParams.snapshotClassFile == "" && !strings.Contains(skip, "VolumeSnapshotDataSource") {
```
leiyiz, line 681: I think if the `snapshotClassFile` is not set, the k/k integration framework would not run snapshot tests? is that not the case

PR #975, opened 2022-05-06
Remove livenessprobe from node-driver-registrar

mauriciopoppe:
**What type of PR is this?**
>

/kind cleanup

The root cause for the kubelet registration issue was found in https://github.com/kubernetes/kubernetes/issues/104584, with that there's no need for this workaround.

```release-note
NONE
```


mauriciopoppe:
@mattcary could you please approve this PR? I saw all tests passed :)

PR #976, opened 2022-05-06
Update arm64 build to use distroless debian image

pwschuurman:
**What type of PR is this?**
/kind feature

The amd64 Dockerfile was updated in #870 to use the distroless debian image. This updates the arm64 build to use the same Dockerfile, and updates the shared library dependency commands.

```release-note
NONE
```


leiyiz (commented):
/lgtm
/approve

PR #985, opened 2022-05-12
Add Makefile target that validates the Dockerfile images contain the correct shared library dependencies

pwschuurman:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**: This uses ldd on the target architecture (amd64/arm64) to validate if all direct and indirect dependencies of a program. The underlying logic leverages ldd to find dependencies that are "not found". This is helpful to catch in presubmit issues where all the runtime libs were not included. See #974 as an example of this.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,22 @@ build-and-push-multi-arch-debug: build-and-push-container-linux-debug build-and-
 
 push-container: build-container
 
+# Used by hack/verify-docker-deps.sh, not used for building artifacts
+validate-container-linux-amd64: init-buildx
+	$(DOCKER) buildx build --platform=linux/amd64 \
+		-t validation_linux_amd64 \
+		--target validation-image \
+		--build-arg BUILDPLATFORM=linux \
+		--build-arg STAGINGVERSION=$(STAGINGVERSION) .
+
+# Used by hack/verify-docker-deps.sh, not used for building artifacts
+validate-container-linux-arm64: init-buildx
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -79,6 +79,20 @@ build-and-push-multi-arch-debug: build-and-push-container-linux-debug build-and-
 
 push-container: build-container
 
+validate-container-linux-amd64: init-buildx
```
mattcary, line 82: Add a comment how these build rules should be used (eg, the validation is done in build-and-push, so these are useful when you want to validate but don't want to figure out / overwrite a $(stagingimage):$(stagingversion)?)

hack/verify-docker-deps.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
mattcary, line 1: Add this to verify-all?

Makefile
mattcary: Or actually I guess they're used by the verify script.

pwschuurman: Added a comment about how this is used.

hack/verify-docker-deps.sh
pwschuurman: Added

pwschuurman: Will add in a later CL. We need https://github.com/kubernetes/test-infra/pull/26312 to go through in order for Docker commands to be supported in verify-all.sh

Issue #982, opened 2022-05-11
Windows CI tests are failing with `invalid character ',' looking for beginning of object key string `


mauriciopoppe:
After a few green runs in CI for Windows I saw that the CI jobs failed with the error `invalid character ',' looking for beginning of object key string`, I checked recent merged PRs and none of them seem to be modifying the Makefile paths related with Windows (the closest PR was https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/976).

mauriciopoppe:
After #986 was merged I saw that the post submit job created the image gcr.io/k8s-staging-cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:master with these contents for the manifest:

```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 957,
         "digest": "sha256:1805f8eb3bcf7b004ec6fa5bb0230cdd0d16e5ef7e121653f9f902d1f7e4279b",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.19042.1706"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 7009,
         "digest": "sha256:51f90c6cc412ac3ad479b74944aae4d57784bbf53d9f59c67d7d4a36b8105fa2",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 7009,
         "digest": "sha256:5d473a7af3d98c32823eafbf8c6efa7d1b51df70c889d6a8ddcb5ca227f4abb4",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 956,
         "digest": "sha256:7451eb02135eefc8abe6e8ac3810840cb8808c88a7207bf4edcdcabd9e508daa",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.17763.2928"
         }
      }
   ]
}
```

mattcary:
Awesome, thanks for tracking this.

PR #986, opened 2022-05-12
Remove workaround to add os.version to the manifest for Windows images

mauriciopoppe:
**What type of PR is this?**
>

/kind bug

It looks like a recent version of docker in CI is adding `os.version` so the workaround to add it on our own is no longer needed, this was causing a double `os.version` to be added to the manifest as seen here:

```
I0512 18:36:46.423] {"Ref":"gcr.io/gce-cvm-upg-lat-1-3-ctl-skew/gcp-persistent-disk-csi-driver:d5486a25-280b-4d7e-a899-98a3f37b0818_ltsc2019","Descriptor":{"mediaType":"application/vnd.docker.distribution.manifest.v2+json","digest":"sha256:cb0b111aba51dfdfcf8fcdec778c20c3bf153c9180bcbaf597c7a5f58c886794","size":956,"platform":{"architecture":"amd64","os":"windows","os.version":"10.0.17763.2928",,"os.version":"10.0.17763.2928"}}, ...}
```

I'm commenting the part that performs the text replacement, if this works then eventually this file and its usages should be removed.

Ref #982 

```release-note
NONE
```

/assign @saikat-royc

### DIFF
--- a/manifest_osversion.sh
+++ b/manifest_osversion.sh
@@ -32,8 +32,8 @@ for ((i=0;i<${#imagetags[@]};++i)); do
   image_folder=$(echo "${IMAGETAG}" | sed "s|/|_|g" | sed "s/:/-/")
   echo ${manifest_folder}
 
-  sed -i -r "s/(\"os\"\:\"windows\")/\0,\"os.version\":$full_version/" \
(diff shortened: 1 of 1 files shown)


mauriciopoppe:
@mattcary could you please approve this PR too?

PR #989, opened 2022-05-18
Update deployment for 1.24 and 1.25

mattcary:
/kind cleanup

Update overlay to work for 1.25 master, and make 1.24 overlay.

As part of this change, the PSP is removed from the master deployment, and for the 1.21-1.23 clusters added via customize. The noauth, debug and dev overlays do not have the PSP deployment as they are supposed to be tracking master.

Tested by doing deploy-driver with stable-1-24 on a  e2e-up from a 1.24 head, and stable-master from k/k head (two weeks after 1.24 cut, so basically 1.25). (tests for the other versions are in-progress).

```release-note
Remove PodSecurityPolicy from deployment for 1.25+ clusters.
```

/assign @saikat-royc


deploy/kubernetes/overlays/stable-1-24/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Kustomization
+namespace:
+  gce-pd-csi-driver
+resources:
+- ../../base
+- psp_controller.yaml
+- psp_linux.yaml
+- psp_windows.yaml
+transformers:
+- ../../images/stable-master
```
saikat-royc, line 11: ../../images/stable-1-24

deploy/kubernetes/images/stable-1-24/image.yaml
```diff
@@ -0,0 +1,56 @@
...
+apiVersion: builtin
+kind: ImageTagTransformer
+metadata:
+  name: imagetag-gcepd-driver
+imageTag:
+  name: gke.gcr.io/gcp-compute-persistent-disk-csi-driver
+  # Don't change stable image without changing pdImagePlaceholder in
+  # test/k8s-integration/main.go
+  newName: k8s.gcr.io/cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver
+  newTag: "v1.4.0"
```
saikat-royc, line 55: let's use the updated image?

mattcary: ah.. good point. But which version? Go back to 1.3.6 or up to 1.6.0 which still has the node death problem?

deploy/kubernetes/base/controller/kustomization.yaml
```diff
@@ -5,5 +5,4 @@ namespace:
 resources:
 - cluster_setup.yaml
 - controller.yaml
-- psp.yaml
```
saikat-royc, line 8: this change in base also means stable-1.23, stable-1.22, stable-1.21 wont have the psp. in this PR we are adding psp to the 1.24 overlay. should we consistent (i.e either remove psp from all 1.21 and beyond or keep psp in 1.21, 1.22, 1.23, 1.24 and removed from 1.25) ?

mattcary: Arg, you're absolutely right, good catch.

I think we should add it to the old ones so that stable-master is most simple.

deploy/kubernetes/images/stable-1-24/image.yaml
saikat-royc: Maybe we can keep it as is. We will soon cut a new release with fix (1.7)

mattcary:
/hold

Until I test this on 1.21 & 1.22 clusters. @saikat-royc if you want to remove the hold and lgtm so we can get the other fix out, please go ahead.

PR #1021, opened 2022-07-16
Remove passwd- file to make CIS benchmark happy

mattcary:
/kind bug

In order to pass the [CIS benchmark](https://www.cisecurity.org/cis-benchmarks/), the /etc/passwd- file can only be root-readable. Even better is to just remove it altogether, as the PR does.

```release-note
Pass [CIS benchmark](https://www.cisecurity.org/cis-benchmarks/) by cleaning up /etc/passwd- in the driver image.
```

/assign @amacaskill


mattcary:
I think it is, I'm trying to pin the tests for this branch to 1.21 to see if that allows us to test.

mattcary:
Sigh, it's failing now because hack/ginkgo-e2e.sh in release-1.21 has outdated flags. I'll have to force-push this and we'll run through the internal GKE tests.

PR #1024, opened 2022-07-19
Pin kubetest2 version

mattcary:
/kind failing-test

Pin kubetest2 version, as recent changes break the version of ginkgo we're using.

/assign @leiyiz

### DIFF
--- a/test/run-k8s-integration-ci.sh
+++ b/test/run-k8s-integration-ci.sh
@@ -54,10 +54,11 @@ export GCE_PD_VERBOSITY=9
 make -C "${PKGDIR}" test-k8s-integration
 
 if [ "$use_kubetest2" = true ]; then
-    go install sigs.k8s.io/kubetest2@latest;
-    go install sigs.k8s.io/kubetest2/kubetest2-gce@latest;
-    go install sigs.k8s.io/kubetest2/kubetest2-gke@latest;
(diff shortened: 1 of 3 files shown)


test/run-k8s-integration-ci.sh
```diff
@@ -54,10 +54,11 @@ export GCE_PD_VERBOSITY=9
 make -C "${PKGDIR}" test-k8s-integration
 
 if [ "$use_kubetest2" = true ]; then
-    go install sigs.k8s.io/kubetest2@latest;
-    go install sigs.k8s.io/kubetest2/kubetest2-gce@latest;
-    go install sigs.k8s.io/kubetest2/kubetest2-gke@latest;
-    go install sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo@latest;
+    kt2_version=0e09086b60c122e1084edd2368d3d27fe36f384f
```
leiyiz, line 57: consider having a comment about where/how to update this when needed

mattcary: I don't think we actually know what the process is to update this is unfortunately?

I think we should also consider that we're doing with release-tools/, as we pull that in but don't actually use it at all.

leiyiz: on filestore side we said something like "ask engprod oncall what version are they using internally if we need to update"

PR #1028, opened 2022-07-22
backoff per {node,disk} pair instead of just node}

### LINKED ISSUE #1029: Missing disks causes nodes to stick due to backoff
Around 07/13 issue started when a Pod attempted to use a non-existent volume we'll denote by X (this was a statically created PD, not dynamically provisioned from the driver). Presumably it existed at some point, because it was correctly attached to a node, and the PV, PVC and volume attachment (VA) still exist.

The VA csi-966b5b879f8fb9137a305680db1aff4e38231718d4dbc7883410d7a58530f161 was created pointing to that volume and a node Y.  This attachment failed due to a missing disk error and put a backoff condition on the node.

This pod was deleted, adding a delete timestamp to the VA object. This triggered detach workflow in csi-attacher. Detach keeps failing infinitely because of missing disk error (and keeping the node in a backoff condition).

At the same time another pod & PVC were scheduled to the node, but the repeated detach attempts happen too fast for the attach to ever succeed; instead it kept hitting the backoff.

We've decided on two actions. One, to be tracked with this issue, is to key the backoff by both disk and node. See the PR about to mention this issue.

The other, which will be tracked in a different issue, will be to assume a missing disk has been detached if it remains not found for a long enough time. Watch for the new issue.

mattcary:
I'll create an issue on this repo and update this message to point to it.

/kind bug

Fixes #1029

```release-note
```
/assign @saikat-royc

### DIFF
--- a/pkg/gce-pd-csi-driver/controller_test.go
+++ b/pkg/gce-pd-csi-driver/controller_test.go
@@ -2149,18 +2149,24 @@ type backoffTesterConfig struct {
 	mockMissingInstance bool // used by the backoff tester to mock a missing instance scenario
 }
 
+func newFakeCsiErrorBackoff(tc *clock.FakeClock) *csiErrorBackoff {
+	return &csiErrorBackoff{flowcontrol.NewFakeBackOff(errorBackoffInitialDuration, errorBackoffMaxDuration, tc)}
+}
+
 func TestControllerUnpublishBackoff(t *testing.T) {
-	backoffTesterForUnpublish(t, &backoffTesterConfig{
-		mockMissingInstance: true,
-	})
 	backoffTesterForUnpublish(t, &backoffTesterConfig{})
 }
 
+func TestControllerUnpublishBackoffMissingInstance(t *testing.T) {
+	backoffTesterForUnpublish(t, &backoffTesterConfig{
+		mockMissingInstance: true,
+	})
+}
+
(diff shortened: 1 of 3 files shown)


pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -2336,36 +2362,34 @@ func backoffTesterForPublish(t *testing.T, config *backoffTesterConfig) {
...
-	runPublishRequest := func(req *csi.ControllerPublishVolumeRequest) <-chan error {
+	runPublishRequest := func(req *csi.ControllerPublishVolumeRequest, reportError bool) error {
 		response := make(chan error)
 		go func() {
 			_, err := driver.cs.ControllerPublishVolume(context.Background(), req)
 			response <- err
 		}()
-		return response
+		go func() {
+			<-readyToExecute <- gcecloudprovider.Signal{ReportError: reportError}
```
saikat-royc, line 2372: the ordering here is a bit difficult to understand. Can we keep the multi-line read from channel in this go routine for readability?
```
	execute := <-readyToExecute
	s1 := gcecloudprovider.Signal{ReportError: true}
	execute <- s1
```

saikat-royc:
LGTM after a minor comment.

pkg/gce-pd-csi-driver/controller_test.go
mattcary: done

PR #1034, opened 2022-07-30
Point changelog to master branch

mattcary:
/kind cleanup
```release-note
None
```


mattcary:
/assign @leiyiz 

Force merging as this touches the changelog only (and we know the verify is broken here)

PR #1035, opened 2022-07-30
Cherry-pick #1028

mattcary:
Cherry-pick #1028
/kind bug
```release-note
```

/assign @leiyiz


mattcary:
argh, got the wrong branch

PR #1038, opened 2022-08-02
Change mistaken space to tab in makefile

mattcary:
/kind bug

Using  a space causes an  annoying spurious message to be printed during making.

```release-note
None
```

/assign @amacaskill


leiyiz (commented):
/lgtm
/approve
/retest

PR #1043, opened 2022-09-02
[wip] test: update kubetest2 version

spiffxp:
**What type of PR is this?**
/kind cleanup

WIP because:
- using this to exercise CI

This brings kubetest2 back up to date, and updates usage to follow the
breaking change that caused us to pin to an older version

Note that this will prevent test binaries from being uploaded into GCS
as part of CI results, but I suspect that wasn't intended to begin with

This picks up the following kubetest2 changes: https://github.com/kubernetes-sigs/kubetest2/compare/0e09086b...26f2492dc

The breaking change mentioned: https://github.com/kubernetes-sigs/kubetest2/pull/183

Which we pinned to avoid in: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1024

I'll open a PR in the other affected repo if I see CI pass here

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/main.go
+++ b/test/k8s-integration/main.go
@@ -725,10 +725,13 @@ func runTestsWithConfig(testParams *testParameters, testConfigArg, reportPrefix 
 		runID = string(uuid.NewUUID())
 	}
 
+	runDir := filepath.Join(testParams.pkgDir, "_rundir")
+
 	// Usage: kubetest2 <deployer> [Flags] [DeployerFlags] -- [TesterArgs]
 	// [Flags]
 	kubeTest2Args := []string{
 		*deploymentStrat,
+		fmt.Sprintf("--rundir=%s", runDir),
 		fmt.Sprintf("--run-id=%s", runID),
 		"--test=ginkgo",
 	}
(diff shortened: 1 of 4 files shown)


spiffxp:
Looks like kubetest2 is more convoluted than I thought:

> copying /tmp/gcp-pd-driver-tmp1442172395/kubernetes/_output/dockerized/bin/linux/amd64/kubectl to /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/_rundir/2385803a-2b05-11ed-92a0-caaedb0733ea/kubectl

vs.

> failed to run ginkgo tester: failed to validate pre-built binary kubectl (checked at "/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/deploy/kubernetes/overlays/stable-master/_rundir/2385803a-2b05-11ed-92a0-caaedb0733ea/kubectl"): stat _rundir/2385803a-2b05-11ed-92a0-caaedb0733ea/kubectl: no such file or directory

Not sure where `deploy/kubernetes/overlays/stable-master` is coming from

mauriciopoppe:
/hold

There are a few files in #1046 that also need to be updated after #1046 is merged.

mattcary:
> 
> EDIT: turns out it comes from
> 
> https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/162f146e698588ce154ed49f3117651541224165/test/k8s-integration/driver.go#L30
> 
> Since I'm not sure where the appropriate place to undo that chdir would be, I'll just point kubetest2's `--rundir` at an absolute dir

Hmm, I suspect we should be able to run customize from the package root (?). Anyway an absolutedir sgtm.

judemars:

PR #1051, opened 2022-09-23
Disable device when unstaging

mattcary:
/kind bug

We have learned that not disabling a device before detaching it can lead to device and VM instability. 

```release-note
Disable devices in node unstage prior to detaching.
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -364,6 +364,13 @@ func (ns *GCENodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
 		return nil, status.Error(codes.Internal, fmt.Sprintf("NodeUnstageVolume failed: %v\nUnmounting arguments: %s\n", err, stagingTargetPath))
 	}
 
+	devicePath, err := getDevicePath(ns, volumeID, "" /* partition, which is unused */)
+	if err != nil {
+		klog.Errorf("Failed to find device path for volume %s. Device may not be detached cleanly (error is ignored and unstaging is continuing): %v", volumeID, err)
+	} else if err := ns.DeviceUtils.DisableDevice(devicePath); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/gce-pd-csi-driver/utils_windows.go
```diff
@@ -93,6 +93,11 @@ func getDevicePath(ns *GCENodeServer, volumeID, partition string) (string, error
 	return proxy.GetDiskNumber(deviceName, partition, volumeKey.Name)
 }
 
+func disableDevice(devicePath string) error {
+	// This is a no-op on windows.
+	return nil
```
w-michal, line 98: Technically, this can be supported, see https://social.technet.microsoft.com/Forums/en-US/d99024c1-75d0-4344-9cce-8a7b895d1885/how-to-safely-remove-hardware-under-hyperv-2008-r2?forum=winservercore But I was not able to repro instability under Windows.

mattcary: Ok, thanks. I'll get this in on linux first

pkg/gce-pd-csi-driver/node.go
```diff
@@ -364,6 +364,13 @@ func (ns *GCENodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
 		return nil, status.Error(codes.Internal, fmt.Sprintf("NodeUnstageVolume failed: %v\nUnmounting arguments: %s\n", err, stagingTargetPath))
 	}
 
+	devicePath, err := getDevicePath(ns, volumeID, "" /* partition, which is unused */)
+	if err != nil {
+		klog.Errorf("Failed to find device path for %s. Device may not be detached cleanly (error is ignored and unstaging is continuing): %v", volumeID, err)
```
saikat-royc, line 369: Failed to find device path for volume %s

mattcary: Good point, thx

PR #1066, opened 2022-10-21
Support add the snapshot labels to images

jenting:
**What type of PR is this?**

/kind feature

We use labels to filter the GCP images.

Fixes #

Add parameters to your VolumeSnapshotClass like so:

```yaml
apiVersion: snapshot.storage.k8s.io/v1
kind: VolumeSnapshotClass
metadata:
  name: csi-gce-pd-snapshot-class
parameters:
  labels: cluster=preview,region=us,custom-label=awesome
  snapshot-type: images
driver: pd.csi.storage.gke.io
deletionPolicy: Delete
```

image

```release-note
Support setting snapshot labels to Images
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/fake-gce.go
+++ b/pkg/gce-cloud-provider/compute/fake-gce.go
@@ -424,6 +424,7 @@ func (cloud *FakeCloudProvider) CreateImage(ctx context.Context, project string,
 		SourceType:        "RAW",
 		Status:            "PENDING",
 		StorageLocations:  snapshotParams.StorageLocations,
+		Labels:            snapshotParams.Labels,
 	}
 
 	switch volKey.Type() {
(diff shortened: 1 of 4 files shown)


jenting:
@mattcary PTAL

mattcary:
/ok-to-test

I'll take a look at this shortly, thanks for the ping.

mattcary:
Ah, I didn't realize we just forgot to do the same with with disk image labels that we do with snapshots. Thanks for this change!

It looks like we already have unit tests in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/common/parameters_test.go; can you add a sort-of-e2e test by adding labels to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/config/image-volumesnapshotclass.yaml?

I don't think it will be easy to validate the labels are applied, but at least we'll catch if things will crash if they're used.

jenting:
> can you add a sort-of-e2e test by adding labels to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/k8s-integration/config/image-volumesnapshotclass.yaml?

I added to `pd-volumesnapshotclass.yaml` and `image-volumesnapshotclass.yaml`.

mattcary:
Awesome, thanks again for your PR.

/lgtm
/approve

PR #1071, opened 2022-10-31
Migrate to csi proxy v2 alpha

### LINKED ISSUE #1070: Migrate CSI Proxy to v2
[KEP 3636](https://github.com/kubernetes/enhancements/issues/3636) migrates CSI Proxy away from releasing a separate Windows binary to directly issuing privileged storage commands in drivers by taking advantage of [HostProcess containers](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/). After the KEP is approved, drivers should migrate to run as HostProcess containers and import the v2 package.

alexander-ding:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Migrates GCP Compute PD CSI Driver to use CSI Proxy v2. Besides updating the Go package code to use the new library, we also update the deployment files to run the driver as HostProcess containers. See the linked issue.

Fixes #1070 

This is a work in progress. We still need to update documentation to reflect the increased minimal Kubernetes version.

```release-note
Migrated Windows CSI node plugin to HostProcess containers

Action required: Ensure that the k8s version is 1.23+ and set the security policies
in the cluster to enable the HostProcess container feature only to system workloads.
```


mauriciopoppe:
/ok-to-test
/test pull-gcp-compute-persistent-disk-csi-driver-e2e-win2019

There's an optional presubmit job for Windows, please look for it in the test-infra codebase

mauriciopoppe:
It was not /test pull-gcp-compute-persistent-disk-csi-driver-e2e-win2019 :(

mauriciopoppe:
I haven't run /test pull-gcp-compute-persistent-disk-csi-driver-e2e-windows-2019 in a while so it might have drifted from the CI one, I'd take that one and compare it with the presubmit

deploy/kubernetes/base/node_windows/node.yaml
```diff
@@ -22,13 +22,21 @@ spec:
       serviceAccountName: csi-gce-pd-node-sa-win
       nodeSelector:
         kubernetes.io/os: windows
+      hostNetwork: true
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\SYSTEM" # TODO: customize
```
mauriciopoppe, line 29: Remove TODO

Dockerfile.Windows
```diff
@@ -24,6 +24,8 @@ RUN cd /code/ && GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAG
 
 FROM ${BASE_IMAGE}
```
mauriciopoppe, line 25: A followup item is to use the new base image that sig-windows created in https://static.sched.com/hosted_files/kccncna2022/c6/HPC-talk.pdf

Dockerfile.Windows
```diff
@@ -24,6 +24,8 @@ RUN cd /code/ && GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAG
...
 LABEL description="PD CSI driver"
+ENV PATH="${PATH};c:\\Windows\\System32\\WindowsPowerShell\\v1.0"
```
mauriciopoppe, line 27: This might not be needed after the new base image

mauriciopoppe (commented):
/lgtm
/assign @mattcary

mauriciopoppe:
The last run https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver#windows-2019-presubmit-gcp-compute-persistent-disk-csi-driver is red but here it's green, probably you can check again after removing the TODO

mauriciopoppe:
One more thing, I created a few tracking issues a while ago in this repo too, I think we should use those

https://github.com/kubernetes-csi/csi-proxy/issues/217 tracks the issues in this repo

mauriciopoppe (commented):
In the release note it should say something along the lines of:

```
Migrated Windows CSI node plugin to HostProcess containers

Action required: Ensure that the k8s version is 1.23+ and set the security policies
in the cluster to enable the HostProcess container feature only to system workloads
```

mattcary:
Thanks for the PR @alexander-ding !

/approve

I'll let @mauriciopoppe give the final lgtm when all the comments are fixed.

mattcary:
Awesome, thanks for following through.

I'll let Mauricio give the final lgtm, my approve should still be good.

alexander-ding:
Test failed but it seemed like a flake. The wait condition seems kind of weird. Also, for some reason the windows test has no test history. 

I had to run retest earlier for the regular e2e test as well due to a flake.

Dockerfile.Windows
```diff
@@ -22,9 +21,8 @@ WORKDIR /code
 ADD . /code/
 RUN cd /code/ && GOARCH=$(echo $TARGETPLATFORM | cut -f2 -d '/') GCE_PD_CSI_STAGING_VERSION=${STAGINGVERSION} make gce-pd-driver-windows
 
-FROM ${BASE_IMAGE}
+FROM mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0
```
mauriciopoppe, line 24: This is great, thanks a lot for digging through it and to make it work.

Makefile
```diff
@@ -55,26 +48,24 @@ build-container: require-GCE_PD_CSI_STAGING_IMAGE require-GCE_PD_CSI_STAGING_VER
...
 		--build-arg STAGINGVERSION=$(STAGINGVERSION) --push .
 
-build-and-push-multi-arch: build-and-push-container-linux-amd64 build-and-push-container-linux-arm64 build-and-push-windows-container-ltsc2019 build-and-push-windows-container-20H2
-	$(DOCKER) manifest create --amend $(STAGINGIMAGE):$(STAGINGVERSION) $(STAGINGIMAGE):$(STAGINGVERSION)_linux_amd64 $(STAGINGIMAGE):$(STAGINGVERSION)_linux_arm64 $(STAGINGIMAGE):$(STAGINGVERSION)_20H2 $(STAGINGIMAGE):$(STAGINGVERSION)_ltsc2019
-	STAGINGIMAGE="$(STAGINGIMAGE)" STAGINGVERSION="$(STAGINGVERSION)" WINDOWS_IMAGE_TAGS="$(WINDOWS_IMAGE_TAGS)" WINDOWS_BASE_IMAGES="$(WINDOWS_BASE_IMAGES)" ./manifest_osversion.sh
+build-and-push-multi-arch: build-and-push-container-linux-amd64 build-and-push-container-linux-arm64 build-and-push-container-windows-ltsc2019
+	$(DOCKER) manifest create --amend $(STAGINGIMAGE):$(STAGINGVERSION) $(STAGINGIMAGE):$(STAGINGVERSION)_linux_amd64 $(STAGINGIMAGE):$(STAGINGVERSION)_linux_arm64 $(STAGINGIMAGE):$(STAGINGVERSION)_windows_ltsc2019
+	$(DOCKER) manifest annotate --os-version 10.0.17763 $(STAGINGIMAGE):$(STAGINGVERSION) $(STAGINGIMAGE):$(STAGINGVERSION)_windows_ltsc2019
```
mauriciopoppe, line 63: A comment saying how you got `10.0.17763` will be useful if we need to update this later

deploy/kubernetes/base/node_windows/node.yaml
```diff
@@ -22,13 +22,21 @@ spec:
       serviceAccountName: csi-gce-pd-node-sa-win
```
mauriciopoppe, line 22: I remember that this codebase has overlays https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/README.md, this change should be only in 1.23+ (or whichever version had HPC enabled by default)

deploy/kubernetes/images/stable-1-21/image.yaml
```diff
@@ -1,53 +0,0 @@
-apiVersion: builtin
```
mauriciopoppe, line 1: Is there a reason why these files were removed? I think they are still in use. I think we could take a look at modifying these files in later commits.

mauriciopoppe (commented):
@mattcary we're trying to make changes in the base manifests but are unsure about modifying the overlays here and also the changes in the internal manifests, I thought about a couple of options and wanted to know your thoughts

- modify the base manifests with v2 and add overlays for low k8s versions that add back the v1 config
- modify the overlays with v2 versions in the k8s version that support HPC

Is there another option that you suggest?

mattcary:
> @mattcary we're trying to make changes in the base manifests but are unsure about modifying the overlays here and also the changes in the internal manifests, I thought about a couple of options and wanted to know your thoughts
> 
> * modify the base manifests with v2 and add overlays for low k8s versions that add back the v1 config
> * modify the overlays with v2 versions in the k8s version that support HPC
> 
> Is there another option that you suggest?

I'd prefer not to minimize touching the old version manifests, because we don't have much in the way of testing around them I think to confirm that they actually work. So I think it's safer to add overlays for dev, prow, stable-master for v2.

Then maybe when we start rolling off old versions (eg 1.21 and 1.22 should drop soon as they are EOL, 1.23 is going EOL early next year) we can switch and have an overlay for the v1 and have v2 in the base.

I think this also means you should copy stable-master to a new 1.25 version that uses v1, then have stable-master with an overlay for v2.

Does that all make sense?

alexander-ding:
> > @mattcary we're trying to make changes in the base manifests but are unsure about modifying the overlays here and also the changes in the internal manifests, I thought about a couple of options and wanted to know your thoughts
> > 
> > * modify the base manifests with v2 and add overlays for low k8s versions that add back the v1 config
> > * modify the overlays with v2 versions in the k8s version that support HPC
> > 
> > Is there another option that you suggest?
> 
> I'd prefer not to minimize touching the old version manifests, because we don't have much in the way of testing around them I think to confirm that they actually work. So I think it's safer to add overlays for dev, prow, stable-master for v2.
> 
> Then maybe when we start rolling off old versions (eg 1.21 and 1.22 should drop soon as they are EOL, 1.23 is going EOL early next year) we can switch and have an overlay for the v1 and have v2 in the base.
> 
> I think this also means you should copy stable-master to a new 1.25 version that uses v1, then have stable-master with an overlay for v2.
> 
> Does that all make sense?

@mattcary to make sure we're on the same page, you're suggesting that we

1. Keep the base manifests as is
2. Rename stable-master to stable-1.25
3. Create a new stable-master that overlays v2 stuff on the base manifest
4. Modify dev, noauth, and prow overlays to use v2 as well

Is that right?

Note that the windows node overlay for v2 would be pretty involved, and we have to copy that for every overlay that needs v2 (and remove it when we move the base manifest to use v2). If that's okay, I can go ahead with that

mauriciopoppe:
thanks for the feedback @mattcary

@alexander-ding I think that the steps you wrote look good to me, I think that it's easy to add things because kustomize will just merge the manifests, it's a little bit difficult to delete things though (e.g. the sections in the Spec that mount and use the named pipes) but it's also doable.

mattcary:
> 1. Keep the base manifests as is
> 2. Rename stable-master to stable-1.25
> 3. Create a new stable-master that overlays v2 stuff on the base manifest
> 4. Modify dev, noauth, and prow overlays to use v2 as well
> 
Oops, sorry for the lag in reply.

Yes, those 4 steps are what I had in mind.


> Is that right?
> 
> Note that the windows node overlay for v2 would be pretty involved, and we have to copy that for every overlay that needs v2 (and remove it when we move the base manifest to use v2). If that's okay, I can go ahead with that

Yeah, understood. I think it's important to keep the base simple while we're tweaking the v2 overlay ---we need to be able to easily make changes to the production overlays which mean 1.23-1.25 at this point.

When v2 development gets stable then we can switch over.

alexander-ding:
@mattcary @mauriciopoppe I pushed some changes with the overlays. However, I noticed that the stable overlays all point to specific image tags, but we don't have a fixed version tag for the latest development image. 

I did some more readings on the release process for this repository, namely the [overlay strategies](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/docs/release/overlays.md) and go/pdcsi-oss-release-process (linked to by the repository README). I think we should take a step back and talk through the release pipeline.

Strictly speaking, this PR doesn't introduce new functionalities, so it can be considered a patch, not a new minor release. In this case, this PR should only update the development overlays (`dev` and `noauth-debug`) and _nothing else_. Then, we need to merge this PR to master and cherry pick the change into the release 1.8 branch, which is currently in prerelease, and release the updated driver as v1.8.0-rc2.

Once the new prerelease is generated, we can create a separate PR to update the `prow-stable-sidecar-rc-master` overlay to use the rc2 prerelease and ensure that tests are green.

When ready to release, we can update all the relevant production overlays (`stable-1-23`, `stable-1-24`, and `stable-master`) to be the same as the `prow-stable-sidecar-rc-master` staging overlay. 

Releasing `stable-1-25` should be a separate process. 

If we decide that this is a big enough change to merit its own minor version, then this release needs to be v1.9.0. We would have to go through the entire pre-release/release process. We should probably wait until v1.8 is stable before merging to master. 

Let me know what you think. I'll go ahead and revert to the overlay changes I outlined (only `dev` and `noauth-debug`).

mattcary:
Good catch.

You're right that to update stable we need a new release. So we shouldn't update those yet.

We will want a minor version bump with these v2 changes as the ability to use host processes is a new feature.

We have already [released 1.8](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/releases/tag/v1.8.0), and it's stable; the stable-master images should have been updated but doing that is down low on our todo list.

The dev overlay doesn't point to a particular image version because for testing we update it with a version built for a test run---that's how the CI pipelines test against master.

So, yeah, let's just update dev and noauth for now. Then when we cut a 1.9 release we can update the prow and stable configs.

I guess we don't have a strategy for testing all the version skew scenarios from this repo?

alexander-ding:
> Good catch.
> 
> You're right that to update stable we need a new release. So we shouldn't update those yet.
> 
> We will want a minor version bump with these v2 changes as the ability to use host processes is a new feature.
> 
> We have already [released 1.8](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/releases/tag/v1.8.0), and it's stable; the stable-master images should have been updated but doing that is down low on our todo list.
> 
> The dev overlay doesn't point to a particular image version because for testing we update it with a version built for a test run---that's how the CI pipelines test against master.
> 
> So, yeah, let's just update dev and noauth for now. Then when we cut a 1.9 release we can update the prow and stable configs.

Sounds good. Let's make this 1.9. Also, thanks for the pointer on the dev overlay. I was worried that not updating `stable-master` would cause the tests to break. 

> I guess we don't have a strategy for testing all the version skew scenarios from this repo?

I don't know about all the testing infra in place, so I can't really speak to it. Maybe @mauriciopoppe would know?

mauriciopoppe:
/retest

I think that modifying the deployment manifests to be HostProcess containers is a feature and I agree we should do this in 1.9.

The infra for windows https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml uses [stable-master](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/run-windows-k8s-integration.sh#L14) so not changing it would be a problem in CI (we still wouldn't be testing the new HostProcess container version). We don't have any test infra for other overlays for Windows.

mauriciopoppe:
I read the logs of pull-gcp-compute-persistent-disk-csi-driver-e2e-windows-2019 and it's using stable-master with the image overridden to be the one built from source, I also checked in the logs that the Windows DaemonSet is not a HostProcess container.

This means that presubmits tests being green doesn't represent what we want to test, I think we could change this in test-infra (setting the overlay to dev) or changing stable-master so that the DaemonSet is a HostProcess container.

alexander-ding:
> I read the logs of pull-gcp-compute-persistent-disk-csi-driver-e2e-windows-2019 and it's using stable-master with the image overridden to be the one built from source, I also checked in the logs that the Windows DaemonSet is not a HostProcess container.
> 
> This means that presubmits tests being green doesn't represent what we want to test, I think we could change this in test-infra (setting the overlay to dev) or changing stable-master so that the DaemonSet is a HostProcess container.

Thanks for noting this. That's really good to know. 

That's pretty concerning that the tests pass at all then. If it is using stable-master with the image built from source, then it should be running CSI Proxy v2 commands and failing, since it's not running in an HPC. The test is passing though, for some reason?

We should definitely go ahead and update the Windows tests, but we should probably also find out why it's passing as is.

alexander-ding:
I'd say using the dev overlay in the Windows test should be sufficient.

mauriciopoppe:
> That's pretty concerning that the tests pass at all then. If it is using stable-master with the image built from source, then it should be running CSI Proxy v2 commands and failing, since it's not running in an HPC. The test is passing though, for some reason?

Yes, I didn't go deeper into the why.

> I'd say using the dev overlay in the Windows test should be sufficient.

Both linux and windows are running stable-master in presubmit which is also good (both have the same deployment model), for testing purposes we could:

- make both linux/windows test the dev overlay
  - we're also modifying the linux path, @mattcary would also need to approve this
- add a temporary Windows presubmit optional job that uses the dev overlay
  - this is easier for us to create a temp job and test during presubmit

I think we could also have many flavors of CI (one testing stable-master, another dev) too, I'd go for the 2nd option if you want to check that your changes work

mattcary:
I'd lean towards using a single test config in dev/ for presubmit. We can test stable-master in the [CI](https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver). 

Running on stable master, the v2 code should fall back to using the proxy, right---that's how it's supposed work in a node skew situation?

alexander-ding:
> 



> I'd lean towards using a single test config in dev/ for presubmit. We can test stable-master in the [CI](https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver).

I agree. I feel like presubmits on PRs to master should focus on the dev overlay. WDYT, @mauriciopoppe ?

> Running on stable master, the v2 code should fall back to using the proxy, right---that's how it's supposed work in a node skew situation?

It doesn't. There's no flag toggling behavior, and the v2 code only uses the v2 CSI proxy.

mattcary:
> 
> It doesn't. There's no flag toggling behavior, and the v2 code only uses the v2 CSI proxy.

Wait, how is this supposed to work for upgrades? ie, an existing cluster has the v1 proxy, you upgrade your master, the nodes stay the same, the upgrade changes the daemonset (on all nodes) --- doesn't that break everything?

alexander-ding:
> > It doesn't. There's no flag toggling behavior, and the v2 code only uses the v2 CSI proxy.
> 
> Wait, how is this supposed to work for upgrades? ie, an existing cluster has the v1 proxy, you upgrade your master, the nodes stay the same, the upgrade changes the daemonset (on all nodes) --- doesn't that break everything?

Thanks for spotting this. Just read up on version skews in detail. I think you're right. If the nodes are 1.22 or earlier, then things would break. I guess one workaround is to not enable v2 for anything under K8s 1.25, since the version skew policy supports 2 minor versions behind, and HPCs are beta in 1.23. 

We can also just implement a toggle (env var or flag?) and have the defaulting behavior you described. Shouldn't be too tricky. I'm happy to do it.

alexander-ding:
The overlay updates are merged in `test-infra`. Retesting
/retest
/test pull-gcp-compute-persistent-disk-csi-driver-e2e-windows-2019

judemars:

mauriciopoppe:
> I'd lean towards using a single test config in dev/ for presubmit. We can test stable-master in the [CI](https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver).

I saw that:
- presubmit uses the dev overlay (both [Linux](https://github.com/kubernetes/test-infra/blob/8653528407b935cffce025c018af16ea91fb8a0a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/config.yaml#L115) and [Windows](https://github.com/kubernetes/test-infra/blob/8653528407b935cffce025c018af16ea91fb8a0a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml#L79))
- periodics use the stable-master overlay ([Linux](https://github.com/kubernetes/test-infra/blob/8653528407b935cffce025c018af16ea91fb8a0a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-postsubmits.yaml#L21) and Windows doesn't set a value so it uses the default stable-master)

In the presubmit runs we saw that the printed logs for `kustomize build <dev overlay>` don't have the changes HPC changes which is weird, @alexander-ding is looking at that

alexander-ding:
@mattcary I talked about the overlay issue with @mauriciopoppe some more, and I don't think there's an easy way to support version skew scenarios like this.

IIUC, once we upgrade the control plane to use the v2 overlay, it changes every node to use the new daemonset pod spec, which runs as HPC. If the worker nodes are under 1.23, they simply cannot understand the HPC fields in the pod manifest and can't run. Adding a flag to control the v2/v1 behavior in the code doesn't help with this. We can't flex the manifest to have HPC fields on nodes >=1.23 and not have HPC fields on nodes < 1.23.

In that case, we should just only use HPC manifests on K8s (control plane) 1.25 and onwards, since worker nodes have to be at least 1.23 in these clusters. 

Is this understanding correct? If so, I think our current plan is good.

alexander-ding:
> > I'd lean towards using a single test config in dev/ for presubmit. We can test stable-master in the [CI](https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver).
> 
> I saw that:
> 
> * presubmit uses the dev overlay (both [Linux](https://github.com/kubernetes/test-infra/blob/8653528407b935cffce025c018af16ea91fb8a0a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/config.yaml#L115) and [Windows](https://github.com/kubernetes/test-infra/blob/8653528407b935cffce025c018af16ea91fb8a0a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml#L79))
> * periodics use the stable-master overlay ([Linux](https://github.com/kubernetes/test-infra/blob/8653528407b935cffce025c018af16ea91fb8a0a/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-postsubmits.yaml#L21) and Windows doesn't set a value so it uses the default stable-master)
> 
> In the presubmit runs we saw that the printed logs for `kustomize build <dev overlay>` don't have the HPC changes which is weird, @alexander-ding is looking at that

The other thing is that the Windows test logs are not persisted, which makes debugging extra tricky. I'm not very familiar with how the testing framework works under the hood. Is this intended?

mattcary:
> There is a kubelet version in the node system info, but I don't think there can be a selector based on that?

Yeah it seems only labels are used [1](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2569), [2](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector)

alexander-ding:
Okay, this took way too long, but I think I figured out what is going on with the Windows tests. The configs on `test-infra` are a bit messed up.

It's always using the default branch (instead of the PR). See [this line](https://github.com/kubernetes/test-infra/blob/ca6fa9cca4d94116d2ec09dcfbb7ed73bb05bff7/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml#L66). It should be `sigs.k8s.io/$(REPO_NAME)=$(PULL_REFS)`, like the Linux presubmit jobs.

The log persistence issue seems to be because the [upload path](https://github.com/kubernetes/test-infra/blob/ca6fa9cca4d94116d2ec09dcfbb7ed73bb05bff7/config/jobs/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/gcp-compute-persistent-disk-csi-driver-windows.yaml#L68) is set to `gs://kubernetes-jenkins/logs`, instead of `kubernetes-jenkins/pr-logs` like the Linux presubmit jobs.

I'll go open a PR to fix this.

alexander-ding:
> I think you're right about the manifest. The only way to flex them that I know if is to have a node selector (eg, the same way we apply the windows manifest only to windows nodes). There is a kubelet version in the node system info, but I don't think there can be a selector based on that?

Yeah, and I think it'd be way too complicated for a non-critical feature anyways. Let's stick to supporting only 1.25+.

alexander-ding:
I think the Windows tests are timing out by the way, since building takes a really long time. I'll go update the timeout.

PR #1075, opened 2022-11-11
Update version parsing for beta prereleases

mattcary:
/kind failing-test
```release-note
None
```


mattcary:
The e2e test fails due to a KMS project error; I've filed a bug with engprod.

/test/pull-gcp-compute-persistent-disk-csi-driver-e2e

PR #1077, opened 2022-11-12
Add compute-endpoint flag to pdcsi driver

judemars:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

Adds ability to pick up compute endpoint from compute-endpoint flag rather than using hard-coded value. This is needed to achieve service regionalization

Fixes #

This is my first change so still working out which tests to run.

```release-note
NONE
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -36,6 +36,7 @@ import (
 var (
 	cloudConfigFilePath  = flag.String("cloud-config", "", "Path to GCE cloud provider config")
 	endpoint             = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	computeEndpoint 		 = flag.String("compute-endpoint", "", "If set, used as the endpoint for the GCE API.")
 	runControllerService = flag.Bool("run-controller-service", true, "If set to false then the CSI driver does not activate its controller service (default: true)")
 	runNodeService       = flag.Bool("run-node-service", true, "If set to false then the CSI driver does not activate its node service (default: true)")
 	httpEndpoint         = flag.String("http-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
@@ -117,7 +118,7 @@ func handle() {
 	//Initialize requirements for the controller service
 	var controllerServer *driver.GCEControllerServer
 	if *runControllerService {
-		cloudProvider, err := gce.CreateCloudProvider(ctx, version, *cloudConfigFilePath)
+		cloudProvider, err := gce.CreateCloudProvider(ctx, version, *cloudConfigFilePath, *computeEndpoint)
 		if err != nil {
 			klog.Fatalf("Failed to get cloud provider: %v", err)
 		}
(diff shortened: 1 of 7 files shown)


mattcary:
Does it make sense to add an e2e test to confirm that passing an endpoint in actually works? IIUC currently no test (e2e or integration) will actually exercise this flag, so we run the risk of accidentally breaking it.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -159,30 +155,40 @@ func readConfig(configPath string) (*ConfigFile, error) {
...
+func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSource oauth2.TokenSource, computeEndpoint string) (*computebeta.Service, error) {
 	client, err := newOauthClient(ctx, tokenSource)
 	if err != nil {
 		return nil, err
 	}
-	service, err := computebeta.NewService(ctx, option.WithHTTPClient(client))
+
+	computeOpts := []option.ClientOption{option.WithHTTPClient(client)}
+	if computeEndpoint != "" {
+		computeOpts = append(computeOpts, option.WithEndpoint(computeEndpoint))
```
leiyiz, line 166: Sorry if I've missed convo on this: does it make sense to have a validation on the endpoint passed in?

cmd/gce-pd-csi-driver/main.go
```diff
@@ -36,6 +36,7 @@ import (
 var (
 	cloudConfigFilePath  = flag.String("cloud-config", "", "Path to GCE cloud provider config")
 	endpoint             = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
+	computeEndpoint      = flag.String("compute-endpoint", "", "If set, used as the endpoint for the GCE API.")
```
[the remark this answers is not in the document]
judemars: If no endpoint is passed into the computebeta.NewService constructor, the default "https://compute.googleapis.com/compute/beta/" is used. This value is being used in the current implementation and the new.

[1]: see "basePath" in https://raw.githubusercontent.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/7592dc19d3bd000f34cf950454c3355ae35f1475/vendor/google.golang.org/api/compute/v0.beta/compute-gen.go

pkg/gce-cloud-provider/compute/gce.go
judemars: What kind of validation did you have in mind? For GKE this endpoint will be passed in from a trusted place.

judemars:
@mattcary for some reason I can't comment on your suggestion re: e2e tests. I think adding them is a great idea but it'll take me a bit to get them up and running locally. Okay if I file a ticket to add an e2e test and do it in a follow up?

judemars:
Alrighty @mattcary I added an e2e test. I couldn't find a way to pass in a valid endpoint and then verify the exact endpoint was passed through apart from just checking for success, but if you know of a way to check that met me know and I'll add it. I opted to go with:

1. Pass in nothing (i.e. use the default) see it succeed in a call to gcs
2. Use an invalid one and see it fail w/ "no such host"
3. Pass in the default one and see it succeed w/ a call to gcs

mattcary:
Awesome, that's solid.

/lgtm
/approve

PR #1079, opened 2022-11-14
Add provisionedIops for pd-extreme

sunnylovestiramisu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature

> /kind flake

Support pd-extreme with provisionedIops, please see [pd-extreme](https://cloud.google.com/compute/docs/disks/extreme-persistent-disk)

Fixes #

The current pd-extreme [restrictions](https://cloud.google.com/compute/docs/disks/extreme-persistent-disk):

* Extreme persistent disks are zonal only. You cannot create regional extreme persistent disks.
* You cannot attach multiple VM instances in read-only mode to an extreme persistent disk.
* You cannot [create an image or machine image](https://cloud.google.com/compute/docs/images/create-delete-deprecate-private-images#create_image) from an extreme persistent disk.
* You cannot [clone](https://cloud.google.com/compute/docs/disks/add-persistent-disk#clone) an extreme persistent disk.

```release-note

Pass in ProvisionedIOPSOnCreate as a parameter for CreateVolume

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,14 @@ See Github [Issues](https://github.com/kubernetes-sigs/gcp-compute-persistent-di
 
 ### CreateVolume Parameters
 
-| Parameter        | Values                    | Default       | Description                                                                                        |
-|------------------|---------------------------|---------------|----------------------------------------------------------------------------------------------------|
-| type             | Any PD type (see [GCP documentation](https://cloud.google.com/compute/docs/disks#disk-types)), eg `pd-ssd` `pd-balanced` | `pd-standard` | Type allows you to choose between standard Persistent Disks  or Solid State Drive Persistent Disks |
-| replication-type | `none` OR `regional-pd`   | `none`        | Replication type allows you to choose between Zonal Persistent Disks or Regional Persistent Disks  |
-| disk-encryption-kms-key | Fully qualified resource identifier for the key to use to encrypt new disks. | Empty string. | Encrypt disk using Customer Managed Encryption Key (CMEK). See [GKE Docs](https://cloud.google.com/kubernetes-engine/docs/how-to/using-cmek#create_a_cmek_protected_attached_disk) for details. |
-| labels           | `key1=value1,key2=value2` |               | Labels allow you to assign custom [GCE Disk labels](https://cloud.google.com/compute/docs/labeling-resources). |
+| Parameter                   | Values                    | Default       | Description                                                                                        |
+|-----------------------------|---------------------------|---------------|----------------------------------------------------------------------------------------------------|
+| type                        | Any PD type (see [GCP documentation](https://cloud.google.com/compute/docs/disks#disk-types)), eg `pd-ssd` `pd-balanced` | `pd-standard` | Type allows you to choose between standard Persistent Disks  or Solid State Drive Persistent Disks |
+| replication-type            | `none` OR `regional-pd`   | `none`        | Replication type allows you to choose between Zonal Persistent Disks or Regional Persistent Disks  |
+| disk-encryption-kms-key     | Fully qualified resource identifier for the key to use to encrypt new disks. | Empty string. | Encrypt disk using Customer Managed Encryption Key (CMEK). See [GKE Docs](https://cloud.google.com/kubernetes-engine/docs/how-to/using-cmek#create_a_cmek_protected_attached_disk) for details. |
+| labels                      | `key1=value1,key2=value2` |               | Labels allow you to assign custom [GCE Disk labels](https://cloud.google.com/compute/docs/labeling-resources). |
+| provisioned-iops-on-create  | string (int64 format). Values typically between 10,000 and 120,000 |               | Indicates how many IOPS to provision for the disk. See the [Extreme persistent disk documentation](https://cloud.google.com/compute/docs/disks/extreme-persistent-disk) for details, including valid ranges for IOPS. |
+
 
 ### Topology
 
--- a/pkg/common/parameters.go
+++ b/pkg/common/parameters.go
@@ -23,10 +23,11 @@ import (
 
 const (
 	// Parameters for StorageClass
-	ParameterKeyType                 = "type"
-	ParameterKeyReplicationType      = "replication-type"
-	ParameterKeyDiskEncryptionKmsKey = "disk-encryption-kms-key"
-	ParameterKeyLabels               = "labels"
+	ParameterKeyType                    = "type"
+	ParameterKeyReplicationType         = "replication-type"
+	ParameterKeyDiskEncryptionKmsKey    = "disk-encryption-kms-key"
+	ParameterKeyLabels                  = "labels"
+	ParameterKeyProvisionedIOPSOnCreate = "provisioned-iops-on-create"
 
 	// Parameters for VolumeSnapshotClass
 	ParameterKeyStorageLocations = "storage-locations"
@@ -80,6 +81,9 @@ type DiskParameters struct {
 	// Values: {map[string]string}
 	// Default: ""
 	Labels map[string]string
+	// Values: {int64}
+	// Default: none
+	ProvisionedIOPSOnCreate int64
 }
 
 // SnapshotParameters contains normalized and defaulted parameters for snapshots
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -58,12 +58,14 @@ See Github [Issues](https://github.com/kubernetes-sigs/gcp-compute-persistent-di
...
+| replication-type            | `none` OR `regional-pd`   | `none`        | Replication type allows you to choose between Zonal Persistent Disks or Regional Persistent Disks  |
+| disk-encryption-kms-key     | Fully qualified resource identifier for the key to use to encrypt new disks. | Empty string. | Encrypt disk using Customer Managed Encryption Key (CMEK). See [GKE Docs](https://cloud.google.com/kubernetes-engine/docs/how-to/using-cmek#create_a_cmek_protected_attached_disk) for details. |
+| labels                      | `key1=value1,key2=value2` |               | Labels allow you to assign custom [GCE Disk labels](https://cloud.google.com/compute/docs/labeling-resources). |
+| provisioned-iops-on-create  | string (int64 format). Values must be between 10,000 and 120,000 |               | Indicates how many IOPS to provision for the disk. See the [Extreme persistent disk documentation](https://cloud.google.com/compute/docs/disks/extreme-persistent-disk). |
```
mattcary, line 67: Let's give ourselves a little more flexibilty:

"string (int64 format. Values typically between 10,000 and 120,000" ... "See the Extreme persistent disk documentation for details, including valid ranges for IOPS.

pkg/common/parameters.go
```diff
@@ -135,6 +140,8 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 			for labelKey, labelValue := range paramLabels {
 				p.Labels[labelKey] = labelValue
 			}
+		case ParameterKeyProvisionedIOPSOnCreate:
+			p.ProvisionedIOPSOnCreate = v
```
mattcary, line 144: let's validate that this is an int64---see ConvertLabelsStringToMap, which does things like validate the characters used in the label. 

I think it would be best to validate the parameters as early as possible.

mattcary (commented):
Let's also add the changes to gce-compute in this same PR.

pkg/common/parameters.go
sunnylovestiramisu: Should we only validate int64 instead of also verifying the range? 

If we validate the range it means the range change in the future, we need to modify the validation logic each time and go through deployments.

judemars:

pkg/common/parameters.go
```diff
@@ -143,6 +147,12 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 			for labelKey, labelValue := range paramLabels {
 				p.Labels[labelKey] = labelValue
 			}
+		case ParameterKeyProvisionedIOPSOnCreate:
+			paramProvisionedIOPSOnCreate, err := ConvertStringToInt64(v)
```
mattcary, line 151: Don't we want to use resource quantity here so that we can convert strings like 100k? See k8s.io/apimachinery/pkg/api/resource

sunnylovestiramisu: What are we supporting here? Are we supporting both "100000GB" and "10000MB" or "100000GiB" and "10000MiB" and + "100k GB" ?

mattcary: I think the use case is to support 100k. I'm not sure if there's a k8s resource parsing api to specify that "k" is okay but "GiB" is not. So I think we should just use the permissive api.

eg you could dig and see what cpu does to parse---can you write 2GiB for a CPU request?

sunnylovestiramisu:
pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration [test](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/1079/pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration/1600929347751907328/) failed with 

```
I1208 20:16:57.033]     Dec  8 20:16:56.333: INFO: At 2022-12-08 20:11:39 +0000 UTC - event for csi-gcepd-sc-extreme--no-volumesnapshotclassjltvh-cloned: {pd.csi.storage.gke.io_e2e-test-prow-minion-group-5v0r_488ec745-8f4f-4b8e-bb07-e8ff64d95cb6 } ProvisioningFailed: failed to provision volume with StorageClass "multivolume-2380-e2e-sc9n4x2": rpc error: code = Internal desc = CreateVolume failed to create single zonal disk "pvc-2a6e4167-27ea-4eff-b8f7-e40b8fdd7137": failed to insert zonal disk: unknown Insert disk error: googleapi: Error 400: Cloning is not supported for disks of pd-extreme type., badRequest
```
which is expected because pd-extreme does not support cloning.

test/k8s-integration/driver-config.go
```diff
@@ -127,13 +127,9 @@ func generateDriverConfigFile(testParams *testParameters) (string, error) {
 		snapshotClassName = "no-volumesnapshotclass"
 	}
 
-	caps = append(caps, "pvcDataSource")
+	//caps = append(caps, "pvcDataSource")
```
mattcary, line 130: Since you add the sc-extreme storage class you can make the capability depend on the storage class choice, like we do for regional a couple lines further down.

test/k8s-integration/driver-config.go
```diff
@@ -127,13 +127,9 @@ func generateDriverConfigFile(testParams *testParameters) (string, error) {
...
 	minimumVolumeSize := "5Gi"
 	numAllowedTopologies := 1
-	if testParams.storageClassFile == regionalPDStorageClass {
```
mattcary, line 133: I don't think we should tear this out?

test/k8s-integration/main.go
```diff
@@ -479,7 +479,7 @@ func handle() error {
 	}
 
 	skipDiskImageSnapshots := false
-	if mustParseVersion(testParams.clusterVersion).lessThan(mustParseVersion("1.22.0")) {
+	if mustParseVersion(testParams.clusterVersion).lessThan(mustParseVersion("1.22.0")) || strings.Contains(testParams.storageClassFile, "sc-extreme") {
 		// Disk image cloning in only supported from 1.22 on.
```
mattcary, line 483: I don't think we need to add the skip for extreme, if we remove the capability in driver-config.go?

test/k8s-integration/driver-config.go
sunnylovestiramisu: pd-extreme does not support regional pd, I just remove this for testing, will revert this at submit.

test/k8s-integration/main.go
sunnylovestiramisu: This one is not doing anything, let me delete this. Initially i was thinking at submit we will revert the last commit [Test pd-extreme k8s integration](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1079/commits/252c7d29dfa9c817f827e82c17a7bcf78e152861) but we actually needed to submit it. Will modify the code

sunnylovestiramisu:
The k8s integration success run is in:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/1079/pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration/1603132775324454912/

The discussion is that we are not checking in the integration test but we can swap the storage class to ```--storageclass-files=sc-extreme.yaml``` in test/run-k8s-integration.sh

test/k8s-integration/driver-config.go
```diff
@@ -127,10 +127,12 @@ func generateDriverConfigFile(testParams *testParameters) (string, error) {
...
 	}
 
-	caps = append(caps, "pvcDataSource")
+	if strings.Contains(testParams.storageClassFile, "sc-standard") {
+		caps = append(caps, "pvcDataSource")
+	}
 	minimumVolumeSize := "5Gi"
 	numAllowedTopologies := 1
-	if testParams.storageClassFile == regionalPDStorageClass {
+	if testParams.storageClassFile == regionalPDStorageClass && !strings.Contains(testParams.storageClassFile, "sc-extreme") {
```
mattcary, line 135: This extreme check is redudent, because the regional storage class filename is already checked.

test/k8s-integration/driver-config.go
```diff
@@ -127,10 +127,12 @@ func generateDriverConfigFile(testParams *testParameters) (string, error) {
 		snapshotClassName = "no-volumesnapshotclass"
 	}
 
-	caps = append(caps, "pvcDataSource")
+	if strings.Contains(testParams.storageClassFile, "sc-standard") {
```
mattcary, line 130: I think we want to check pvcDataSource on everything but extreme, not just on standard.

 (if we picked one we'd do balanced anyway)

sunnylovestiramisu: I was thinking hyperdisk also does not support cloning?

mattcary: We'll cross that bridge when we get to it. I think some SKUs will.

PR #1083, opened 2022-11-17
Fix resize-related sanity test failures

judemars:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

Our sanity tests are currently failing 4 test cases: 1 because of go/pdcsi-oss-driver/issues/990 to be addressed in a follow up in another repo, and 3 others with the error:

```
Could not determine if filesystem \"/tmp/csi/stage\" needs to be resized: failed to parse size of device /dev/disk/fake-path
```
This PR fixes the latter 3 failures by mocking out the resize method.

Note this also required moving deviceutils into its own pkg to avoid a circular dependency from mountmanager (device utils) -> resize -> mountmanager (windows).

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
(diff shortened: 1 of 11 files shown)


mattcary:
/ok-to-test
/lgtm
/approve

(there's a bigger discussion around sanity checks on another thread)

PR #1085, opened 2022-11-18
limit grpc logging info to a configurable character limit

leiyiz:
**What type of PR is this?**
/kind bug

ListVolume call generates tons of log items because by default the entirety of the grpc response body gets logged out. That consumes log space really fast and we want to limit how much can be logged in the same message.

This might impact some visibility into what's going on with ListVolume call.

Fixes #

didn't find obvious way to make the log limit configurable. Pls advise of otherwise

```release-note
Logging for gRPC call is now limited to 10000 char by default, and configurable with flag --grpc-log-char-cap
```

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -69,7 +69,11 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
 	if err != nil {
 		klog.Errorf("%s returned with error: %v", info.FullMethod, err)
 	} else {
-		klog.V(4).Infof("%s returned with response: %s", info.FullMethod, resp)
+		cappedStr := fmt.Sprintf("%v", resp)
+		if len(cappedStr) > maxLogChar {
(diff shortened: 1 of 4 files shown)


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -28,7 +28,8 @@ import (
 )
 
 const (
-	fsTypeXFS = "xfs"
+	fsTypeXFS  = "xfs"
+	maxLogChar = 10000
```
mattcary, line 32: Do you think it's not too messy to plumb this limit through with a command-line flag? Then it it turns out 10k is still too large we could easily ratchet it down with just a yaml tweak and not a code change + version release.

leiyiz: done

amacaskill:
lgtm but just looks like you need to change the tests that call gceDriver.Run(endpoint) --> gceDriver.Run(endpoint, grpcLogCharCap)

leiyiz:
fixed test failure, @amacaskill ptal

judemars:

PR #1087, opened 2022-11-21
Lower threshold on backoff for sanity tests

### LINKED ISSUE #990: Sanity test do not work with the PDCSI driver attach detach behavior change
#988 introduces a behavior change in PDCSI driver controller publish/unpublish call. If any error is encountered in the controller publish or unpublish path, the driver marks the target node with a backoff condition. The implication of this is that, until the backoff time has expired, all ControllerPublish/Unpublish calls for the target node will be directly denied by the driver.

The sanity tests runs a series of negative/positive test cases on the same underlying driver. This means if a negative test case for controller publish or unpublish runs first (e.g `ControllerPublishVolume should fail when the volume is already published but is incompatible` [code](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/vendor/github.com/kubernetes-csi/csi-test/v4/pkg/sanity/controller.go#L1000)) this failure causes the driver to mark the target node in a backoff condition. This causes the next postive controller publish test to fail with error (`Controller Service [Controller Server] should work` [code](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/vendor/github.com/kubernetes-csi/csi-test/v4/pkg/sanity/controller.go#L1063))
```
rpc error: code = Unavailable desc = ControllerPublish not permitted on node "projects/test-project/zones/country-region-zone/instances/test-name" due to backoff condition
```

the test cases can fail randomly based on the sequence of the test cases run.

See a failed run [here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/sigs.k8s.io_gcp-compute-persistent-disk-csi-driver/988/pull-gcp-compute-persistent-disk-csi-driver-sanity/1526688796463075328/)

judemars:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

The [scenario of the failing test](https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/controller.go#L1001)  ("should fail when the volume is already published but is incompatible"):

1. We create a volume -> succeeds
2. We publish that volume -> succeeds
3. We re-publish that volume, expecting an error bc it was already published -> fails (and triggers backoff)
4. [Cleanup function](https://github.com/kubernetes-csi/csi-test/blob/e3d351ac3db771d36893f3ee74df1c687a4071ca/pkg/sanity/resources.go#L270) tries to unpublish volume, and this fails the test case with the error `ControllerUnpublish not permitted on node "projects/test-project/zones/country-region-zone/instances/test-name" due to backoff condition` because of Step 3.

So, the solution we decided in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/990 was to lower the duration of backoff for sanity tests so that the backoff triggered in step 3 above will not cause step 4 to fail. 

See https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/990.

Fixes #990 

```release-note
NONE
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -42,7 +42,9 @@ var (
 	httpEndpoint         = flag.String("http-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath          = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
-	extraVolumeLabelsStr = flag.String("extra-labels", "", "Extra labels to attach to each PD created. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'. See https://cloud.google.com/compute/docs/labeling-resources for details")
+	errorBackoffInitialDurationMs = flag.Int("backoff-initial-duration-ms", 200, "The amount of ms for the initial duration of the backoff condition for controller publish/unpublish CSI operations. Default is 200.")
+	errorBackoffMaxDurationMs     = flag.Int("backoff-max-duration-ms", 300000, "The amount of ms for the max duration of the backoff condition for controller publish/unpublish CSI operations. Default is 300000 (5m).")
+	extraVolumeLabelsStr          = flag.String("extra-labels", "", "Extra labels to attach to each PD created. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'. See https://cloud.google.com/compute/docs/labeling-resources for details")
 
 	attachDiskBackoffDuration = flag.Duration("attach-disk-backoff-duration", 5*time.Second, "Duration for attachDisk backoff")
 	attachDiskBackoffFactor   = flag.Float64("attach-disk-backoff-factor", 0.0, "Factor for attachDisk backoff")
@@ -122,7 +124,9 @@ func handle() {
 		if err != nil {
 			klog.Fatalf("Failed to get cloud provider: %v", err)
 		}
-		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider)
+		initialBackoffDuration := time.Duration(*errorBackoffInitialDurationMs) * time.Millisecond
+		maxBackoffDuration := time.Duration(*errorBackoffMaxDurationMs) * time.Microsecond
+		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider, initialBackoffDuration, maxBackoffDuration)
 	} else if *cloudConfigFilePath != "" {
 		klog.Warningf("controller service is disabled but cloud config given - it has no effect")
 	}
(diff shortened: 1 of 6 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -57,6 +57,8 @@ var (
 
 	maxprocs = flag.Int("maxprocs", 1, "GOMAXPROCS override")
 
+	errorBackoffInitialDuration = 200 * time.Millisecond
```
saikat-royc, line 60: Please make this a flag for errorBackoffMaxDuration and errorBackoffInitialDuration, with default values

judemars: Good idea! Done

saikat-royc: can you also make the errorBackoffMaxDuration configurable through the flag

cmd/gce-pd-csi-driver/main.go
```diff
@@ -42,7 +42,8 @@ var (
 	httpEndpoint         = flag.String("http-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
 	metricsPath          = flag.String("metrics-path", "/metrics", "The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.")
 
-	extraVolumeLabelsStr = flag.String("extra-labels", "", "Extra labels to attach to each PD created. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'. See https://cloud.google.com/compute/docs/labeling-resources for details")
+	errorBackoffInitialDurationMs = flag.Int("backoff-initial-duration-ms", 200, "The amount of ms for the initial duration of the backoff condition. Default is 200.")
```
saikat-royc, line 45: backoff condition for controller publish/unpublish CSI operations

judemars: Added, thanks

cmd/gce-pd-csi-driver/main.go
judemars: Woops sorry I missed that, added errorBackoffMaxDuration. Note I did it in ms to match the initial duration flag, but lmk if you think I should change to minutes/seconds.

cmd/gce-pd-csi-driver/main.go
```diff
@@ -122,7 +124,9 @@ func handle() {
 		if err != nil {
 			klog.Fatalf("Failed to get cloud provider: %v", err)
 		}
-		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider)
+		initialBackoffDuration := time.Duration(*errorBackoffInitialDurationMs) * time.Millisecond
+		maxBackoffDuration := time.Duration(*errorBackoffMaxDurationMs) * time.Microsecond
```
scbrubaker02, line 128: @judemars, looks like this should be time.Millisecond.

judemars: Thanks! Sorry about that! Assigned https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1098 to you.

PR #1084, opened 2022-11-18
Upgrade klog v1 to v2 and fix error wrapping

sunnylovestiramisu:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change

 /kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**:
Simiar to [gcp-filestore-csi-driver/pr/381](https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/pull/381). In V1 klog.Errorf does not support %w but we need to unwrap the error correctly.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
NONE

```release-note

```

PR #1092, opened 2022-11-22
Separate user errors from internal errors

sunnylovestiramisu:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Exclude user errors for googleapis from codes.internal errors for better error categorization
Conversation from https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1084

Fixes #

You are reviewing three different things:
1. fmt.Errorf should use %w for errors
2. logging should use %v for err, or err.Error() with err nil check
3. simplifying status.Error to status.Errorf to avoid the Sprintf call

NONE
```release-note

```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -158,12 +158,12 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	capBytes, err := getRequestCapacity(capacityRange)
 	if err != nil {
-		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume Request Capacity is invalid: %v", err.Error()))
+		return nil, status.Errorf(codes.InvalidArgument, "CreateVolume Request Capacity is invalid: %v", err.Error())
 	}
 
 	err = validateVolumeCapabilities(volumeCapabilities)
 	if err != nil {
-		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("VolumeCapabilities is invalid: %v", err.Error()))
+		return nil, status.Errorf(codes.InvalidArgument, "VolumeCapabilities is invalid: %v", err.Error())
 	}
 
 	// Apply Parameters (case-insensitive). We leave validation of
@@ -185,25 +185,25 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 	case replicationTypeNone:
 		zones, err = pickZones(ctx, gceCS, req.GetAccessibilityRequirements(), 1)
 		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume failed to pick zones for disk: %v", err.Error()))
+			return nil, status.Errorf(codes.InvalidArgument, "CreateVolume failed to pick zones for disk: %v", err.Error())
 		}
 		if len(zones) != 1 {
-			return nil, status.Errorf(codes.Internal, fmt.Sprintf("Failed to pick exactly 1 zone for zonal disk, got %v instead", len(zones)))
+			return nil, status.Errorf(codes.Internal, "Failed to pick exactly 1 zone for zonal disk, got %v instead", len(zones))
 		}
 		volKey = meta.ZonalKey(name, zones[0])
 
 	case replicationTypeRegionalPD:
 		zones, err = pickZones(ctx, gceCS, req.GetAccessibilityRequirements(), 2)
 		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume failed to pick zones for disk: %v", err.Error()))
+			return nil, status.Errorf(codes.InvalidArgument, "CreateVolume failed to pick zones for disk: %v", err.Error())
 		}
 		region, err := common.GetRegionFromZones(zones)
 		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume failed to get region from zones: %v", err.Error()))
+			return nil, status.Errorf(codes.InvalidArgument, "CreateVolume failed to get region from zones: %v", err.Error())
 		}
 		volKey = meta.RegionalKey(name, region)
 	default:
-		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", params.ReplicationType))
+		return nil, status.Errorf(codes.InvalidArgument, "CreateVolume replication type '%s' is not supported", params.ReplicationType)
 	}
 
 	volumeID, err := common.KeyToVolumeID(volKey, gceCS.CloudProvider.GetDefaultProject())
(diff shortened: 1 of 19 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -212,7 +212,11 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	volumeID, err := common.KeyToVolumeID(volKey, gceCS.CloudProvider.GetDefaultProject())
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to convert volume key to volume ID: %v", err)
+		klog.Errorf("Failed to convert volume key to volume ID: %w", err)
```
mattcary, line 215: Ah, so much boilerplate for every error. It will be hard to be consistent about all of this in the future. Instead of IsUserError could we have something like CodeForError so the whole thing looks like the following?

if err != nil {
  return nil, status.Errorf(CodeForError(err), "Failed to convert volume key to volume ID: %v", err)
}

That does mean that we don't get the klog---but we weren't getting that before anyway. So maybe we could have
if err != nil {
  return nil, LoggedError("Failed to convert volume key to volume ID: %w", err)
}

and then LoggedError will klog as well as produce the status.Errorf.

mattcary: golang error handling is crap as far as I can tell---in addition to all the boilerplate, go's aversion to stack traces mean you never really know where an error originally came from.

In particular I like doing a klog in addition to an error return, because the klog gives you generally better information about where the error came from---but the error return is necessary as well because that will get plumbed into an RPC reply which will give clues in the sidecars or kubelet logs.

Maybe if LoggedError(msg, err) looks like { klog(msg, err); return status.Errorf(code, err)}, and we use %v in msg, then it will be enough? We don't care about the %v conversion in the klog, as we only human-read that, it's the status.Errorf call that's important?

msau42: In general, if we are passing the error as a return value, I prefer to leave it up to the caller to log and only log if we are going to swallow an error and not pass it up. That helps avoid duplicate logging of errors at multiple levels.

mattcary: Generally I've had more problems with seeing an error up high in the stack, and not being able to trace back to where it was emitted, than any ambiguities around duplicated messages in the logs. That may be the style in go but if it's possible to push back against it, I'd like to....

judemars:

judemars:
Woops:

/test pull-gcp-compute-persistent-disk-csi-driver-sanity

judemars:
Failed again. Will try again:

/test pull-gcp-compute-persistent-disk-csi-driver-sanity

judemars:
https://github.com/kubernetes/test-infra/pull/28158 should disable the sanity PR gate, sorry for the hassle!

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -333,23 +334,23 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
+			return nil, LoggedError("CreateVolume failed to create regional disk "+name+": ", err)
 		}
 	default:
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("CreateVolume replication type '%s' is not supported", params.ReplicationType))
 	}
 
 	ready, err := isDiskReady(disk)
 	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("CreateVolume disk %v had error checking ready status: %v", volKey, err))
+		return nil, status.Error(codes.Internal, fmt.Sprintf("CreateVolume disk %v had error checking ready status: %v", volKey, err.Error()))
```
mattcary, line 353: Should we LoggedError here too?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -484,20 +487,20 @@ func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Con
 	_, err = gceCS.CloudProvider.GetDisk(ctx, project, volKey, gce.GCEAPIVersionV1)
 	if err != nil {
 		if gce.IsGCENotFoundError(err) {
-			return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find disk %v: %v", volKey.String(), err))
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("Could not find disk %v: %v", volKey.String(), err.Error()))
```
mattcary, line 490: Here and below, status.Errorf can be used to avoid the Sprintf call.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -590,9 +593,9 @@ func (gceCS *GCEControllerServer) executeControllerUnpublishVolume(ctx context.C
 	project, volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, project, volKey)
 	if err != nil {
 		if gce.IsGCENotFoundError(err) {
-			return nil, status.Errorf(codes.NotFound, "ControllerUnpublishVolume could not find volume with ID %v: %v", volumeID, err)
+			return nil, status.Errorf(codes.NotFound, "ControllerUnpublishVolume could not find volume with ID %v: %v", volumeID, err.Error())
 		}
-		return nil, status.Errorf(codes.Internal, "ControllerUnpublishVolume error repairing underspecified volume key: %v", err)
+		return nil, LoggedError("ControllerUnpublishVolume error repairing underspecified volume key: ", err)
```
mattcary, line 598: Why are we logging only here, and not when the volume is not found? Probably we should log everything (?)

pkg/gce-pd-csi-driver/controller.go
sunnylovestiramisu: isDiskReady does not throw user error we defined in this pr:

```
func isDiskReady(disk *gce.CloudDisk) (bool, error) {
	status := disk.GetStatus()
	switch status {
	case "READY":
		return true, nil
	case "FAILED":
		return false, fmt.Errorf("Disk %s status is FAILED", disk.GetName())
	case "CREATING":
		klog.V(4).Infof("Disk %s status is CREATING", disk.GetName())
		return false, nil
	case "DELETING":
		klog.V(4).Infof("Disk %s status is DELETING", disk.GetName())
		return false, nil
	case "RESTORING":
		klog.V(4).Infof("Disk %s status is RESTORING", disk.GetName())
		return false, nil
	default:
		klog.V(4).Infof("Disk %s status is: %s", disk.GetName(), status)
	}

	return false, nil
}
``` 

Do we still want to use the LoggedError?

pkg/gce-pd-csi-driver/controller.go
sunnylovestiramisu: We can add logging for notFound. Initially this pr is just fix unwrap error incorrectly. But at this point it is a blend of adding logs + fixing error unwrap.

pkg/gce-pd-csi-driver/controller.go
mattcary: ah, then you're right it's not necessary.

It would be nice to have a more uniform location to log errors so its clear what is available for debugging. But that's out of scope here.

pkg/gce-pd-csi-driver/controller.go
mattcary: ok, fair enough.

pkg/gce-pd-csi-driver/controller.go
mattcary: thx for fixing

pkg/gce-pd-csi-driver/controller.go
sunnylovestiramisu: Okay i added the change for fixing status.Error to status.Errorf as well in this pr but committed it in a second commit. It looks much simpler :D

PR #1098, opened 2022-12-05
Fix micro > milli max backoff bug

judemars:
**What type of PR is this?**
/kind bug

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1087 added a flag for max backoff threshold in ms but it converts it to microseconds rather than milliseconds. This fixes that.

Fixes #

```release-note
NONE
```


judemars:
/assign scbrubaker02

Thanks for catching this!

PR #1109, opened 2023-01-11
Succeed unpublish when a zonal, underspecified PD is not found

### LINKED ISSUE #1108: Unpublish retries indefinitely for nonexistent PDs.
When the customer manually deletes a PD, gce.RepairUnderspecifiedVolumeKey() will return a NotFound error, which causes the unpublish step to fail indefinitely.

I explored several options to nudge the state machine:
1. Succeed after a max number of retry attempts.
2. Succeed on the first NotFound error, as it doesn't appear to be possible to reach the unpublish step with a non-existent PD.

artemvmin:
/kind bug

This PR specifically targets non-existent zonal disks with underspecified volume IDs. The case of missing nodes and missing PDs with fully specified volume IDs already succeeds the unpublish step.

- Add test case for fully specified, not found PDs during unpublish.
- Clean up the testing structure by introducing a backoffTestDriver generator, which can be reused by both publish and unpublish tests.
- Remove redundant comments from the test code.

Fixes #1108 

**Release note:**
```release-note
NONE
```

### DIFF
--- a/test/e2e/tests/setup_e2e_test.go
+++ b/test/e2e/tests/setup_e2e_test.go
@@ -56,7 +56,6 @@ func init() {
 }
 
 func TestE2E(t *testing.T) {
-
 	flag.Parse()
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Google Compute Engine Persistent Disk Container Storage Interface Driver Tests")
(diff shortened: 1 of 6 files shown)


saikat-royc:
I think for clarity we should mention in the PR description that PR fixes non existent zonal disk in the csi migration RepairUnderspecifiedVolumeKey. For other type of disks that does not take repair underspecified key,  https://github.com/artemvmin/gcp-compute-persistent-disk-csi-driver/blob/a242102a6626cf040e24f2f13084c505a30d9414/pkg/gce-pd-csi-driver/controller.go#L629
diskIsAttached() is expected to return no disk for a non-existent disk.

saikat-royc:
We should also add a test case to the [e2e test](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/e2e/tests/single_zone_e2e_test.go) for non-existent underspecified volume id.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -287,6 +287,36 @@ var _ = Describe("GCE PD CSI Driver", func() {
 
 	})
 
+	It("Should complete entire disk lifecycle with underspecified volume ID and missing volume", func() {
```
sunnylovestiramisu, line 290: Nit: should we also add this test case for pd-extreme?

test/e2e/tests/setup_e2e_test.go
```diff
@@ -147,6 +116,40 @@ var _ = AfterSuite(func() {
 	}
 })
 
+func NewTestContext(zone string) *remote.TestContext {
```
artemvmin, line 119: Reviewer note: I was experimenting with spinning up a new instance for my test case without it being added to the global pool of test contexts. I didn't end up using it, but the refactor allows future test cases to perform a similar setup.

artemvmin:
Tested locally. This should be passing. The test-infra clone appears to be causing problems...

test/e2e/tests/single_zone_e2e_test.go
artemvmin: Done.

artemvmin:
Yep, seeing the same error across other test jobs on https://prow.k8s.io/.

artemvmin:
Now seeing Jenkins errors:

`TypeError: 'zip' object is not subscriptable`
`TypeError: a bytes-like object is required, not 'str'`

test/e2e/tests/single_zone_pd_extreme_e2e_test.go
```diff
@@ -272,4 +272,64 @@ var _ = Describe("GCE PD CSI Driver pd-extreme", func() {
 			Expect(gce.IsGCEError(err, "notFound")).To(BeTrue(), "Expected disk to not be found")
 		}()
 	})
+
+	It("Should complete entire disk lifecycle with underspecified volume ID", func() {
```
msau42, line 276: @sunnylovestiramisu How many test cases can be shared between extreme pd and other pd types where the only difference is the type of disk created?

It may be worth considering refactoring the test cases such that you can share the same test logic and just swap out the disk type.

test/e2e/tests/single_zone_pd_extreme_e2e_test.go
```diff
@@ -272,4 +272,64 @@ var _ = Describe("GCE PD CSI Driver pd-extreme", func() {
...
+		testContext := getRandomTestContext()
+
+		p, z, _ := testContext.Instance.GetIdentity()
+		client := testContext.Client
+		instance := testContext.Instance
+
+		volName, _ := createAndValidateUniqueZonalDisk(client, p, z, standardDiskType)
```
msau42, line 283: should this be the extreme pd disk type?

artemvmin: Yes. Done.

test/e2e/tests/single_zone_pd_extreme_e2e_test.go
artemvmin: Happy to do this. It started as a small subset, but seeing as this test case is compatible with pd_extreme, there may be others.

We should put this back into a single file and use [table-driven tests](https://pkg.go.dev/github.com/onsi/ginkgo@v1.16.5/extensions/table).

artemvmin: I filed #1117 to take care of this. I had issues with the ginkgo dependency, so I will do it in a separate PR. I'm considering upgrading to ginkgo 2.X

test/e2e/tests/single_zone_pd_extreme_e2e_test.go
artemvmin: 🙉 Sorry, I'm coming from hg/fig, where amend automatically adds modified files.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1122,11 +1155,11 @@ func equalWithinEpsilon(a, b, epsiolon int64) bool {
 	return b-a < epsiolon
 }
 
-func createAndValidateUniqueZonalDisk(client *remote.CsiClient, project, zone string, diskType string) (volName, volID string) {
```
artemvmin, line 1125: **Reviewer note:** While the [Google golang style guide](https://github.com/google/styleguide/blob/gh-pages/go/decisions.md#named-result-parameters) doesn't prohibit named return values, this structure does cause subtle bugs when mixed with short assignment and deferred functions. For example. if `err` wasn't explicitly declared and line 1129 used short assignment, the returned string would always be empty. I find it best to avoid named result parameters and document return types with a function comment when necessary.

artemvmin:
I ran this locally and all tests passed. These timeouts look unrelated.

/retest

PR #1128, opened 2023-02-08
Change iops params to not handle GB/MB etc.

sunnylovestiramisu:
**What type of PR is this?**
/kind feature

After discussion, the iops fields should not be handling GB/MB etc. because PD is treating this value as a configuration for the whole disk

Fixes #

The change of HdX and HdT is not released yet.
```release-note
NONE
```


pkg/common/utils.go
```diff
@@ -251,15 +255,39 @@ func ValidateSnapshotType(snapshotType string) error {
...
+	if err != nil {
+		// the string ends with m, M or k, K
+		last := string(str[len(str)-1])
+
+		switch last {
+		case "k":
+			multiplier = thousand
+		case "K":
+			multiplier = thousand
+		case "m":
```
msau42, line 277: "m" means milli, not mega

pkg/common/utils.go
```diff
@@ -251,15 +255,39 @@ func ValidateSnapshotType(snapshotType string) error {
 	}
 }
 
-// ConvertGiBStringToInt64 converts a GiB string to int64
-func ConvertGiBStringToInt64(str string) (int64, error) {
+// ConvertStringToInt64 converts a string to int64
+func ConvertStringToInt64(str string) (int64, error) {
```
msau42, line 259: It may be better to use kubernetes resource quantity parsing instead of rolling your own: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity

sunnylovestiramisu: But they do not have our use case. They do not convert these values?

pkg/common/utils.go
sunnylovestiramisu: Here we do million and thousand?

pkg/common/utils.go
```diff
@@ -251,15 +255,39 @@ func ValidateSnapshotType(snapshotType string) error {
...
-	return volumehelpers.RoundUpToGiB(quantity)
+
+	multiplier := int64(1)
+	value, err := strconv.ParseInt(str, 10, 64)
+	if err != nil {
+		// the string ends with m, M or k, K
+		last := string(str[len(str)-1])
+
+		switch last {
+		case "k":
```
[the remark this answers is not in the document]
sunnylovestiramisu: Why do we care about base 2?

pkg/common/utils.go
msau42: In the SI system, the exact letters and capitalization is important.
M = 10^6
m = 1/1000

pkg/common/utils.go
msau42: I'm not sure I understand. It looks like this function is converting from a string, possibly abbreviated in SI notation, to the raw number right?

pkg/common/utils.go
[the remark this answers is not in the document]
msau42: In SI notation, k = 1000 and Ki = 2^10

https://physics.nist.gov/cuu/Units/binary.html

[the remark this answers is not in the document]
sunnylovestiramisu: Oh I was asking why do we need support for base 2 input, because it is unnatural for humans. Do our customers ask for that?

pkg/common/utils.go
sunnylovestiramisu: Yes and we want to support:
```
"1000" -> int64(1000)
"1K"-> int64(1000)
"1M" -> int64(1,000,000)
Maybe "1G" -> int64(1,000,000,000)
```

sunnylovestiramisu: The "K" is not supported in [resource#Quantity](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity). But other than that mostly covered. We do not need to support the "K"?

pkg/common/utils_test.go
```diff
@@ -588,73 +588,91 @@ func TestConvertGiBStringToInt64(t *testing.T) {
...
+			"invalid empty string",
+			"",
+			10000,
+			true,
 		},
 		{
-			"round G to GiB",
-			"1000G",
-			932,
+			"invalid string",
```
leiyiz, line 619: do you think it's worth it to add a case for `1KiB` or `1GB` as invalid string?

test/k8s-integration/config/sc-extreme.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 provisioner: pd.csi.storage.gke.io
 parameters:
   type: pd-extreme
-  provisioned-iops-on-create: '10000Gi'
+  provisioned-iops-on-create: '10000'
   # Add labels for testing.
   labels: key1=value1,key2=value2
 volumeBindingMode: WaitForFirstConsumer
```
leiyiz, line 11: nit: new line at EOF

pkg/common/utils_test.go
sunnylovestiramisu: We can add that.

PR #1131, opened 2023-02-14
Add more unit tests for hyperdisk feature

sunnylovestiramisu:
**What type of PR is this?**

/kind cleanup

Add more tests to clarify use case, clean up test format.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/common/utils_test.go
+++ b/pkg/common/utils_test.go
@@ -612,13 +612,13 @@ func TestConvertStringToInt64(t *testing.T) {
 		{
 			desc:        "invalid empty string",
 			inputStr:    "",
-			expInt64:    10000,
+			expInt64:    0,
 			expectError: true,
 		},
 		{
 			desc:        "invalid string",
 			inputStr:    "ew%65",
-			expInt64:    10000,
+			expInt64:    0,
 			expectError: true,
 		},
 		{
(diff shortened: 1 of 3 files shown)


pkg/common/utils_test.go
```diff
@@ -706,70 +706,82 @@ func TestConvertMiBStringToInt64(t *testing.T) {
 		expectError bool
 	}{
 		{
-			"valid number string",
-			"10000",
-			1,
-			false,
+			desc:        "valid number string",
+			inputStr:    "10000",
```
leiyiz, line 710: why is 10000 converted to 1 and 1000Ki is also 1?

leiyiz:
also consider change the name from `ConvertMiBStringToInt64(str string)` to `ConvertMiStringToInt64(str string)`?

pkg/common/utils_test.go
sunnylovestiramisu: The round up is doing roundedUp := volumeSizeBytes / allocationUnitBytes(Here we use MiB which is int 1048576)

10000 / 1048576 round up is 1
1000Ki / 1048576 round up is 1
1000k / 1048576 round up is 1 as well.

sunnylovestiramisu:
> also consider change the name from `ConvertMiBStringToInt64(str string)` to `ConvertMiStringToInt64(str string)`?

That is a good suggestion, will change to that now

leiyiz (commented):
/lgtm
/approve

PR #1180, opened 2023-04-04
Update go version to 1.20

sunnylovestiramisu:
**What type of PR is this?**

/kind cleanup

Update go version to match k/k 1.27: https://github.com/kubernetes/kubernetes/blob/master/.go-version

Fixes #

```release-note
Update go version to 1.20.2 for k/k 1.27
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 
 ARG BUILDPLATFORM
 
-FROM --platform=$BUILDPLATFORM golang:1.19.6 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20.3 as builder
 
 ARG STAGINGVERSION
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,7 +14,7 @@
 
 ARG BUILDPLATFORM
 
-FROM --platform=$BUILDPLATFORM golang:1.19.6 as builder
+FROM --platform=$BUILDPLATFORM golang:1.20.2 as builder
```
msau42, line 17: It looks like 1.20.3 is being released today with a number of CVE fixes:
https://groups.google.com/g/golang-announce/c/71Wg3N0IZk0/m/xchF2bEZAgAJ

sunnylovestiramisu: Updated.

PR #1182, opened 2023-04-10
prune changelog for 1.10 release

amacaskill:
/kind documentation

None
/assign @sunnylovestiramisu


CHANGELOG.md
```diff
@@ -1 +1 @@
-See [CHANGELOG/README.md](./CHANGELOG/README.md).
\ No newline at end of file
+See the [1.10 changelog in the master branch](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/CHANGELOG/CHANGELOG-1.10.md).
```
sunnylovestiramisu, line 1: This should be 1.0.md

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/CHANGELOG/CHANGELOG-1.0.md

CHANGELOG.md
```diff
@@ -1 +1 @@
-See [CHANGELOG/README.md](./CHANGELOG/README.md).
\ No newline at end of file
+See the [1.10 changelog in the master branch](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/CHANGELOG/CHANGELOG-1.0.md).
```
msau42, line 1: I'm confused, the text says 1.10 changelog but the link is pointing to the 1.0 changelog (that is being removed?)

msau42 (commented):
~~What is the intent of this change? Do we want the master branch changelog to reflect only the latest release changes, and that prior release changelogs are only kept in the release branches?

If so, have we confirmed that the changelogs in release branches are up to date? I recall seeing various changelog PRs that only updated them in the master branch.~~

EDIT: Nevermind I noticed this is against 1.10 branch.

CHANGELOG.md
amacaskill: Oops sorry I had 1.10 both places, then Sunny said to change to 1.0, but didn't change the text... Doing that now.

msau42: Shouldn't it be the 1.10 changelog?

amacaskill: That's what I had but I was told me to [change to 1.0](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1182#discussion_r1162130707) so I changed it back.

sunnylovestiramisu: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/CHANGELOG/CHANGELOG-1.10.md does not exist. I think we put everything in the 1.0 change log.

sunnylovestiramisu: It should be point to this one? https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/CHANGELOG/CHANGELOG-1.1.md

amacaskill: I think the ordering is 1.9 -> 1.10 -1.11. We have never gone above 1.9 so this is a little confusing. But this is what kubernetes does: https://screenshot.googleplex.com/7p6BCrJbfhdU3Bn

I will create a changelog for 1.10 in master in a separate PR

sunnylovestiramisu: Sorry I was thinking that there is no major change in pd-csi so I did not realize we are bumping up the minor version.

msau42: Are there any new features we're adding in 1.10?

sunnylovestiramisu: No new features but there are some fixes, mkfs and xfs_io error fixes etc, but these are/should be cherry-picked to release-1.9

amacaskill: The [volume cloning fix](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1150) adds some new functionality around not having to specify allowedTopologies to get the cloning to work. Also [we are updating go version to match k/k 1.27](https://github.com/kubernetes/kubernetes/blob/master/.go-version), but maybe that is not considered a new feature.

CHANGELOG.md
amacaskill, line 1: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/CHANGELOG/CHANGELOG-1.10.md doesn't exist yet. I will create this in master, but this PR doesn't need to wait for this I don't think.

PR #1215, opened 2023-05-02
Add tests for ROX mode during Node Stage 

### LINKED ISSUE #1088: read only volumes should not be expanded and we need OSS ROX testing
#972 added unconditional filesystem resize and it should not happen for read only volumes. We are not seeing failures in mounting as resizing RO volume only produce an warning but this could have broken ROX functionality.

1. we should not try to expand RO volumes
2. we should write an OSS ROX test ASAP

judemars:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This is a follow up to https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1088 which does 2 things:
1. Adds more fine grained testing for whether or not resize is attempted under certain scenarios during node stage.
2. Skips resize attempt if volume capability from request indicates readonly

Fixes #1088

```release-note
Skips resize attempt if volume capability from request indicates readonly.
```

### DIFF
--- a/pkg/deviceutils/fake-device-utils.go
+++ b/pkg/deviceutils/fake-device-utils.go
@@ -17,12 +17,16 @@ package deviceutils
 import "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/resizefs"
 
 type fakeDeviceUtils struct {
+	skipResize bool
 }
 
-var _ DeviceUtils = &fakeDeviceUtils{}
+var _ DeviceUtils = &fakeDeviceUtils{
+}
 
-func NewFakeDeviceUtils() *fakeDeviceUtils {
-	return &fakeDeviceUtils{}
+func NewFakeDeviceUtils(skipResize bool) *fakeDeviceUtils {
+	return &fakeDeviceUtils{
+		skipResize: skipResize,
+	}
 }
 
 // Returns list of all /dev/disk/by-id/* paths for given PD.
@@ -41,6 +45,9 @@ func (_ *fakeDeviceUtils) DisableDevice(devicePath string) error {
 	return nil
 }
 
-func (_ *fakeDeviceUtils) Resize(resizer resizefs.Resizefs, devicePath string, deviceMountPath string) (bool, error) {
-	return false, nil
+func (du *fakeDeviceUtils) Resize(resizer resizefs.Resizefs, devicePath string, deviceMountPath string) (bool, error) {
+	if du.skipResize {
+		return false, nil
+	}
+	return resizer.Resize(devicePath, deviceMountPath)
 }
--- a/pkg/deviceutils/fake-device-utils.go
+++ b/pkg/deviceutils/fake-device-utils.go
@@ -20,8 +20,7 @@ type fakeDeviceUtils struct {
 	skipResize bool
 }
 
-var _ DeviceUtils = &fakeDeviceUtils{
-}
+var _ DeviceUtils = &fakeDeviceUtils{}
 
 func NewFakeDeviceUtils(skipResize bool) *fakeDeviceUtils {
 	return &fakeDeviceUtils{
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -342,10 +342,12 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
 
 	// Part 4: Resize filesystem.
 	// https://github.com/kubernetes/kubernetes/issues/94929
-	resizer := resizefs.NewResizeFs(ns.Mounter)
-	_, err = ns.DeviceUtils.Resize(resizer, devicePath, stagingTargetPath)
-	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("error when resizing volume %s from device '%s' at path '%s': %v", volumeID, devicePath, stagingTargetPath, err.Error()))
+	if !readonly {
+		resizer := resizefs.NewResizeFs(ns.Mounter)
+		_, err = ns.DeviceUtils.Resize(resizer, devicePath, stagingTargetPath)
+		if err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("error when resizing volume %s from device '%s' at path '%s': %v", volumeID, devicePath, stagingTargetPath, err.Error()))
+		}
 	}
 
 	klog.V(4).Infof("NodeStageVolume succeeded on %v to %s", volumeID, stagingTargetPath)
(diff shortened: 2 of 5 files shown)


leiyiz (commented):
/approve

pkg/deviceutils/fake-device-utils.go
```diff
@@ -17,12 +17,16 @@ package deviceutils
 import "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/resizefs"
 
 type fakeDeviceUtils struct {
+	skipResize bool
 }
 
-var _ DeviceUtils = &fakeDeviceUtils{}
+var _ DeviceUtils = &fakeDeviceUtils{
```
RomanBednar: Unnecessary newline, later fixed in separate commit with update-gofmt.sh so this change should not be here in this commit.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -342,10 +342,12 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
 
 	// Part 4: Resize filesystem.
 	// https://github.com/kubernetes/kubernetes/issues/94929
-	resizer := resizefs.NewResizeFs(ns.Mounter)
-	_, err = ns.DeviceUtils.Resize(resizer, devicePath, stagingTargetPath)
-	if err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("error when resizing volume %s from device '%s' at path '%s': %v", volumeID, devicePath, stagingTargetPath, err.Error()))
+	if !readonly {
```
RomanBednar, line 345: This commit is named "Add tests..." and is changing non-test code. This should be in a separate commit.

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -381,26 +382,49 @@ func TestNodeStageVolume(t *testing.T) {
...
+		req           *csi.NodeStageVolumeRequest
+		deviceSize    int
+		blockExtSize       int
+		readonlyBit   string
+		expResize     bool
+		expErrCode    codes.Code
 	}{
 		{
-			name: "Valid request",
+			name: "Valid request, no resize bc size",
```
RomanBednar, line 394: nit: I'd avoid shortening test names. Same for the remaining tests.

```suggestion
                        name: "Valid request, no resize because block and filesystem sizes match",
```

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -517,6 +586,9 @@ func TestNodeStageVolume(t *testing.T) {
 		if tc.expErrCode != codes.OK {
 			t.Fatalf("Expected error: %v, got no error", tc.expErrCode)
 		}
+		if tc.expResize != resizeCalled {
+			t.Fatalf("Test expected resize to be called: %t but was it called: %t", tc.expResize, resizeCalled)
+		}
```
RomanBednar, line 603: Could be more readable with:

```suggestion
		if tc.expResize == true && resizeCalled == false {
			t.Fatalf("Test did not call resize, but it was expected.")
		}
		if tc.expResize == false && resizeCalled == true {
			t.Fatalf("Test called resize, but it was not expected.")
		}
```

However, can we even get a condition where `tc.expResize=false` and `resizeCalled=true`? I think not because `tc.expResize=false` will skip adding command actions for the test and the moment we hit resize code this will panic with `panic: ran out of Command() actions.`

RomanBednar:
Not related to your PR but recent change to `DeviceUtils` interface that added `DisableDevice` method is missing implementation of that method for [this](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/commit/4d7eb2a50c71d53aee9027c2c78ab41883fd9f46#diff-a9e833f45b96af7d94455dcbb34fb9875324c2c3b1e5246133658462888ead67R94) struct. So running unit tests on different platform than linux or windows fails:

```
# sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/deviceutils
pkg/deviceutils/device-utils.go:96:21: cannot use &deviceUtils{} (value of type *deviceUtils) as type DeviceUtils in variable declaration:
        *deviceUtils does not implement DeviceUtils (missing DisableDevice method)
```

I think there should be one more file (e.g. `device-utils_unsupported.go`) with "empty" interface implementation. Example: https://github.com/kubernetes-sigs/ibm-vpc-block-csi-driver/blob/564aa688c069300ef4350fa9d7ec5c841a32a257/vendor/github.com/IBM/ibm-csi-common/pkg/mountmanager/mount_unsupported.go#L52

Even when this is added same issue occurs with `utils_*.go`:

```
$ go test -v -run TestNodeStageVolume
# sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-pd-csi-driver [sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-pd-csi-driver.test]
./node.go:144:13: undefined: preparePublishPath
./node.go:156:21: undefined: getDevicePath
./node.go:282:21: undefined: getDevicePath
./node.go:296:12: undefined: prepareStagePath
./node.go:321:8: undefined: formatAndMount
./node.go:332:10: undefined: formatAndMount
./node.go:432:16: undefined: getBlockSizeBytes
./node.go:489:21: undefined: getDevicePath
./node.go:518:22: undefined: resizefs.NewResizeFs
./node.go:525:24: undefined: getBlockSizeBytes
.
.
.
```

Surely adding empty interfaces does not solve much but at least there are meaningful messages like "unsupported platform" instead of compile errors.

pkg/deviceutils/fake-device-utils.go
judemars: This will merge w/ squash so this will go away.

pkg/gce-pd-csi-driver/node.go
judemars: Again, this will merge with squash so individual commits will go away. Are you suggesting moving this into a separate PR?

pkg/gce-pd-csi-driver/node_test.go
judemars: I don't think it's wise to have our test cases tailored to the interworkings of the command executor. I'm fine to expand this into the two cases though as you describe.

pkg/gce-pd-csi-driver/node.go
RomanBednar: Sure, if we're squashing then commit separation does not affect the history - so it does not help here. But where it helps a lot are reviews. If commits are not well organized (and described) then reviewers have to look at the PR diff as a whole and it can get difficult to read it. It's not a hard requirement for sure, just a suggestion.

judemars: Gotcha. For a commit this small I'd argue it's more hassle than benefit to go back and clean up commits when they will be squashed, but I'll definitely keep that in mind in the future as I didn't realize folks look at individual commits during review. I've only ever looked at the entire diff. Thanks for bringing it up though so I know in the future!

PR #1227, opened 2023-05-15
Add metrics for CSI server side error count

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

It adds an CSI server side error count metric for PDCSI and Hyperdisk volumes
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -195,6 +195,12 @@ func useVolumeCloning(req *csi.CreateVolumeRequest) bool {
 
 func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	var err error
+	diskTypeForMetric := ""
+	defer func() {
+		if err != nil {
+			gceCS.Metrics.RecordOperationErrorMetrics("CreateVolume", err, diskTypeForMetric)
+		}
+	}()
 	// Validate arguments
 	volumeCapabilities := req.GetVolumeCapabilities()
 	name := req.GetName()
@@ -276,6 +282,7 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	// Validate if disk already exists
 	existingDisk, err := gceCS.CloudProvider.GetDisk(ctx, gceCS.CloudProvider.GetDefaultProject(), volKey, gceAPIVersion)
+	diskTypeForMetric = metrics.GetDiskType(existingDisk)
 	if err != nil {
 		if !gce.IsGCEError(err, "notFound") {
 			return nil, common.LoggedError("CreateVolume, failed to getDisk when validating: ", err)
@@ -330,6 +337,7 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 			// Verify that the volume in VolumeContentSource exists.
 			diskFromSourceVolume, err := gceCS.CloudProvider.GetDisk(ctx, project, sourceVolKey, gceAPIVersion)
+			diskTypeForMetric = metrics.GetDiskType(diskFromSourceVolume)
 			if err != nil {
 				if gce.IsGCEError(err, "notFound") {
 					return nil, status.Errorf(codes.NotFound, "CreateVolume source volume %s does not exist", volumeContentSourceVolumeID)
@@ -386,6 +394,7 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	// Create the disk
 	var disk *gce.CloudDisk
+	diskTypeForMetric = params.DiskType
 	switch params.ReplicationType {
 	case replicationTypeNone:
 		if len(zones) != 1 {
@@ -393,8 +402,6 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		}
 		disk, err = createSingleZoneDisk(ctx, gceCS.CloudProvider, name, zones, params, capacityRange, capBytes, snapshotID, volumeContentSourceVolumeID, multiWriter)
 		if err != nil {
-			// Emit metric for expected disk type from storage class
-			defer gceCS.Metrics.RecordOperationErrorMetrics("CreateVolume", err, params.DiskType)
 			return nil, common.LoggedError("CreateVolume failed to create single zonal disk "+name+": ", err)
 		}
 	case replicationTypeRegionalPD:
@@ -403,8 +410,6 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		}
 		disk, err = createRegionalDisk(ctx, gceCS.CloudProvider, name, zones, params, capacityRange, capBytes, snapshotID, volumeContentSourceVolumeID, multiWriter)
 		if err != nil {
-			// Emit metric for expected disk type from storage class
-			defer gceCS.Metrics.RecordOperationErrorMetrics("CreateVolume", err, params.DiskType)
 			return nil, common.LoggedError("CreateVolume failed to create regional disk "+name+": ", err)
 		}
 	default:
@@ -413,8 +418,6 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	ready, err := isDiskReady(disk)
 	if err != nil {
-		// Emit metric for expected disk type from storage class as the disk is not ready and might not have PD type populated
-		defer gceCS.Metrics.RecordOperationErrorMetrics("CreateVolume", err, params.DiskType)
 		return nil, status.Errorf(codes.Internal, "CreateVolume disk %v had error checking ready status: %v", volKey, err.Error())
 	}
 	if !ready {
@@ -427,6 +430,13 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 }
 
 func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
+	var err error
+	diskTypeForMetric := ""
+	defer func() {
+		if err != nil {
+			gceCS.Metrics.RecordOperationErrorMetrics("DeleteVolume", err, diskTypeForMetric)
+		}
+	}()
 	// Validate arguments
 	volumeID := req.GetVolumeId()
 	if len(volumeID) == 0 {
@@ -442,6 +452,8 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 	}
 
 	project, volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, project, volKey)
+	disk, _ := gceCS.CloudProvider.GetDisk(ctx, project, volKey, gce.GCEAPIVersionV1)
+	diskTypeForMetric = metrics.GetDiskType(disk)
 	if err != nil {
 		if gce.IsGCENotFoundError(err) {
 			klog.Warningf("DeleteVolume treating volume as deleted because cannot find volume %v: %v", volumeID, err.Error())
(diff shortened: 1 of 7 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -399,6 +403,10 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		}
 		disk, err = createRegionalDisk(ctx, gceCS.CloudProvider, name, zones, params, capacityRange, capBytes, snapshotID, volumeContentSourceVolumeID, multiWriter)
 		if err != nil {
+			// Emit metric for error
+			if strings.Contains(params.DiskType, "pd") {
```
sunnylovestiramisu, line 407: Do we still need this diskType check? Because we are doing it in RecordOperationErrorMetrics

pkg/metrics/metrics.go
```diff
@@ -79,6 +114,38 @@ func (mm *metricsManager) EmitGKEComponentVersion() error {
...
 
+// CodeForError returns a pointer to the grpc error code that maps to the http
+// error code for the passed in user googleapi error. Returns codes.Internal if
+// the given error is not a googleapi error caused by the user. The following
+// http error codes are considered user errors:
+// (1) http 400 Bad Request, returns grpc InvalidArgument,
+// (2) http 403 Forbidden, returns grpc PermissionDenied,
+// (3) http 404 Not Found, returns grpc NotFound
+// (4) http 429 Too Many Requests, returns grpc ResourceExhausted
+func CodeForError(err error) *codes.Code {
```
sunnylovestiramisu, line 125: Why are we copying this code over? We can use the existing CodeForError in the pkg/gce-pd-csi-driver/utils.go?

pkg/metrics/metrics.go
```diff
@@ -70,7 +91,21 @@ func (mm *metricsManager) recordComponentVersionMetric() error {
...
+	operationErr error,
+	diskType string) {
+	var driverName string
+	if strings.Contains(diskType, "hyperdisk") {
+		driverName = hyperdiskDriverName
+	}
+	if strings.Contains(diskType, "pd") {
+		driverName = pdcsiDriverName
+	}
+	pdcsiOperationErrorsMetric.WithLabelValues(driverName, "/csi.v1.Controller/"+operationName, LoggedError(operationErr).Error(), diskType).Set(1.0)
```
sunnylovestiramisu, line 105: We want the error code here not the error message. So CodeForError not LoggedError.

pkg/gce-pd-csi-driver/utils_test.go
```diff
@@ -323,7 +324,7 @@ func TestCodeForError(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Logf("Running test: %v", tc.name)
-		actualCode := *CodeForError(tc.inputErr)
+		actualCode := *common.CodeForError(tc.inputErr)
```
sunnylovestiramisu, line 327: Can we move this unit test separately to another utils_test.go under common?

pkg/gce-pd-csi-driver/controller.go
Sneha-at: Missed this disk type check, it is now removed

pkg/metrics/metrics.go
Sneha-at: The CodeForError in the pkg/gce-pd-csi-driver/utils.go resulted in circular dependency thus moving it to pkg/common/utils.go as discussed

pkg/metrics/metrics.go
Sneha-at: Updated this to capture the error code

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -36,13 +36,15 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/strings/slices"
 
-	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
+	common "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
```
sunnylovestiramisu, line 39: This seems not renaming to a different name, why do we need the common?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -704,10 +714,12 @@ func (gceCS *GCEControllerServer) executeControllerUnpublishVolume(ctx context.C
 		klog.V(4).Infof("ControllerUnpublishVolume succeeded for disk %v from node %v. Already not attached.", volKey, nodeID)
 		return &csi.ControllerUnpublishVolumeResponse{}, nil
 	}
-
+	diskToUnpublish, _ := gceCS.CloudProvider.GetDisk(ctx, gceCS.CloudProvider.GetDefaultProject(), volKey, gce.GCEAPIVersionV1)
```
sunnylovestiramisu, line 717: We should handle the error of this GetDisk call as well, I would say make it not disruptive? and move it to before emit the metrics?

We can do L720:
```
if err != nil {
	diskToUnpublish, getDiskError := gceCS.CloudProvider.GetDisk(ctx, gceCS.CloudProvider.GetDefaultProject(), volKey, gce.GCEAPIVersionV1)

	// If there is an error, log it as a warning only and do not emit the metrics, because diskToUnpublish will be nil

	// Emit metric for error 
	gceCS.Metrics.RecordOperationErrorMetrics("ControllerUnpublishVolume", err, diskToUnpublish.GetPDType())
	return nil, common.LoggedError("unknown detach error: ", err)
}
```

Sneha-at: Updated error handling

pkg/metrics/metrics.go
```diff
@@ -37,28 +41,42 @@ var (
 		Name: "component_version",
 		Help: "Metric to expose the version of the PDCSI GKE component.",
 	}, []string{"component_version"})
+
+	pdcsiOperationErrorsMetric = metrics.NewGaugeVec(
+		&metrics.GaugeOpts{
+			Subsystem:      "csidriver",
+			Name:           "pdcsi_operation_errors",
```
sunnylovestiramisu, line 48: This needs to change to operation_errors as well.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -391,22 +393,34 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		}
 		disk, err = createSingleZoneDisk(ctx, gceCS.CloudProvider, name, zones, params, capacityRange, capBytes, snapshotID, volumeContentSourceVolumeID, multiWriter)
 		if err != nil {
-			return nil, LoggedError("CreateVolume failed to create single zonal disk "+name+": ", err)
+			// Emit metric for expected disk type from storage class
+			if params.DiskType != "" {
```
msau42, line 397: Can we use a defer statement to log the metrics, so that we don't miss any new return branches we could add inthe future?

msau42, line 397: For the places that we don't know the disk type, should we record an error anyway with empty disktype?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -332,7 +334,7 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 				if gce.IsGCEError(err, "notFound") {
 					return nil, status.Errorf(codes.NotFound, "CreateVolume source volume %s does not exist", volumeContentSourceVolumeID)
 				} else {
-					return nil, LoggedError("CreateVolume unknown get disk error when validating: ", err)
+					return nil, common.LoggedError("CreateVolume unknown get disk error when validating: ", err)
```
msau42, line 337: this isn't your change, but what makes an error "unknown"? should we just say "failed to getDisk"?

pkg/metrics/metrics.go
```diff
@@ -20,15 +20,19 @@ import (
...
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 )
 
 const (
 	// envGKEPDCSIVersion is an environment variable set in the PDCSI controller manifest
 	// with the current version of the GKE component.
-	envGKEPDCSIVersion = "GKE_PDCSI_VERSION"
+	envGKEPDCSIVersion  = "GKE_PDCSI_VERSION"
+	hyperdiskDriverName = "hyperdisk.csi.storage.gke.io"
```
msau42, line 34: I don't think we need to fake the driver name since we are distinguishing by disk type

pkg/metrics/metrics.go
```diff
@@ -37,28 +41,42 @@ var (
...
 	return mm.registry
 }
 
-func (mm *metricsManager) registerComponentVersionMetric() {
+func (mm *MetricsManager) registerComponentVersionMetric() {
 	mm.registry.MustRegister(gkeComponentVersion)
 }
 
-func (mm *metricsManager) recordComponentVersionMetric() error {
+func (mm *MetricsManager) RegisterHyperdiskMetric() {
```
msau42, line 75: These metrics are not hyperdisk specific

pkg/metrics/metrics.go
```diff
@@ -37,28 +41,42 @@ var (
 		Name: "component_version",
 		Help: "Metric to expose the version of the PDCSI GKE component.",
 	}, []string{"component_version"})
+
+	pdcsiOperationErrorsMetric = metrics.NewGaugeVec(
```
msau42, line 45: I don't think Gauge is what we want for errors. Counter fits better.

sunnylovestiramisu: What counter metrics are you referring to? Can you point the reference? It is difficult to go from the code editor to find what you meant without context. 

I have seen NewHistogramVec and NewGaugeVec so far.

pkg/metrics/metrics.go
sunnylovestiramisu: But if in the long run we want to make some changes in the sidecar to distinguish two different driver_name, this step seems a bit more proactive?

pkg/gce-pd-csi-driver/controller.go
Sneha-at: So something like this?
```
	if params.DiskType != "" {
		defer gceCS.Metrics.RecordOperationErrorMetrics("CreateVolume", err, params.DiskType)
	}
```
Could you please share an example of "so that we don't miss any new return branches we could add" for me to understand better.

Sneha-at: Recording metrics for unknown disk types

pkg/metrics/metrics.go
Sneha-at: Renamed tp RegisterPDCSIMetric

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -391,22 +393,28 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 		}
 		disk, err = createSingleZoneDisk(ctx, gceCS.CloudProvider, name, zones, params, capacityRange, capBytes, snapshotID, volumeContentSourceVolumeID, multiWriter)
 		if err != nil {
-			return nil, LoggedError("CreateVolume failed to create single zonal disk "+name+": ", err)
+			// Emit metric for expected disk type from storage class
+			defer gceCS.Metrics.RecordOperationErrorMetrics("CreateVolume", err, params.DiskType)
```
msau42, line 397: What I mean with using defer, is to add the defer statement at the top of the function. That way every return branch will emit a metric, and if we add more return branches in the future, we won't forget to emit a metric for it. So something like

```
func CreateVolume() {
  // declare some variables used by the metrics recording
  err error
  diskType string

  defer func() {
    if err != nil {
      recordErrorMetric("CreateVolume", err, diskType)
    }
  }{}

  // rest of the function
}
```

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -328,11 +337,12 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 			// Verify that the volume in VolumeContentSource exists.
 			diskFromSourceVolume, err := gceCS.CloudProvider.GetDisk(ctx, project, sourceVolKey, gceAPIVersion)
+			diskTypeForMetric = metrics.GetDiskType(diskFromSourceVolume)
```
msau42, line 340: I don't think we should override the diskType in this case. This is the cloning case and refers to the disk we want to clone from, but not necessarily the target diskType

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -384,22 +394,23 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	// Create the disk
 	var disk *gce.CloudDisk
+	diskTypeForMetric = params.DiskType
```
msau42, line 397: Can we set this much earlier, right when we have the information?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -449,20 +469,28 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
...
+		}
+	}()
 	// Only valid requests will be accepted
-	_, _, err := gceCS.validateControllerPublishVolumeRequest(ctx, req)
+	project, volKey, err := gceCS.validateControllerPublishVolumeRequest(ctx, req)
 	if err != nil {
 		return nil, err
 	}
-
+	diskToPublish, _ := gceCS.CloudProvider.GetDisk(ctx, project, volKey, gce.GCEAPIVersionV1)
```
msau42, line 492: this may need some more refactoring. `GetDisk` is already called inside `executeControllerPublishVolume`

msau42, line 492: I think we should move the GetDisk after the backoff check so that we don't spam GetDisk during the backoff period

msau42: ditto for unpublish

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -434,12 +452,14 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
 	}
 
 	project, volKey, err = gceCS.CloudProvider.RepairUnderspecifiedVolumeKey(ctx, project, volKey)
+	disk, _ := gceCS.CloudProvider.GetDisk(ctx, project, volKey, gce.GCEAPIVersionV1)
```
msau42, line 455: This should probably be done after we get the operation lock

pkg/metrics/metrics.go
```diff
@@ -123,3 +146,13 @@ func IsGKEComponentVersionAvailable() bool {
 
 	return true
 }
+
+func GetDiskType(disk *gce.CloudDisk) string {
+	var diskType string
+	if disk != nil {
+		diskType = disk.GetPDType()
+	} else {
+		diskType = ""
```
sunnylovestiramisu, line 155: Do not think we need this else block. diskType is initialized as "".

PR #1232, opened 2023-05-24
Use errors.As so we can detect wrapped errors, and check for existing error codes in CodesForError

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Use errors.As so we can detect wrapped errors, and check for existing error codes in CodesForError. Currently wrapped errors cannot be detected in IsGCEError. Also if we return a grpc error code that is not an error we filter out by LoggedError, it will get turned into an Internal error.  

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -231,6 +231,13 @@ func containsZone(zones []string, zone string) bool {
 // (1) "context deadline exceeded", returns grpc DeadlineExceeded,
 // (2) "context canceled", returns grpc Canceled
 func CodeForError(err error) *codes.Code {
+	if err == nil {
+		return nil
+	}
+
+	if errCode := existingErrorCode(err); errCode != nil {
+		return errCode
+	}
 	if code := isContextError(err); code != nil {
 		return code
 	}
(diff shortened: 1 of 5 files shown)


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -231,6 +231,13 @@ func containsZone(zones []string, zone string) bool {
 // (1) "context deadline exceeded", returns grpc DeadlineExceeded,
 // (2) "context canceled", returns grpc Canceled
 func CodeForError(err error) *codes.Code {
+	if err == nil {
```
sunnylovestiramisu, line 234: FYI: This [PR#1227](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1227/) moved CodeForError to the common pkg

amacaskill: That PR isn't merged yet, so I think its ok to merge this as is.

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -255,6 +262,16 @@ func CodeForError(err error) *codes.Code {
 	return &internalErrorCode
 }
 
+func existingErrorCode(err error) *codes.Code {
+	if err == nil {
+		return nil
+	}
+	if status, ok := status.FromError(err); ok {
+		return errCodePtr(status.Code())
```
sunnylovestiramisu, line 270: Since we are returning from multiple places, should we add more logs? For example on line 239, 249, 262 can return internal error code but we do not know exactly which one is returning it.

amacaskill: I don't really think it matters if the internal error is returned on 239, 249, or 262. That being said, if we really want to know what the error was/why internal error was returned, `LoggedError` should have already logged what the error was before entering the `CodeForError` function, so I think adding logging within here is a little excessive and unnecessary. What do you think?

PR #1251, opened 2023-06-05
Delete stable overlays and update docs with recommended versions.

mattcary:
/kind cleanup

As part of cleaning up the testgrids (see https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver), we've decided to clean up the old k8s version overlays. We no longer have test jobs that verify that these overlays still actually work so it seems better not to promise as such.

```release-note
None
```

/assign @msau42


README.md
```diff
@@ -21,34 +21,28 @@ Latest stable image: `registry.k8s.io/cloud-provider-gcp/gcp-compute-persistent-
...
 
-The manifest bundle which captures all the driver components (driver pod which includes the containers csi-provisioner, csi-resizer, csi-snapshotter, gce-pd-driver, csi-driver-registrar;
-csi driver object, rbacs, pod security policies etc) can be picked up from the master branch [overlays](deploy/kubernetes/overlays) directory. We structure the overlays directory, per minor version of kubernetes because not all driver components can be used with all kubernetes versions.
-
-Example:
-
-`stable-1-21` overlays bundle can be used to deploy all the components of the driver on kubernetes 1.21.
+| Kubernetes Version | PD CSI Driver Version |
+|--------------------|-----------------------|
+| 1.28               | v1.10.2               |
```
msau42, line 38: 1.28 is still in active development so not sure if we want to put HEAD as a placeholder instead?

mattcary: Ah yeah good point.

README.md
```diff
@@ -21,34 +21,28 @@ Latest stable image: `registry.k8s.io/cloud-provider-gcp/gcp-compute-persistent-
...
 
-The manifest bundle which captures all the driver components (driver pod which includes the containers csi-provisioner, csi-resizer, csi-snapshotter, gce-pd-driver, csi-driver-registrar;
-csi driver object, rbacs, pod security policies etc) can be picked up from the master branch [overlays](deploy/kubernetes/overlays) directory. We structure the overlays directory, per minor version of kubernetes because not all driver components can be used with all kubernetes versions.
-
-Example:
-
-`stable-1-21` overlays bundle can be used to deploy all the components of the driver on kubernetes 1.21.
+| Kubernetes Version | PD CSI Driver Version |
+|--------------------|-----------------------|
+| HEAD               | v1.10.2               |
```
msau42, line 38: Maybe instead of the exact driver patch version, we can just specify the minor version, ie `1.10.x`. One less thing to update whenever we do a patch release.

mattcary: Good idea, done

mattcary:
Oops, I dropped this. I responded to your comments.

mattcary:
ping? @msau42

PR #1252, opened 2023-06-06
Remove nouid32 which breaks xfs

mattcary:
See https://github.com/kubernetes/cloud-provider-gcp/issues/557 and [this example of a failing test](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-gcp-compute-persistent-disk-csi-driver-release-k8s-master-integration/1665792101725507584).

It appears the nouid32 mount option is not supported by XFS.

I guess it works with ext4 --- but brief internet searching didn't find anything about it, so I don't know if it's something we need to support?
/kind failing-test

```release-note
None
```

/assign @msau42

### DIFF
--- a/test/k8s-integration/config/test-config-template.in
+++ b/test/k8s-integration/config/test-config-template.in
@@ -22,7 +22,6 @@ DriverInfo:
     NumRestarts: 10
   SupportedMountOption:
     debug:
-    nouid32:
   SupportedSizeRange:
     Min: {{.MinimumVolumeSize}}
     Max: 64Ti


msau42:
https://man7.org/linux/man-pages/man5/ext4.5.html
```
       nouid32
              Disables 32-bit UIDs and GIDs.  This is for
              interoperability with older kernels which only store and
              expect 16-bit values.
```

Sounds fine to remove. The test is only testing that mount options work, the specific options are not important.

PR #1254, opened 2023-06-07
Add force attach to storage class

mattcary:
/kind feature

This adds an alpha-level storage class parameter to always perform a force attach. This is only valid for RePD disks (although this is not validated --- using this parameter with zonal disks will cause attaches to fail).

This parameter is not guaranteed to have continued support or to be stable in any way. In addition, using this can cause data loss.

```release-note
Add alpha-level force attach storage class parameter.
```

### DIFF
--- a/pkg/common/parameters.go
+++ b/pkg/common/parameters.go
@@ -29,6 +29,7 @@ const (
 	ParameterKeyLabels                        = "labels"
 	ParameterKeyProvisionedIOPSOnCreate       = "provisioned-iops-on-create"
 	ParameterKeyProvisionedThroughputOnCreate = "provisioned-throughput-on-create"
+	ParameterAvailabilityClass                = "availability-class"
 
 	// Parameters for VolumeSnapshotClass
 	ParameterKeyStorageLocations = "storage-locations"
@@ -37,6 +38,10 @@ const (
 	DiskSnapshotType             = "snapshots"
 	DiskImageType                = "images"
 	replicationTypeNone          = "none"
+
+	// Parameters for AvailabilityClass
+	ParameterNoAvailabilityClass       = "none"
+	ParameterRegionalHardFailoverClass = "regional-hard-failover"
 
 	// Keys for PV and PVC parameters as reported by external-provisioner
 	ParameterKeyPVCName      = "csi.storage.k8s.io/pvc/name"
@@ -83,6 +88,8 @@ type DiskParameters struct {
 	// Values: {int64}
 	// Default: none
 	ProvisionedThroughputOnCreate int64
+	// Default: false
+	ForceAttach bool
 }
 
 // SnapshotParameters contains normalized and defaulted parameters for snapshots
@@ -155,6 +162,14 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 				return p, fmt.Errorf("parameters contain invalid provisionedThroughputOnCreate parameter: %w", err)
 			}
 			p.ProvisionedThroughputOnCreate = paramProvisionedThroughputOnCreate
+		case ParameterAvailabilityClass:
+			paramAvailabilityClass, err := ConvertStringToAvailabilityClass(v)
+			if err != nil {
+				return p, fmt.Errorf("parameters contain invalid availability class parameter: %w", err)
+			}
+			if paramAvailabilityClass == ParameterRegionalHardFailoverClass {
+				p.ForceAttach = true
+			}
 		default:
 			return p, fmt.Errorf("parameters contains invalid option %q", k)
 		}
--- a/pkg/common/parameters_test.go
+++ b/pkg/common/parameters_test.go
@@ -178,6 +178,27 @@ func TestExtractAndDefaultParameters(t *testing.T) {
 				Labels:               map[string]string{"key1": "value1", "label-1": "value-a", "label-2": "label-value-2"},
 			},
 		},
+		{
+			name:       "availability class parameters",
+			parameters: map[string]string{ParameterAvailabilityClass: ParameterRegionalHardFailoverClass},
+			expectParams: DiskParameters{
+				DiskType:        "pd-standard",
+				ReplicationType: "none",
+				ForceAttach:     true,
+				Tags:            map[string]string{},
+				Labels:          map[string]string{},
+			},
+		},
+		{
+			name:       "no force attach parameters",
+			parameters: map[string]string{ParameterAvailabilityClass: ParameterNoAvailabilityClass},
+			expectParams: DiskParameters{
+				DiskType:        "pd-standard",
+				ReplicationType: "none",
+				Tags:            map[string]string{},
+				Labels:          map[string]string{},
+			},
+		},
 	}
 
 	for _, tc := range tests {
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -284,6 +284,28 @@ func ConvertMiStringToInt64(str string) (int64, error) {
 	return volumehelpers.RoundUpToMiB(quantity)
 }
 
+// ConvertStringToBool converts a string to a boolean.
+func ConvertStringToBool(str string) (bool, error) {
+	switch strings.ToLower(str) {
+	case "true":
+		return true, nil
+	case "false":
+		return false, nil
+	}
+	return false, fmt.Errorf("Unexpected boolean string %s", str)
+}
+
+// ConvertStringToAvailabilityClass converts a string to an availability class string.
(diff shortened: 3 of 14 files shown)


pkg/common/utils.go
```diff
@@ -284,6 +284,17 @@ func ConvertMiStringToInt64(str string) (int64, error) {
 	return volumehelpers.RoundUpToMiB(quantity)
 }
 
+// ConvertStringToBool converts a true/false string into a bool, or error.
+func ConvertStringToBool(str string) (bool, error) {
```
mattcary, line 288: I feel like there should be a standard k8s conversion library to use, but I couldn't find it.

msau42: k8s api convention is to avoid bools as much as possible, because in the future if you want to add a 3rd value it gets hard: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

msau42:
What's the plan for detach? Should repd always use force attach regardless?

pkg/common/utils.go
mattcary: Yes, however the gce api this maps to is a bool: https://cloud.google.com/compute/docs/reference/rest/v1/instances/attachDisk

I think at this point keeping this mapping direct is best, we don't know what abstraction layer we want to have between this driver and the api yet.

mattcary:
> What's the plan for detach? Should repd always use force attach regardless?

We don't know. For the PoC of an availability controller we're building we will be doing force attach unconditionally, because as far as we can tell it takes too long to figure out if it's necessary or not (this is what cloud sql does for instance).

But we have to evaluate. This is why I have the scary alpha/unsafe name, I think there's a good chance well figure out a better solution after we work through real-world options. I expect that this parameter will be removed & replaced with a better one.

msau42:
I'm wondering for the purpose of POC you can just have a feature gate, instead of an opt-in storageclass parameter

mattcary:
> I'm wondering for the purpose of POC you can just have a feature gate, instead of an opt-in storageclass parameter that ends up as an immutable field on the PV object

Do you mean a command-line flag that would enable this for the driver per-cluster? We need to enable this only on certain volumes, we'll be testing it in clusters where we only want some workloads to use this as we don't want to risk data loss by force-attaching cavalierly.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -714,9 +714,13 @@ func (cloud *CloudProvider) AttachDisk(ctx context.Context, project string, volK
...
 		Source:     source,
 		Type:       diskType,
+		// This parameter is ignored in the call, the ForceAttach decorator
+		// (query parameter) is the important one. We'll set it in both places
+		// in case that behavior changes.
+		ForceAttach: forceAttach,
 	}
 
-	op, err := cloud.service.Instances.AttachDisk(project, instanceZone, instanceName, attachedDiskV1).Context(ctx).Do()
+	op, err := cloud.service.Instances.AttachDisk(project, instanceZone, instanceName, attachedDiskV1).Context(ctx).ForceAttach(forceAttach).Do()
```
msau42, line 723: Eventually if this goes to production we may want a metric + log

pkg/common/parameters.go
```diff
@@ -155,6 +158,12 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 				return p, fmt.Errorf("parameters contain invalid provisionedThroughputOnCreate parameter: %w", err)
 			}
 			p.ProvisionedThroughputOnCreate = paramProvisionedThroughputOnCreate
+		case ParameterForceAttach:
```
msau42, line 161: do you want to validate that it's only set for repd?

pkg/gce-cloud-provider/compute/gce-compute.go
mattcary: The api call is in the audit logs, I'm not sure we need much more?

pkg/common/parameters.go
mattcary: Maybe not here (especially if we go with the availabilityClass idea), but certainly should validate before making the attach call, as using force attach on a zonal disk results in an error according to the docs.

pkg/common/utils.go
msau42: You can still validate the values. It does shift the meaning of the field to express a higher level intent and opens it up to more options in the future

mattcary: Yeah, fair enough. Anyway we don't need to decide that now. The concern I have is that if change things we may not use some of these alpha values but we don't want to reject old storage classes that still have the parameters. Maybe there's some way to do a warning.

pkg/common/parameters.go
mattcary: Done, I ended up doing the validation in the create volume request.

mattcary:
Hmm, the e2e test succeeded locally after fixing the request-vs-query param problem. I'll have to dig.

test/e2e/tests/multi_zone_e2e_test.go
```diff
@@ -141,17 +149,114 @@ var _ = Describe("GCE PD CSI Driver Multi-Zone", func() {
...
+		controllerInstance := controllerContext.Instance
+
+		p, _, _ := controllerInstance.GetIdentity()
+
+		region, err := common.GetRegionFromZones(zones)
+		Expect(err).To(BeNil(), "Failed to get region from zones")
+
+		// Create Disk
+		volName := testNamePrefix + string(uuid.NewUUID())
+		volID, err := controllerClient.CreateVolume(volName, map[string]string{
```
pwschuurman, line 184: Would it be worth it to also validate that the volumeContext is set correctly, when availability-class is set?

pkg/common/parameters.go
```diff
@@ -155,6 +162,14 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 				return p, fmt.Errorf("parameters contain invalid provisionedThroughputOnCreate parameter: %w", err)
 			}
 			p.ProvisionedThroughputOnCreate = paramProvisionedThroughputOnCreate
+		case ParameterAvailabilityClass:
+			paramAvailabilityClass, err := ConvertStringToAvailabilityClass(v)
+			if err != nil {
+				return p, fmt.Errorf("parameters contain invalid parameterForceAttach parameter: %w", err)
```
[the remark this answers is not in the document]
mattcary: Oops, the perils of changing the api during the review, thanks.

test/remote/client-wrappers.go
```diff
@@ -134,13 +134,18 @@ func (c *CsiClient) DeleteVolume(volId string) error {
...
+func (c *CsiClient) ControllerPublishVolume(volId, nodeId string, forceAttach bool) error {
 	cpreq := &csipb.ControllerPublishVolumeRequest{
 		VolumeId:         volId,
 		NodeId:           nodeId,
 		VolumeCapability: stdVolCap,
 		Readonly:         false,
 	}
+	if forceAttach {
+		cpreq.VolumeContext = map[string]string{
+			"alpha-unsafe-force-attach": "true",
```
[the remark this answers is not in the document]
mattcary: Yup that might be the failure.

pkg/common/parameters_test.go
```diff
@@ -178,6 +178,27 @@ func TestExtractAndDefaultParameters(t *testing.T) {
 				Labels:               map[string]string{"key1": "value1", "label-1": "value-a", "label-2": "label-value-2"},
 			},
 		},
+		{
+			name:       "availabilty class parameters",
```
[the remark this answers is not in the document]
mattcary: done

test/e2e/tests/multi_zone_e2e_test.go
mattcary: good point, thx.

pkg/common/parameters.go
```diff
@@ -155,6 +162,14 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 				return p, fmt.Errorf("parameters contain invalid provisionedThroughputOnCreate parameter: %w", err)
 			}
 			p.ProvisionedThroughputOnCreate = paramProvisionedThroughputOnCreate
+		case ParameterAvailabilityClass:
+			paramAvailabilityClass, err := ConvertStringToAvailabilityClass(v)
+			if err != nil {
+				return p, fmt.Errorf("parameters contain invalid availability class parameter: %w", err)
```
pwschuurman, line 168: nit: Hyphenate "availability-class" to make it more clear what parameter is invalid.

pwschuurman (approved):
/lgtm

PR #1260, opened 2023-06-13
Use v1.10.3 as RC

mattcary:
> /kind cleanup

```release-note
None
```

/assign @pwschuurman


mattcary:
Merged as the integration test won't use the rc deployment anyway.

PR #1261, opened 2023-06-14
Filter out GCE operation error codes caused by user errors

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

 /kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Filter out GCE operation error codes caused by user errors. We are getting lots of errors counting against our SLO for GCE operation errors that are user errors. We have seen Internal errors returned from the following GCE operation error codes: `OPERATION_CANCELED_BY_USER`, `RESOURCE_NOT_FOUND`, `RESOURCE_IN_USE_BY_ANOTHER_RESOURCE`, `INVALID_USAGE`. I added a filtered out other error codes (quota error codes) which we haven't observed yet, but those are clearly user error codes. The errors that we have seen so far look like this:
```
ControllerPublishVolume returned with error: rpc error: code = Internal desc = unknown Attach error: failed when waiting for zonal op: operation operation-xxx failed (OPERATION_CANCELED_BY_USER): Operation was canceled by user ''.

ControllerPublishVolume returned with error: rpc error: code = Internal desc = unknown Attach error: failed when waiting for zonal op: operation operation-xxx failed (RESOURCE_NOT_FOUND): The resource 'projects/xxx/zones/xxx/instances/xxx' was not found

ControllerPublishVolume returned with error: rpc error: code = Internal desc = unknown Attach error: failed when waiting for zonal op: operation operation-xxx failed (RESOURCE_IN_USE_BY_ANOTHER_RESOURCE): The disk resource 'projects/xxx/zones/xxx/disks/xxx' is already being used by 'projects/xxx/zones/xxx/instances/xxx'

ControllerUnpublishVolume returned with error: rpc error: code = Internal desc = unknown detach error: operation operation-xxx failed (INVALID_USAGE): No attached disk found with device name 'pvc-xxx'
```

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -856,6 +856,10 @@ func (cloud *CloudProvider) WaitForAttach(ctx context.Context, project string, v
 }
 
 func wrapOpErr(name string, opErr *computev1.OperationErrorErrors) error {
+	if opErr == nil {
+		return nil
+	}
+
 	if opErr.Code == "UNSUPPORTED_OPERATION" {
 		if diskType := pdDiskTypeUnsupportedRegex.FindStringSubmatch(opErr.Message); diskType != nil {
 			return &UnsupportedDiskError{
@@ -863,7 +867,29 @@ func wrapOpErr(name string, opErr *computev1.OperationErrorErrors) error {
 			}
 		}
 	}
-	return fmt.Errorf("operation %v failed (%v): %v", name, opErr.Code, opErr.Message)
+	grpcErrCode := codeForGCEOpError(*opErr)
+	return status.Errorf(grpcErrCode, "operation %v failed (%v): %v", name, opErr.Code, opErr.Message)
+}
+
+// codeForGCEOpError return the grpc error code for the passed in
+// gce operation error.
+func codeForGCEOpError(err computev1.OperationErrorErrors) codes.Code {
+	userErrors := map[string]codes.Code{
+		"RESOURCE_NOT_FOUND":                        codes.NotFound,
+		"RESOURCE_ALREADY_EXISTS":                   codes.AlreadyExists,
+		"RESOURCE_IN_USE_BY_ANOTHER_RESOURCE":       codes.InvalidArgument,
+		"OPERATION_CANCELED_BY_USER":                codes.Aborted,
+		"QUOTA_EXCEEDED":                            codes.ResourceExhausted,
+		"ZONE_RESOURCE_POOL_EXHAUSTED":              codes.Unavailable,
+		"ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS": codes.Unavailable,
+		"REGION_QUOTA_EXCEEDED":                     codes.ResourceExhausted,
+		"RATE_LIMIT_EXCEEDED":                       codes.ResourceExhausted,
+		"INVALID_USAGE":                             codes.InvalidArgument,
+	}
+	if code, ok := userErrors[err.Code]; ok {
+		return code
+	}
+	return codes.Internal
 }
 
 func opIsDone(op *computev1.Operation) (bool, error) {
--- a/pkg/gce-cloud-provider/compute/gce-compute_test.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 
 	computev1 "google.golang.org/api/compute/v1"
+	"google.golang.org/grpc/codes"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -856,14 +856,40 @@ func (cloud *CloudProvider) WaitForAttach(ctx context.Context, project string, v
...
+// codeForGCEOpError return the grpc error code for the passed in
+// gce operation error.
+func codeForGCEOpError(err computev1.OperationErrorErrors) codes.Code {
+	userErrors := map[string]codes.Code{
+		"RESOURCE_NOT_FOUND":                        codes.NotFound,
+		"RESOURCE_ALREADY_EXISTS":                   codes.AlreadyExists,
+		"RESOURCE_IN_USE_BY_ANOTHER_RESOURCE":       codes.InvalidArgument,
+		"OPERATION_CANCELED_BY_USER":                codes.Aborted,
+		"QUOTA_EXCEEDED":                            codes.ResourceExhausted,
+		"ZONE_RESOURCE_POOL_EXHAUSTED":              codes.Unavailable,
```
saikat-royc, line 883: should we use codes.ResourceExhausted for ZONE_RESOURCE_POOL_EXHAUSTED, ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS as well

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -856,14 +856,40 @@ func (cloud *CloudProvider) WaitForAttach(ctx context.Context, project string, v
 }
 
 func wrapOpErr(name string, opErr *computev1.OperationErrorErrors) error {
```
saikat-royc, line 858: Also what about error handling for cases where the operation is not even initiated, and we encounter error in say insertDisk, attachDisk where the WaitForOp is not called. e.g [code](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L724)

Also we may need to handle uncastable error (i.e errors.As cannot case to gce.UnsupportedDiskError) like the ones we saw in filestore. May need some manual testing here (1. insert disk when quota is exhausted, 2. attachdisk when disk already attached) or we can address them lazily if we encounter them.

pkg/gce-cloud-provider/compute/gce-compute.go
amacaskill: For these codes, it looks like GCE maps it to the Unavailable grpc error code, so that is why I chose Unavailable. That being said, we don't filter out Unavailable fromm our SLO, so if we don't want these, then I should use resource exhausted. What do you think?

pkg/gce-cloud-provider/compute/gce-compute.go
amacaskill: I have looked into quiet a few clusters, and I haven't seen any errors that are incorrectly returned as an Internal error at this point ( where the operation is not even initiated). I will look into some more clusters, but I have probably looked into at least 20 at this point, and I haven't seen errors like that. 

> Also we may need to handle uncastable error (i.e errors.As cannot cast to gce.UnsupportedDiskError)  like the ones we saw in filestore. 

I'm a little confused by this comment. Are you saying we could be missing some cases where we need to  [return an UnsupportedDiskError](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/26fe3898b13083a02e28fa5dd16410c15136e2c5/pkg/gce-cloud-provider/compute/gce-compute.go#L859) (which we do when the operation error has the UNSUPPORTED_OPERATION code). This won't happen like in filestore because wrapOpErr passes in an operation error of type computev1.OperationErrorErrors directly. This error type doesn't actually implement the error interface, so this will never happen. (errors.As cannot even be run on the operation error of type computev1.OperationErrorErrors since it doesn't implement error interface).

pkg/gce-cloud-provider/compute/gce-compute.go
judemars: We looked at this for the CreateVolume SLO as well and ultimately decided we would want to get notified if users were hitting an irregular amount of ZONE_RESOURCE_POOL_EXHAUSTED errors since they're not under their control. So +1 to Unavailable since it allows us to track it (or not) separately from other user-caused ResourceExhausted errors.

Also it looks like this PR accomplishes what I was doing in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1263 (classifying QUOTA_EXCEEDED as ResourceExhausted) so I'll close that - thanks!

saikat-royc: Ack. Please capture the reasoning in a comment in the code

pkg/gce-cloud-provider/compute/gce-compute.go
saikat-royc: Ack. I missed the OperationErrorErrors type. 

And yes, for unsupported operations, I think we would return kIternal for the UnsupportedDiskError. Until we repro this scenario, it would be difficult to handle this error. So that would be a followup for future to understand the sceario better.

As discussed, lets capture the known issues in some task:
1. AttachDisk errors not handled by errors.As() [code](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/release-1.10/pkg/gce-pd-csi-driver/controller.go#L638)
2. [InsertDisk](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/release-1.10/pkg/gce-pd-csi-driver/controller.go#L1781), [DeleteDisk](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/release-1.10/pkg/gce-pd-csi-driver/controller.go#L480), [DetachDisk](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/release-1.10/pkg/gce-pd-csi-driver/controller.go#L758) errors are not handled.

PR #1262, opened 2023-06-16
Remove updating disk type for existing disks

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The variable `diskTypeForMetric` is currently being overwritten when we check for existing disks in Create Volume calls. Removing the assignment for `diskTypeForMetric` in Create volume while validating existing disk. 

Fixes #

This is related to PR #1227

```release-note
Bugfix for empty disk type being registered in metric for Create volume function.

```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -283,7 +283,6 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	// Validate if disk already exists
 	existingDisk, err := gceCS.CloudProvider.GetDisk(ctx, gceCS.CloudProvider.GetDefaultProject(), volKey, gceAPIVersion)
-	diskTypeForMetric = metrics.GetDiskType(existingDisk)
 	if err != nil {
 		if !gce.IsGCEError(err, "notFound") {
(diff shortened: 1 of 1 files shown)


mattcary:
Update the release note (something like "Bugfix for blank disk type in volume creation metric").

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -283,7 +283,6 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 	// Validate if disk already exists
 	existingDisk, err := gceCS.CloudProvider.GetDisk(ctx, gceCS.CloudProvider.GetDefaultProject(), volKey, gceAPIVersion)
-	diskTypeForMetric = metrics.GetDiskType(existingDisk)
```
msau42, line 286: What happens if the disk type of the existing disk doesn't match the disk type being requested in the storageclass?

Sneha-at: The existing disk is being validated on 293 below 
```
err = gceCS.CloudProvider.ValidateExistingDisk
```
The function `ValidateExistingDisk` checks for disk type, disk size, location and other params.

PR #1278, opened 2023-06-27
Fix provisioned-iops-on-create passing logic

sunnylovestiramisu:
**What type of PR is this?**

/kind bug

The logic of passing in provisioned-iops-on-create is not at the right func. The tests picked the default value and the verification is passing.

Fixes #

```release-note
Fix provisioned-iops-on-create passing logic
```

### DIFF
--- a/pkg/common/utils_test.go
+++ b/pkg/common/utils_test.go
@@ -598,6 +598,12 @@ func TestConvertStringToInt64(t *testing.T) {
 			expectError: false,
 		},
 		{
+			desc:        "test higher number",
+			inputStr:    "15000",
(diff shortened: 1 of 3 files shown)


mattcary:
still lgtm

/lgtm

sunnylovestiramisu:
Okay the number failed again, i will try via gcloud cli first before finalize the number.

The validation for pd-extreme seems off, if we do not specify iops, it sets to 100000. However if we set iops, it triggers validations.

sunnylovestiramisu:
hack/verify-all.sh succeeded locally, do not know why it is not passing here.

PR #1286, opened 2023-06-28
chore: unnecessary use of fmt.Sprintf

testwill:
**What type of PR is this?**
> /kind cleanup

unnecessary use of fmt.Sprintf

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -167,7 +167,7 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 			return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to create block file at target path %v: %v", targetPath, err.Error()))
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -167,7 +167,7 @@ func (ns *GCENodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePub
 			return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to create block file at target path %v: %v", targetPath, err.Error()))
 		}
 	} else {
-		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("NodePublishVolume volume capability must specify either mount or block mode"))
+		return nil, status.Error(codes.InvalidArgument, "NodePublishVolume volume capability must specify either mount or block mode")
```
sunnylovestiramisu, line 170: Why this is the only one that is not needed? What about other status.Error in the above lines?

PR #1291, opened 2023-07-11
Added support for confidential storage to disk parametrs and metric

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR adds support in PDCSI driver to create confidential hyperdisk storage on GCE.

Fixes #

This change has been tested manually as we cannot add e2e tests yet.
Scenarios tested:
1. Create a hyperdisk-balanced confidential disk with PDCSI driver.
2. Created a confidential disk from a non-confidential snapshot (Snapshot source disk type: pd-standard)

```release-note
Added support in PDCSI driver to create confidential hyperdisk storage on GCE.
```

### DIFF
--- a/pkg/common/parameters.go
+++ b/pkg/common/parameters.go
@@ -30,6 +30,7 @@ const (
 	ParameterKeyProvisionedIOPSOnCreate       = "provisioned-iops-on-create"
 	ParameterKeyProvisionedThroughputOnCreate = "provisioned-throughput-on-create"
 	ParameterAvailabilityClass                = "availability-class"
+	ParameterKeyEnableConfidentialCompute     = "enable-confidential-storage"
 
 	// Parameters for VolumeSnapshotClass
 	ParameterKeyStorageLocations = "storage-locations"
@@ -88,6 +89,9 @@ type DiskParameters struct {
 	// Values: {int64}
 	// Default: none
 	ProvisionedThroughputOnCreate int64
+	// Values: {bool}
+	// Default: false
+	EnableConfidentialCompute bool
 	// Default: false
 	ForceAttach bool
 }
@@ -170,6 +174,21 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 			if paramAvailabilityClass == ParameterRegionalHardFailoverClass {
 				p.ForceAttach = true
 			}
+		case ParameterKeyEnableConfidentialCompute:
+			paramEnableConfidentialCompute, err := ConvertStringToBool(v)
+			if err != nil {
+				return p, fmt.Errorf("parameters contain invalid value for enable-confidential-storage parameter: %w", err)
+			}
+
+			if paramEnableConfidentialCompute {
+				// DiskEncryptionKmsKey is needed to enable confidentialStorage
+				if val, ok := parameters[ParameterKeyDiskEncryptionKmsKey]; !ok || !isValidDiskEncryptionKmsKey(val) {
+					return p, fmt.Errorf("Valid %v is required to enbale ConfidentialStorage", ParameterKeyDiskEncryptionKmsKey)
+				}
+			}
+
+			p.EnableConfidentialCompute = paramEnableConfidentialCompute
+
 		default:
 			return p, fmt.Errorf("parameters contains invalid option %q", k)
 		}
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -397,3 +397,9 @@ func LoggedError(msg string, err error) error {
 	klog.Errorf(msg+"%v", err.Error())
 	return status.Errorf(*CodeForError(err), msg+"%v", err.Error())
 }
+
+func isValidDiskEncryptionKmsKey(DiskEncryptionKmsKey string) bool {
+	// Validate key against default kmskey pattern
+	kmsKeyPattern := regexp.MustCompile("projects/[^/]+/locations/([^/]+)/keyRings/[^/]+/cryptoKeys/[^/]+")
+	return kmsKeyPattern.MatchString(DiskEncryptionKmsKey)
+}
--- a/pkg/gce-cloud-provider/compute/cloud-disk.go
+++ b/pkg/gce-cloud-provider/compute/cloud-disk.go
@@ -223,3 +223,14 @@ func (d *CloudDisk) GetMultiWriter() bool {
 		return false
 	}
 }
+
+func (d *CloudDisk) GetEnableConfidentialCompute() bool {
+	switch {
+	case d.disk != nil:
+		return false
+	case d.betaDisk != nil:
+		return d.betaDisk.EnableConfidentialCompute
+	default:
+		return false
+	}
+}
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -43,7 +43,7 @@ const (
 	pdDiskTypeUnsupportedPattern = `\[([a-z-]+)\] features are not compatible for creating instance`
 )
 
-var hyperdiskTypes = []string{"hyperdisk-extreme", "hyperdisk-throughput"}
+var hyperdiskTypes = []string{"hyperdisk-extreme", "hyperdisk-throughput", "hyperdisk-balanced"}
 var pdDiskTypeUnsupportedRegex = regexp.MustCompile(pdDiskTypeUnsupportedPattern)
 
 type GCEAPIVersion string
@@ -258,6 +258,9 @@ func (cloud *CloudProvider) ListSnapshots(ctx context.Context, filter string) ([
 
 func (cloud *CloudProvider) GetDisk(ctx context.Context, project string, key *meta.Key, gceAPIVersion GCEAPIVersion) (*CloudDisk, error) {
 	klog.V(5).Infof("Getting disk %v", key)
+
+	// Override GCEAPIVersion as hyperdisk is only available in beta and we cannot get the disk-type with get disk call.
+	gceAPIVersion = GCEAPIVersionBeta
 	switch key.Type() {
 	case meta.Zonal:
 		if gceAPIVersion == GCEAPIVersionBeta {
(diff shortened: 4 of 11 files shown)


pkg/common/parameters.go
```diff
@@ -88,6 +89,9 @@ type DiskParameters struct {
 	// Values: {int64}
 	// Default: none
 	ProvisionedThroughputOnCreate int64
+	// Values: {bool}
+	//Default: False
```
sunnylovestiramisu, line 93: Space after // :)

pkg/common/parameters.go
```diff
@@ -170,6 +174,21 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
...
 			}
+		case ParameterKeyEnbaleConfidentialStorage:
+			paramEnbaleConfidentialStorage, err := ConvertStringToBool(v)
+			if err != nil {
+				return p, fmt.Errorf("parameters contain invalid value for EnbaleConfidentialStorage parameter: %w", err)
+			}
+
+			if paramEnbaleConfidentialStorage {
+				// DiskEncryptionKmsKey is needed to enable confidentialStorage
+				if val, ok := parameters[ParameterKeyDiskEncryptionKmsKey]; !ok || val == "" {
```
sunnylovestiramisu, line 185: Should we validate the regex of the DiskEncryptionKmsKey?

pkg/gce-cloud-provider/compute/cloud-disk.go
```diff
@@ -223,3 +223,14 @@ func (d *CloudDisk) GetMultiWriter() bool {
 		return false
 	}
 }
+
+func (d *CloudDisk) GetEnableConfidentialStorage() bool {
```
sunnylovestiramisu, line 227: Add unit test coverage for this?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -258,6 +258,9 @@ func (cloud *CloudProvider) ListSnapshots(ctx context.Context, filter string) ([
 
 func (cloud *CloudProvider) GetDisk(ctx context.Context, project string, key *meta.Key, gceAPIVersion GCEAPIVersion) (*CloudDisk, error) {
 	klog.V(5).Infof("Getting disk %v", key)
+
+	// Override GCEAPIVersion as hyperdisk is only available in beta and we cannot get the disk-type with get disk call.
```
sunnylovestiramisu, line 262: Nit: If you think we can switch to only use Beta APIs for the whole pdcsi driver and it will be easier to void conversion between v1 and beta, you can make the change to remove usage of v1 GCE.

If it is more complicated, let's do refactor later.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -558,7 +561,6 @@ func (cloud *CloudProvider) insertZonalDisk(
 		opName        string
 		gceAPIVersion = GCEAPIVersionV1
 	)
-
 	if multiWriter || containsBetaDiskType(hyperdiskTypes, params.DiskType) {
```
sunnylovestiramisu, line 564: So since the default right now is Beta API, should this logic also change? Do we still want to use V1 for PD disks?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -204,8 +205,9 @@ func useVolumeCloning(req *csi.CreateVolumeRequest) bool {
 func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	var err error
 	diskTypeForMetric := ""
+	enableConfidentialStorage := ""
```
sunnylovestiramisu, line 208: Shouldn't the default be false for the metrics?

pkg/metrics/metrics.go
```diff
@@ -152,10 +154,12 @@ func IsGKEComponentVersionAvailable() bool {
 	return true
 }
 
-func GetDiskType(disk *gce.CloudDisk) string {
+func GetMetricParameters(disk *gce.CloudDisk) (string, string) {
```
sunnylovestiramisu, line 157: unit test this for beta disk and v1 disk?

pkg/gce-cloud-provider/compute/cloud-disk.go
```diff
@@ -223,3 +223,14 @@ func (d *CloudDisk) GetMultiWriter() bool {
...
+	switch {
+	case d.disk != nil:
+		return false
+	case d.betaDisk != nil:
+		return d.betaDisk.EnableConfidentialCompute
```
sunnylovestiramisu, line 232: Side note, I tried to search the string EnableConfidentialCompute in this PR but it only showed up here. You need to pass in EnableConfidentialCompute to the Disk struct for disk.insert GCP PD APIs so I guess that is missing. The disk.insert will not recognize EnableConfidentialStorage.

pkg/common/parameters.go
Sneha-at: I am conflicted on that, I did not find any function which validates the key and I am not sure if we should add another check which can break if something changes underneath. What is your opinion?

pkg/gce-cloud-provider/compute/gce-compute.go
Sneha-at: Currently we have Create Volume and Get Disk calls going to beta version, do we want to include all calls given that we would revert some of these once we have the features in V1? Also, while reverting we will have to ensure to add beta for other features (like multi-writer).

pkg/gce-cloud-provider/compute/gce-compute.go
Sneha-at: Depends on if we decide to move forward with default Beta API

pkg/gce-cloud-provider/compute/cloud-disk.go
Sneha-at: Post our discussion, we decided to rename all values to `EnableConfidentialCompute` to match with PD team

pkg/gce-cloud-provider/compute/gce-compute.go
Sneha-at: We will not be switching other function calls as it might lead to other unforeseen issues.

pkg/gce-cloud-provider/compute/gce-compute.go
Sneha-at: No change needed as we are not switching default to Beta

pkg/common/parameters.go
Sneha-at: Added regex check for the key

pkg/common/utils.go
```diff
@@ -397,3 +397,9 @@ func LoggedError(msg string, err error) error {
 	klog.Errorf(msg+"%v", err.Error())
 	return status.Errorf(*CodeForError(err), msg+"%v", err.Error())
 }
+
+func isValidDiskEncryptionKmsKey(DiskEncryptionKmsKey string) bool {
+	// Validate key against default kmskey pattern
+	kmsKeyPattern := regexp.MustCompile("projects/[^/]+/locations/([^/]+)/keyRings/[^/]+/cryptoKeys/[^/]+")
```
sunnylovestiramisu, line 403: This may worth unit testing, like i think both global key or regional key should pass the regexp check.

pkg/common/parameters.go
```diff
@@ -30,6 +30,7 @@ const (
 	ParameterKeyProvisionedIOPSOnCreate       = "provisioned-iops-on-create"
 	ParameterKeyProvisionedThroughputOnCreate = "provisioned-throughput-on-create"
 	ParameterAvailabilityClass                = "availability-class"
+	ParameterKeyEnableConfidentialCompute     = "enable-confidential-storage"
```
sunnylovestiramisu, line 33: This param needs to be called enable-confidential-storage? The naming should not be changed publicly but we can convert it within the driver to EnableConfidentialCompute.

pkg/common/parameters.go
```diff
@@ -170,6 +174,21 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
 			if paramAvailabilityClass == ParameterRegionalHardFailoverClass {
 				p.ForceAttach = true
 			}
+		case ParameterKeyEnableConfidentialCompute:
+			paramEnableConfidentialCompute, err := ConvertStringToBool(v)
+			if err != nil {
+				return p, fmt.Errorf("parameters contain invalid value for EnableConfidentialCompute parameter: %w", err)
```
sunnylovestiramisu, line 180: Since customers are seeing this message, it may make sense to call it enable-confidential-storage? Otherwise it is confusing.

pkg/common/parameters.go
```diff
@@ -170,6 +174,21 @@ func ExtractAndDefaultParameters(parameters map[string]string, driverName string
...
+			}
+
+			if paramEnableConfidentialCompute {
+				// DiskEncryptionKmsKey is needed to enable confidentialStorage
+				if val, ok := parameters[ParameterKeyDiskEncryptionKmsKey]; !ok || !isValidDiskEncryptionKmsKey(val) {
+					return p, fmt.Errorf("Valid %v is needed to enbale ConfidentialStorage", ParameterKeyDiskEncryptionKmsKey)
```
sunnylovestiramisu, line 186: Nit: how about using "required" instead of "needed"?

sunnylovestiramisu:
Please also document your manual test cases in the PR and verify metrics emission. The rest looks good to me.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -203,9 +204,10 @@ func useVolumeCloning(req *csi.CreateVolumeRequest) bool {
 
 func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	var err error
-	diskTypeForMetric := ""
+	diskTypeForMetric := "unknownDiskType"
+	enableConfidentialCompute := "unkownMode"
```
sunnylovestiramisu, line 208: Typo: unknown instead of unkown

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -440,9 +442,10 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 
 func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVolumeRequest) (*csi.DeleteVolumeResponse, error) {
 	var err error
-	diskTypeForMetric := ""
+	diskTypeForMetric := "unknownDiskType"
+	enableConfidentialCompute := "unkownMode"
```
sunnylovestiramisu, line 446: Same here. And probably a few others, please replace all the 'unkownMode'.

PR #1169, opened 2023-03-08
Serial mkfs

artemvmin:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Add concurrency limit to prevent OOMing when issuing multiple concurrent mkfs calls

**Which issue(s) this PR fixes**:
Fixes #1168

**Does this PR introduce a user-facing change?**:
```release-note
Fixes CVE-2022-41723
```

PR #1294, opened 2023-07-18
Add option to serialize formatAndMount

mattcary:
/kind bug

This a follow-on to #1168 and #1169. In addition to mkfs, we are now seeing reports from the field of fsck OOMing.

This change serializes formatAndMount in this driver. In parallel we will be extended mount-utils, and will remove this fix once the mount-utils release comes out (which releases every month or two).

```release-note
Add option for serializing formatAndMount, including fsck as well as mkfs.
```

/asssign @saikat-royc

### DIFF
--- a/pkg/gce-pd-csi-driver/utils_linux.go
+++ b/pkg/gce-pd-csi-driver/utils_linux.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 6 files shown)


pkg/gce-pd-csi-driver/utils_linux.go
```diff
@@ -47,7 +48,25 @@ func getDevicePath(ns *GCENodeServer, volumeID, partition string) (string, error
 	return devicePath, nil
 }
 
-func formatAndMount(source, target, fstype string, options []string, m *mount.SafeFormatAndMount) error {
+func (ns *GCENodeServer) formatAndMount(source, target, fstype string, options []string, m *mount.SafeFormatAndMount) error {
+	if ns.formatAndMountSemaphore != nil {
+		done := make(chan any)
+		defer close(done)
+
+		ns.formatAndMountSemaphore <- struct{}{}
```
saikat-royc, line 56: can we put a comment here this call blocks until the formatAndMountSemaphore channel becomes empty again (equivalent to an acquire())

pkg/gce-pd-csi-driver/utils_linux.go
```diff
@@ -47,7 +48,25 @@ func getDevicePath(ns *GCENodeServer, volumeID, partition string) (string, error
...
+
+		go func() {
+			defer func() { <-ns.formatAndMountSemaphore }()
+
+			timeout := time.NewTimer(ns.formatAndMountTimeout)
+			defer timeout.Stop()
+
+			select {
+			case <-done:
+			case <-timeout.C:
```
saikat-royc, line 66: just so I understand:
this means we may still have scenarios of concurrent FormatAndMount calls happening if we expire the time

saikat-royc, line 66: so this means, there may still be scenarios where there are concurrent execution of FormatAndMount() calls, if the timer expires, since we release the semaphore?

pkg/gce-pd-csi-driver/utils_linux.go
mattcary: done

pkg/gce-pd-csi-driver/utils_linux.go
saikat-royc: Ack

PR #1298, opened 2023-07-18
Use original error code when backing off

mattcary:
/kind cleanup

This will make it easier to filter errors property for alerting.

-->
```release-note
Use original error code when responding with a backoff error on publish or unpublish.
```

/assign @amacaskill

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -92,10 +92,12 @@ type GCEControllerServer struct {
 	errorBackoff *csiErrorBackoff
 }
 
-type csiErrorBackoff struct {
-	backoff *flowcontrol.Backoff
-}
 type csiErrorBackoffId string
+
+type csiErrorBackoff struct {
+	backoff    *flowcontrol.Backoff
+	errorCodes map[csiErrorBackoffId]codes.Code
+}
 
 type workItem struct {
 	ctx          context.Context
@@ -496,13 +498,13 @@ func (gceCS *GCEControllerServer) ControllerPublishVolume(ctx context.Context, r
 
 	backoffId := gceCS.errorBackoff.backoffId(req.NodeId, req.VolumeId)
 	if gceCS.errorBackoff.blocking(backoffId) {
-		return nil, status.Errorf(codes.Unavailable, "ControllerPublish not permitted on node %q due to backoff condition", req.NodeId)
+		return nil, status.Errorf(gceCS.errorBackoff.code(backoffId), "ControllerPublish not permitted on node %q due to backoff condition", req.NodeId)
 	}
 
 	resp, err, diskTypeForMetric := gceCS.executeControllerPublishVolume(ctx, req)
 	if err != nil {
-		klog.Infof("For node %s adding backoff due to error for volume %s: %v", req.NodeId, req.VolumeId, err.Error())
-		gceCS.errorBackoff.next(backoffId)
+		klog.Infof("For node %s adding backoff due to error for volume %s: %v", req.NodeId, req.VolumeId, err)
+		gceCS.errorBackoff.next(backoffId, common.CodeForError(err))
 	} else {
 		klog.Infof("For node %s clear backoff due to successful publish of volume %v", req.NodeId, req.VolumeId)
 		gceCS.errorBackoff.reset(backoffId)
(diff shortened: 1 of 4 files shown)


pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -3057,13 +3059,16 @@ func TestControllerUnpublishBackoff(t *testing.T) {
 				t.Errorf("expected error")
 			}
 
-			// The above failure should cause driver to call Backoff.Next() again and a backoff duration of 400 ms duration is set starting at the 200th millisecond.
-			// For the 200-599 ms, the backoff condition is true, and new controller publish requests will be deined.
+			// The above failure should cause driver to call backoff.next() again and a
+			// backoff duration of 400 ms duration is set starting at the 200th
+			// millisecond.  For the 200-599 ms, the backoff condition is true, with an
+			// internal error this time, and new controller publish requests will be
+			// deined.
```
amacaskill, line 3066: denied

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -3057,13 +3059,16 @@ func TestControllerUnpublishBackoff(t *testing.T) {
...
 			for i := 0; i < 399; i++ {
 				tc.config.clock.Step(step)
 				var err error
 				_, err = driver.cs.ControllerUnpublishVolume(context.Background(), unpubreq)
-				if !isUnavailableError(err) {
+				if !isInternalError(err) {
```
amacaskill, line 3071: Why  is this checking for Internal errors instead of Unavailable?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -3213,13 +3219,16 @@ func TestControllerPublishBackoff(t *testing.T) {
 				t.Errorf("expected error")
 			}
 
-			// The above failure should cause driver to call Backoff.Next() again and a backoff duration of 400 ms duration is set starting at the 200th millisecond.
-			// For the 200-599 ms, the backoff condition is true, and new controller publish requests will be deined.
+			// The above failure should cause driver to call backoff.next() again and a
+			// backoff duration of 400 ms duration is set starting at the 200th
+			// millisecond.  For the 200-599 ms, the backoff condition is true, with an
+			// internal error this time, and new controller publish requests will be
+			// deined.
```
amacaskill, line 3226: denied

pkg/gce-pd-csi-driver/controller_test.go
mattcary: oops thx done

pkg/gce-pd-csi-driver/controller_test.go
mattcary: Because unavailable is the default if we haven't set an error code for the backoff. I'll clarify in a comment.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1828,10 +1830,27 @@ func (b *csiErrorBackoff) blocking(id csiErrorBackoffId) bool {
...
+	}
+	// If we haven't recorded a code, return unavailable, which signals a problem with the driver
+	// (ie, it's not filtered out as a user problem).
+	klog.Errorf("using default code for %s", id)
+	return codes.Unavailable
+}
+
+func (b *csiErrorBackoff) next(id csiErrorBackoffId, code *codes.Code) {
 	b.backoff.Next(string(id), b.backoff.Clock.Now())
+	if code != nil {
```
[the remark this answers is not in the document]
mattcary: I like making CodeForError return a code and not a pointer, I think that would fix your objection?

I think the main thing is if the code is unset then returning Unavailble, which matches the existing backoff error code.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1828,10 +1830,22 @@ func (b *csiErrorBackoff) blocking(id csiErrorBackoffId) bool {
 	return blk
 }
 
-func (b *csiErrorBackoff) next(id csiErrorBackoffId) {
+func (b *csiErrorBackoff) code(id csiErrorBackoffId) codes.Code {
+	if code, ok := b.errorCodes[id]; ok {
+		return code
+	}
+	// If we haven't recorded a code, return unavailable, which signals a problem with the driver
```
amacaskill, line 1837: My comment was that your  asssumption here is wrong. `next()` will always record whatever error code it is passed, which is the error code returned from CodeForError. CodeForError ALWAYS returns an error code, so we will never default to this case here where you return Unavailable. From your comment, it seems like you want to return Unavailable when the error is not a User error, so if you want to do that then, you should check if code from code := errorCodes[id], is Internal, and if it is return Unavailable. Or change your comment to say  we will never hit this case and remove the line that says " (ie, it's not filtered out as a user problem)."

pkg/gce-pd-csi-driver/controller.go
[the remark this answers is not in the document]
amacaskill: Changing CodeForErrror to return not a pointer helps, but I still have a comment about "I think the main thing is if the code is unset then returning Unavailble, which matches the existing backoff error code."

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1298/files#r1270954464

pkg/gce-pd-csi-driver/controller.go
mattcary: I was concerned with what to do if err is nil. That's not the case for the current use of next(), so it's kind of theoretical (just as a nil CodeForError was back when it was a pointer).

mattcary: oh, nvm, CodeForError is now reasonable if err is nil.

mattcary: Yes you're right, I've updated the comment

PR #1303, opened 2023-07-20
fix: volumeID is incorrent when project name contains alpha/beta/v1

### LINKED ISSUE #1302: VolumeID is incorrect when project name contains beta
We created a GKE cluster in our project named xxxxx-beta and tried to create a statefulset with persistentvolume.

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: web
spec:
  serviceName: "nginx"
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:stable
        ports:
        - containerPort: 80
        volumeMounts:
        - name: www
          mountPath: /usr/share/nginx/html
  volumeClaimTemplates:
  - metadata:
      name: www
    spec:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "standard-rwo"
      resources:
        requests:
          storage: 10Gi
```

The PVC and PV were created and bound as expected, but the volume could not be attached to the pod due to the following reason:

AttachVolume.Attach failed for volume "pvc-c970f876-7563-4ef8-ad77-e1aafc02ff69" : rpc error: code = InvalidArgument desc = ControllerPublishVolume volume ID is invalid: failed to get id components. Expected projects/{project}/zones/{zone}/disks/{name}. Got: zones/asia-southeast1-c/disks/pvc-c970f876-7563-4ef8-ad77-e1aafc02ff69

GKE version: 1.27.2-gke.1200
pd-csi version: v1.10.3

hoozecn:
**What type of PR is this?**
/kind bug

Fixes #1302 

**Does this PR introduce a user-facing change**:
```release-note
Fix: AttachVolume.Attach failed when name of gcp project ends with beta
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller_test.go
+++ b/pkg/gce-pd-csi-driver/controller_test.go
@@ -3304,6 +3304,16 @@ func TestCleanSelfLink(t *testing.T) {
 			in:   "https://www.partnerapis.com/compute/alpha/projects/project/zones/zone/disks/disk",
 			want: "projects/project/zones/zone/disks/disk",
 		},
+		{
+			name: "project name contains keyword",
+			in:   "https://www.partnerapis.com/compute/alpha/projects/proj-alpha/zones/zone/disks/disk",
+			want: "projects/proj-alpha/zones/zone/disks/disk",
+		},
+		{
+			name: "project name contains keyword 2",
+			in:   "https://www.partnerapis.com/compute/v1/projects/proj-alpha/zones/zone/disks/disk",
+			want: "projects/proj-alpha/zones/zone/disks/disk",
+		},
 	}
 
 	for _, tc := range testCases {
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -1745,7 +1745,7 @@ func generateCreateVolumeResponse(disk *gce.CloudDisk, zones []string, params co
 }
 
 func cleanSelfLink(selfLink string) string {
-	r, _ := regexp.Compile("https:\\/\\/www.*apis.com\\/.*(v1|beta|alpha)\\/")
+	r, _ := regexp.Compile(`https://www.*apis.com/[^/]+/(v1|beta|alpha)/`)
 	return r.ReplaceAllString(selfLink, "")
 }
 


mattcary:
/ok-to-test

Awesome great find. I'll take a look.

mattcary:
Can we use net/url to parse the url for real? That will save us the next oops-regexps-are-hard error.

I think log along the lines of the following should be right?

url =url.Parse(selfLink)
if url.Scheme != http return selflink
if url.Host !~ www.*apis.com return selflink
elts := strings.Split(url.path)
if len(elts) > 0 && elts[0] =~ vi|beta|alpha return strings.Join(elts[1:], "/")

If you don't want to make this change I'll do it.

Thanks again for finding this!

mattcary:
hmm, my pseudocode isn't quite right as the link looks like https://www.googleapis.com/compute/v1/projects/mattcary-gke-dev3/zones/us-central1-a/disks/pvc-6d2dd754-cd70-48c8-9944-ddfa40608585, so the "compute" needs to be removed as well.

Let me back up. Other places assume the volume id is projects/PPP/zones/ZZZ/disks/NNN (or perhaps projects/PPP/region/RRR/disks/NNN). Maybe it's best to extract that rather than trying to fiddle around with the prefix.

mattcary:
Hmm, after looking over the code, "cleanSelfLink" is just a really poor way to do what it's trying to do. Let me do a slightly larger refactor instead.

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -3304,6 +3304,16 @@ func TestCleanSelfLink(t *testing.T) {
 			in:   "https://www.partnerapis.com/compute/alpha/projects/project/zones/zone/disks/disk",
 			want: "projects/project/zones/zone/disks/disk",
 		},
+		{
+			name: "project name contains keyword",
```
sunnylovestiramisu, line 3308: Can we specify the key word in the test name instead of using 1 and 2? :)

sunnylovestiramisu:
We can probably refer to the existing regex validation from the GCE side for this string?

sunnylovestiramisu:
Sorry clicked close by mistake

mattcary:
I think that #1304 is preferable to this---it has a better chance of fixing the next error we haven't thought of yet :-)

It's still running through the integration tests, which confirm if my idea to parse very strictly is okay.

hoozecn:
cool. lgtm. closing this one. when would it could be applied to the gke? or is there a maual way to to patch the cluster that i already created. dont want to create another project frankly

> I think that #1304 is preferable to this---it has a better chance of fixing the next error we haven't thought of yet :-)
> 
> It's still running through the integration tests, which confirm if my idea to parse very strictly is okay.

/

mattcary:
We should be able to start pushing this out next week, it will take a couple weeks to go out. We should be able to backport to all affected versions. I'll update the timing here (but please ping if I forget).

PR #1304, opened 2023-07-21
Replace cleanSelfLink with getResourceId

### LINKED ISSUE #1302: VolumeID is incorrect when project name contains beta

mattcary:
/kind bug

Fixes #1302

```release-note
Fix resource parsing when the gcp project name ends with alpha, beta or v1
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller_test.go
+++ b/pkg/gce-pd-csi-driver/controller_test.go
@@ -930,7 +930,11 @@ func TestListVolumePagination(t *testing.T) {
 			var d []*gce.CloudDisk
 			for i := 0; i < tc.diskCount; i++ {
 				// Create diskCount dummy disks
-				d = append(d, gce.CloudDiskFromV1(&compute.Disk{Name: fmt.Sprintf("%v", i)}))
+				name := fmt.Sprintf("disk-%v", i)
+				d = append(d, gce.CloudDiskFromV1(&compute.Disk{
+					Name:     name,
+					SelfLink: fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/project/zones/zone/disk/%s", name),
+				}))
 			}
 			gceDriver := initGCEDriver(t, d)
 			tok := ""
@@ -988,7 +992,11 @@ func TestListVolumeArgs(t *testing.T) {
 			var d []*gce.CloudDisk
 			for i := 0; i < diskCount; i++ {
 				// Create 600 dummy disks
-				d = append(d, gce.CloudDiskFromV1(&compute.Disk{Name: fmt.Sprintf("%v", i)}))
+				name := fmt.Sprintf("disk-%v", i)
+				d = append(d, gce.CloudDiskFromV1(&compute.Disk{
+					Name:     name,
+					SelfLink: fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/project/zones/zone/disk/%s", name),
+				}))
 			}
 			gceDriver := initGCEDriver(t, d)
 			lvr := &csi.ListVolumesRequest{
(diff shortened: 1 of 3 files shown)


pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -988,7 +992,11 @@ func TestListVolumeArgs(t *testing.T) {
 			var d []*gce.CloudDisk
 			for i := 0; i < diskCount; i++ {
 				// Create 600 dummy disks
-				d = append(d, gce.CloudDiskFromV1(&compute.Disk{Name: fmt.Sprintf("%v", i)}))
+				name := fmt.Sprintf("disk-%v", i)
+				d = append(d, gce.CloudDiskFromV1(&compute.Disk{
+					Name:     name,
+					SelfLink: fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/project/zone/zone/disk/%s", name),
```
sunnylovestiramisu, line 998: Shouldn't it be projects/project/**zones**/zone ?

mattcary: ah, thanks. That's not causing the tests to fail though, since we don't validate that part of the resource (there are many options for that---regional, global, ...?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -3304,13 +3313,61 @@ func TestCleanSelfLink(t *testing.T) {
 			in:   "https://www.partnerapis.com/compute/alpha/projects/project/zones/zone/disks/disk",
 			want: "projects/project/zones/zone/disks/disk",
 		},
+		{
+			name: "alpha project",
+			in:   "https://www.googleapis.com/compute/alpha/projects/alphaproject/zones/zone/disks/disk",
```
judemars, line 3318: Nit: another nice test would be something like project name = betabeta

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -3304,13 +3313,61 @@ func TestCleanSelfLink(t *testing.T) {
...
+			in:    "https://www.googleapis.com/computers/v1/projects/project/zones/zone/disks/disk",
+			error: true,
+		},
+		{
+			name:  "bad host",
+			in:    "https://compute.googleapis.com/compute/v1/projects/project/zones/zone/disks/disk",
+			error: true,
+		},
+		{
+			name:  "bad version",
```
judemars, line 3352: Let's add a few more test cases:

content-compute.googleapis.com

compute.blahapis.com

content-compute.us-central1.rep.mtls.googleapis.com

pkg/gce-pd-csi-driver/controller_test.go
mattcary: done

pkg/gce-pd-csi-driver/controller_test.go
mattcary: done

judemars (approved):

mattcary:
/retest

Looking at history the job time has been ~18m, and this has also picked up the fsck change which may have increased the run time a bit. We may just need to increase the timeout on the test.

mattcary:
Manually created #1308 for release-1.9 cherry-pick.

1.8 is not affected.

PR #1318, opened 2023-07-31
[1.9]Added support for confidential storage to disk parameters and metric

Sneha-at:
Cherry picking #1291 to release-1.9

```release-note
Added support in PDCSI driver to create confidential hyperdisk storage on GCE.
```


mattcary:
Note that the vendor bump is unfortunately necessary to pick up gcp compute client updates for Hyperdisk.

/lgtm
/approve

PR #1240, opened 2023-05-26
Automated cherry pick of #1150 #1232 #1227 on release-1.9.

sunnylovestiramisu:
Cherry pick of #1150 #1232 #1227 on release-1.9.

#1150: satisfy volume cloning topology requirements when choosing zone for CreateVolume
#1232: Use errors.As so we can detect wrapped errors, and check for
#1227: Adding new metric pdcsi_operation_errors to fetch error

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note
#1150: satisfy volume cloning topology requirements when choosing zone for CreateVolume

#1232: Use errors.As so we can detect wrapped errors

#1227: Adding new metric pdcsi_operation_errors to fetch error
```

PR #1324, opened 2023-08-02
Automated cherry pick of #1230: filter out context errors

judemars:
Cherry pick of #1230 on release-1.9.

#1230: filter out context errors

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -20,9 +20,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
+	"strings"
+
+	"context"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -212,3 +217,66 @@ func containsZone(zones []string, zone string) bool {
...
+// codes.Internal if the given error is not a googleapi error caused by the user.
+// The following http error codes are considered user errors:
+// (1) http 400 Bad Request, returns grpc InvalidArgument,
+// (2) http 403 Forbidden, returns grpc PermissionDenied,
+// (3) http 404 Not Found, returns grpc NotFound
+// (4) http 429 Too Many Requests, returns grpc ResourceExhausted
+// The following errors are considered context errors:
+// (1) "context deadline exceeded", returns grpc DeadlineExceeded,
+// (2) "context canceled", returns grpc Canceled
+func CodeForError(err error) *codes.Code {
```
leiyiz, line 232: Looks like this bit of code was moved to `common/utils.go` in #1227 and I believe it's packported together with Matt's change all the way to 1.9. Have we checked if adding it here again is the right thing to do?

judemars: Thanks for pointing this out @leiyiz. After some digging I found #1240 actually effectively backports #1230, but not via the same commit. So we don't need this CL after all.

PR #1371, opened 2023-09-14
Upgrade google.golang.org/grpc from v1.55.0 -> v1.55.1 to address https://github.com/grpc/grpc-go/issues/6373

pwschuurman:
Upgrade google.golang.org/grpc from v1.55.0 -> v1.55.1 to address https://github.com/grpc/grpc-go/issues/6373

**What type of PR is this?**
/kind bug

Updates a grpc golang dependency patch version that had a bug which caused nil pointer exception.

```release-note
Fix bug caused by https://github.com/grpc/grpc-go/issues/6373 that results in driver binary crashes when parsing certain error grpc objects
```


pwschuurman:
Updated dependency with:

```
go get google.golang.org/grpc@v1.55.1
go mod vendor
```

go.sum
```diff
@@ -2259,6 +2259,8 @@ google.golang.org/grpc v1.40.0/go.mod h1:ogyxbiOoUXAkP+4+xa6PZSE9DZgIHtSpzjDTB9K
 google.golang.org/grpc v1.47.0/go.mod h1:vN9eftEi1UMyUsIF80+uQXhHjbXYbm0uXoFCACuMGWk=
 google.golang.org/grpc v1.55.0 h1:3Oj82/tFSCeUrRTg/5E/7d/W5A1tj6Ky1ABAuZuv5ag=
```
Sneha-at, line 2260: Are these expected to still be here?

pwschuurman: I think so. We have older versions of grpc as well (v1.47.0)

pwschuurman: Ah, sorry. Forgot to run `go mod tidy`

PR #1243, opened 2023-05-26
Automated cherry pick of #1079: Add provisionedIops for pd-extreme #1128: Change iops params directly convert string to int64

sunnylovestiramisu:
Cherry pick of #1150 #1079 #1101 #1128 on release-1.7.

#1150: satisfy volume cloning topology requirements when choosing zone for CreateVolume
#1079: Add provisionedIops for pd-extreme
#1128: Change iops params directly convert string to int64

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note
#1150: satisfy volume cloning topology requirements when choosing zone for CreateVolume
#1079: Add provisionedIops for pd-extreme
#1128: Change iops params directly convert string to int64

```

PR #1373, opened 2023-09-15
Upgrade google.golang.org/grpc from v1.55.0 -> v1.55.1 to address grpc/grpc-go#6373

pwschuurman:
Upgrade google.golang.org/grpc from v1.55.0 -> v1.55.1 to address https://github.com/grpc/grpc-go/issues/6373

**What type of PR is this?**
/kind bug

Updates a grpc golang dependency patch version that had a bug which caused nil pointer exception.

```release-note
Fix bug caused by https://github.com/grpc/grpc-go/issues/6373 that results in driver binary crashes when parsing certain error grpc objects
```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1146,6 +1146,8 @@ google.golang.org/grpc v1.40.0/go.mod h1:ogyxbiOoUXAkP+4+xa6PZSE9DZgIHtSpzjDTB9K
 google.golang.org/grpc v1.45.0/go.mod h1:lN7owxKUQEqMfSyQikvvk5tf/6zMPsrK+ONuO11+0rQ=
 google.golang.org/grpc v1.55.0 h1:3Oj82/tFSCeUrRTg/5E/7d/W5A1tj6Ky1ABAuZuv5ag=
 google.golang.org/grpc v1.55.0/go.mod h1:iYEXKGkEBhg1PjZQvoYEVPTDkHo1/bjTnfwTeGONTY8=
+google.golang.org/grpc v1.55.1 h1:36vzoa06ohIaveIgzr0qWDCImkKeVjvnNSV6uOmwnOw=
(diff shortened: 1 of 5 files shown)


pwschuurman:
Updated with:

```
go get google.golang.org/grpc@v1.55.1
go mod vendor
git restore vendor/google.golang.org/api/compute/v0.beta/compute-gen.go
```

Note: I reverted `vendor/google.golang.org/api/compute/v0.beta/compute-gen.go`, as I believe this was also done in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1243. Running `go mod vendor` on current `release-1.7` branch causes this file to be updated. Additionally `go mod tidy` fails, and has many other dependency updates associated with it. To reduce the risk of introducing further regressions, only the `grpc` package was updated.

go.sum
```diff
@@ -1146,6 +1146,8 @@ google.golang.org/grpc v1.40.0/go.mod h1:ogyxbiOoUXAkP+4+xa6PZSE9DZgIHtSpzjDTB9K
 google.golang.org/grpc v1.45.0/go.mod h1:lN7owxKUQEqMfSyQikvvk5tf/6zMPsrK+ONuO11+0rQ=
 google.golang.org/grpc v1.55.0 h1:3Oj82/tFSCeUrRTg/5E/7d/W5A1tj6Ky1ABAuZuv5ag=
 google.golang.org/grpc v1.55.0/go.mod h1:iYEXKGkEBhg1PjZQvoYEVPTDkHo1/bjTnfwTeGONTY8=
```
Sneha-at, line 1148: Should we leave this here as `go mod tidy` isn't working?

pwschuurman: Yeah, I think it's fine. The checksum entries are only used to validate that code for that package wasn't modifies. It will be cleaned up on the next successful run (once we upgrade other packages as well).

PR #1388, opened 2023-09-25
Remove buildplatform as arg from dockerfile

mattcary:
This causes errors in the release process.

/kind bug
```release-note
None
```


mattcary:
oops, this should be in release-1.8 and not the master branch.

PR #1462, opened 2023-10-20
Clear the secrets from request for klog print in `logGRPC()`

### LINKED ISSUE #1372: Driver can display the secret content of node-expand-secret-name
# What happened?

An user with permissions to create or modify `StorageClass` may print any `Secret`.

The problem comes from a [`klog` print](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-pd-csi-driver/utils.go#L63-L66) which is not sanitized:

```
	// Note that secrets are not included in any RPC message. In the past protosanitizer and other log
	// stripping was shown to cause a significant increase of CPU usage (see
	// https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/356#issuecomment-550529004).
	klog.V(4).Infof("%s called with request: %s", info.FullMethod, req)
```

The comment says that secrets are not included in any RPC message, however there many RPC messages that might include secrets. For example `NodeExpandVolume` as explained [here](https://kubernetes.io/blog/2022/09/21/kubernetes-1-25-use-secrets-while-expanding-csi-volumes-on-node-alpha/). If one can add:

```
parameters:
  csi.storage.k8s.io/node-expand-secret-name: test-secret
```

to the `StorageClass`, then after PV resize one can see in logs:

```
I0915 03:09:16.116829       1 utils.go:66] /csi.v1.Node/NodeExpandVolume called with request: volume_id:"projects/openshift-gce-devel/zones/us-central1-a/disks/pvc-11fd4016-7702-4c0c-9fae-71eddea3e4fa" volume_path:"/var/lib/kubelet/pods/2df06db8-b19a-4b7d-8aab-4d6eb65f0df0/volumes/kubernetes.io~csi/pvc-11fd4016-7702-4c0c-9fae-71eddea3e4fa/mount" capacity_range:<required_bytes:3221225472 > staging_target_path:"/var/lib/kubelet/plugins/kubernetes.io/csi/pd.csi.storage.gke.io/0b0a5be8b1e2d20b1e05a9f1f6d7b1ea36fcf6acfade1a9e448a088e20fe38bf/globalmount" volume_capability:<mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > > secrets:<key:"password" value:"t0p-Secret0" > secrets:<key:"username" value:"admin" > 
```

Note that there are 7 more RPC message types with secrets listed [here](https://kubernetes-csi.github.io/docs/secrets-and-credentials-storage-class.html).

GCP CSI driver does not seem to use these features, so the threat may come only from malicious user with elevated permissions who intentinally want to expose secrets.

# What did you expect to happen?

Log print to be sanitized. E.g.:

```
I0915 03:29:10.005165       1 utils.go:81] /csi.v1.Node/NodeExpandVolume called with request: {"capacity_range":{"required_bytes":4294967296},"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pd.csi.storage.gke.io/0b0a5be8b1e2d20b1e05a9f1f6d7b1ea36fcf6acfade1a9e448a088e20fe38bf/globalmount","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}},"volume_id":"projects/openshift-gce-devel/zones/us-central1-a/disks/pvc-11fd4016-7702-4c0c-9fae-71eddea3e4fa","volume_path":"/var/lib/kubelet/pods/2df06db8-b19a-4b7d-8aab-4d6eb65f0df0/volumes/kubernetes.io~csi/pvc-11fd4016-7702-4c0c-9fae-71eddea3e4fa/mount"}
```

# How can we reproduce it?

1. Create secret

```
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
  namespace: default
data:
stringData:
  username: admin
  password: t0p-Secret
```

2. Create sc

```
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: test5
parameters:
  csi.storage.k8s.io/node-expand-secret-name: test-secret
  csi.storage.k8s.io/node-expand-secret-namespace: default
reclaimPolicy: Retain
```

3. Create pvc/pod

4. Update csi driver log to Debug

5. Resize the volume capacity

6. Check csi driver logs

```
I0915 03:09:16.116829       1 utils.go:66] /csi.v1.Node/NodeExpandVolume called with request: volume_id:"projects/openshift-gce-devel/zones/us-central1-a/disks/pvc-11fd4016-7702-4c0c-9fae-71eddea3e4fa" volume_path:"/var/lib/kubelet/pods/2df06db8-b19a-4b7d-8aab-4d6eb65f0df0/volumes/kubernetes.io~csi/pvc-11fd4016-7702-4c0c-9fae-71eddea3e4fa/mount" capacity_range:<required_bytes:3221225472 > staging_target_path:"/var/lib/kubelet/plugins/kubernetes.io/csi/pd.csi.storage.gke.io/0b0a5be8b1e2d20b1e05a9f1f6d7b1ea36fcf6acfade1a9e448a088e20fe38bf/globalmount" volume_capability:<mount:<fs_type:"ext4" > access_mode:<mode:SINGLE_NODE_WRITER > > secrets:<key:"password" value:"t0p-Secret0" > secrets:<key:"username" value:"admin" > 
```

## Anything else we need to know?

n/a

/cc @msau42 , @mattcary , @cjcullen, @tallclair, @jsafrane

mpatlasov:
Malicious user can put a secret in request as explained here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1372.

**What type of PR is this?**

/kind bug

Malicious user can put a secret in request as explained here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1372.

This PR simply clears the secrets from request.

Fixes #1372

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/utils.go
+++ b/pkg/gce-pd-csi-driver/utils.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"google.golang.org/grpc"
@@ -59,14 +60,30 @@ func NewNodeServiceCapability(cap csi.NodeServiceCapability_RPC_Type) *csi.NodeS
 	}
 }
 
+// Reflect magic below simply clears Secrets map from request.
+func clearSecrets(req interface{}) interface{} {
+	v := reflect.ValueOf(&req).Elem()
+	e := reflect.New(v.Elem().Type()).Elem()
+	e.Set(v.Elem())
+	f := reflect.Indirect(e).FieldByName("Secrets")
+	if f.IsValid() && f.CanSet() && f.Kind() == reflect.Map {
+		f.Set(reflect.MakeMap(f.Type()))
+		v.Set(e)
+	}
+	return req
+}
+
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	if info.FullMethod == ProbeCSIFullMethod {
 		return handler(ctx, req)
 	}
-	// Note that secrets are not included in any RPC message. In the past protosanitizer and other log
+	// Note that secrets may be included in some RPC messages
+	// (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1372),
+	// but the driver ignores them. In the past protosanitizer and other log
 	// stripping was shown to cause a significant increase of CPU usage (see
 	// https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/356#issuecomment-550529004).
-	klog.V(4).Infof("%s called with request: %s", info.FullMethod, req)
+	// That is why we use hand-crafted clearSecrets() below rather than protosanitizer.
+	klog.V(4).Infof("%s called with request: %s", info.FullMethod, clearSecrets(req))
 	resp, err := handler(ctx, req)
 	if err != nil {
 		klog.Errorf("%s returned with error: %v", info.FullMethod, err.Error())
--- a/pkg/gce-pd-csi-driver/utils_test.go
+++ b/pkg/gce-pd-csi-driver/utils_test.go
@@ -18,11 +18,14 @@ limitations under the License.
 package gceGCEDriver
 
 import (
+	"context"
 	"fmt"
+	"reflect"
 	"testing"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 )
 
@@ -721,5 +724,55 @@ func TestValidateStoragePoolZones(t *testing.T) {
 				t.Errorf("%s: -want, +got \n%s", input, diff)
 			}
 		}
+	}
+}
+
+func TestClearSecrets(t *testing.T) {
+	vc := &csi.VolumeCapability{
+		AccessType: &csi.VolumeCapability_Mount{
+			Mount: &csi.VolumeCapability_MountVolume{},
+		},
+		AccessMode: &csi.VolumeCapability_AccessMode{
+			Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+		},
+	}
+
+	req := &csi.NodeExpandVolumeRequest{
+		VolumePath:       "/path",
+		VolumeCapability: vc,
+		Secrets: map[string]string{
+			"key": "value",
+		},
+	}
+
+	clearedReq := &csi.NodeExpandVolumeRequest{
+		VolumePath:       "/path",
+		VolumeCapability: vc,
+		Secrets:          map[string]string{},
+	}
+
+	if !reflect.DeepEqual(clearSecrets(req), clearedReq) {
+		t.Fatalf("Unexpected change: %v vs. %v", clearSecrets(req), clearedReq)
+	}
+}
+
+func TestLogGRPC(t *testing.T) {
+	info := &grpc.UnaryServerInfo{
+		FullMethod: "foo",
+	}
+
+	req := struct {
+		Secrets map[string]string
+	}{map[string]string{
+		"password": "pass",
+	}}
+
+	handler := func(ctx context.Context, req any) (any, error) {
+		return nil, nil
+	}
+
+	_, err := logGRPC(nil, req, info, handler)
+	if err != nil {
+		t.Fatalf("logGRPC returns error %v", err)
 	}
 }


pkg/gce-pd-csi-driver/utils.go
```diff
@@ -63,7 +64,12 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
 	// Note that secrets are not included in any RPC message. In the past protosanitizer and other log
 	// stripping was shown to cause a significant increase of CPU usage (see
 	// https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/356#issuecomment-550529004).
-	klog.V(4).Infof("%s called with request: %s", info.FullMethod, req)
+	// However malicious user still can put a secret in request as explained here:
+	// https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1372
+	if klog.V(4).Enabled() {
```
msau42, line 69: We regularly use log level 4 in production. Since this driver doesn't actually require any secret, can we just clear the secrets from the request?

mpatlasov: @msau42 , the request may be of different types, i.e. `NodeExpandVolumeRequest` is not the only type with `Secrets` (e.g. `CreateVolumeRequest`, `DeleteVolumeRequest`, etc.). I didn't find any straightforward way to clear the secrets from a request of arbitrary type, can you tell me how you suggest to do it?

msau42: Do you think we can leverage the reflect package? https://pkg.go.dev/reflect#example-Value.FieldByName potentially looks promising. Or could we even use https://pkg.go.dev/reflect#StructTag to create a more efficient protosanitizer?

mpatlasov:
Last push reworks PR: instead of calling `protosanitizer.StripSecrets(req)`, now it simply clears the secrets from request.

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -63,6 +64,17 @@ func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, h
...
+	// https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1372
+	// Reflect magic below simply clears Secrets map from request.
+	v := reflect.ValueOf(&req).Elem()
+	e := reflect.New(v.Elem().Type()).Elem()
+	e.Set(v.Elem())
+	f := reflect.Indirect(e).FieldByName("Secrets")
+	if f.IsValid() && f.CanSet() && f.Kind() == reflect.Map {
+		f.Set(reflect.MakeMap(f.Type()))
+		v.Set(e)
+	}
```
jsafrane, line 77: Please put it to a separate function.

mpatlasov:
Last push addressed Jan's [comment](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1462#discussion_r1379866109):

 * Reflection magic went to a separate function
 * Comment fixed
 * Unit-test added

mpatlasov:
@msau42 , can you `approve` please?

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -56,14 +57,30 @@ func NewNodeServiceCapability(cap csi.NodeServiceCapability_RPC_Type) *csi.NodeS
 	}
 }
 
+// Reflect magic below simply clears Secrets map from request.
+func clearSecrets(req interface{}) interface{} {
```
msau42, line 61: @jsafrane how do you feel about putting this into csi-lib-utils as an alternative to protosanitizer so that it is not specific to a csi driver?

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -56,14 +57,30 @@ func NewNodeServiceCapability(cap csi.NodeServiceCapability_RPC_Type) *csi.NodeS
 	}
 }
 
+// Reflect magic below simply clears Secrets map from request.
```
msau42, line 60: It would be good to document all the assumptions about how secret fields are named in the csi spec.

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -56,14 +57,30 @@ func NewNodeServiceCapability(cap csi.NodeServiceCapability_RPC_Type) *csi.NodeS
...
+}
+
 func logGRPC(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	if info.FullMethod == ProbeCSIFullMethod {
 		return handler(ctx, req)
 	}
-	// Note that secrets are not included in any RPC message. In the past protosanitizer and other log
+	// Note that secrets may be included in some RPC messages
+	// (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1372),
+	// but the driver ignores them. In the past protosanitizer and other log
```
msau42, line 79: Are you able to do a benchmark test to show that the reflect method does not signficantly increase CPU usage?

pkg/gce-pd-csi-driver/utils.go
jsafrane: Sounds good to me, but with a huge warning that only field explicitly named `Secrets` is cleared.

pkg/gce-pd-csi-driver/utils.go
mpatlasov: Hey @msau42 , I added a test to benchmark logGRPC. For 1M iterations, the reflect method increases execution time by only 1%:
```
# Before applying patch with `clearSecrets(req)`
$ go test -c -o /tmp/test ./pkg/gce-pd-csi-driver/ && time /tmp/test -test.run=TestLogGRPC -test.count=1000000
PASS
real	0m20.632s

# After applying patch with `clearSecrets(req)`
$ go test -c -o /tmp/test ./pkg/gce-pd-csi-driver/ && time /tmp/test -test.run=TestLogGRPC -test.count=1000000
PASS
real	0m20.750s
```

mpatlasov:
Hey @msau42 , can you lgtm/approve this, please. (see [my comment above](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1462#discussion_r1732930436))

mpatlasov:
Hey @msau42 , can you lgtm/approve this, please. (see my [comment above](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1462#discussion_r1732930436))

mpatlasov:
pull-gcp-compute-persistent-disk-csi-driver-verify is permafailing: https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-gcp-compute-persistent-disk-csi-driver-verify

pkg/gce-pd-csi-driver/utils.go
jsafrane: There were some improvements in csi-lib-utils recently to sanitize better / faster: https://github.com/kubernetes-csi/csi-lib-utils/pull/184

msau42: That's great! @jsafrane do the results there show a signficant difference in CPU/memory requirements compared to this approach?

mpatlasov: @msau42 , I opened [a WiP PR](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2083) to see how performance of logGRPC() changes if it calls modern protosanitizer.StripSecrets(). The result is encouraging: logGRPC() would become only 10% slower if we introduce protosanitizer.StripSecrets(req) back.

My clearSecrets() is much slower than modern protosanitizer.StripSecrets (for the same test iterating over logGRPC for 100,000,000 times as in [WiP PR](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2083)):
```
# logGRPC() calling protosanitizer.StripSecrets()
$ go test -c -o /tmp/test ./pkg/gce-pd-csi-driver
$ time /tmp/test -test.run=TestLogGRPC
...
real	1m7.355s

# logGRPC() calling clearSecrets()
$ go test -c -o /tmp/test ./pkg/gce-pd-csi-driver
$ time /tmp/test -test.run=TestLogGRPC
...
real	1m43.179s
```

I think we can safely discard that clearSecrets approach and simply revert back protosanitizer.StripSecrets(req). What do you think?

PR #1514, opened 2023-11-17
Add error logging when error is encountered fetching a VolumeSnapshotContent source image

### LINKED ISSUE #1513: Err when fetching VolumeSnapshotContent image from GCE API can result in nil-ptr dereference crash
In ListSnapshots, we [properly wrap an error](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/93d67cb28e5c7e8c64386ac24da2bd9419c01372/pkg/gce-pd-csi-driver/controller.go#L1380) from the GCE API when an error is encountered when fetching a snapshot source from the GCE API. For image snapshot source, this is not properly wrapped (only 404 errors are wrapped). This can result in a [nil-ptr dereference](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/93d67cb28e5c7e8c64386ac24da2bd9419c01372/pkg/gce-pd-csi-driver/controller.go#L1393C5-L1393C5), causing the driver to crash.

Narrowing down the list of errors for the image snapshot case (which does have a correct logging) indicates that this could fail for the following reasons. These are all valid and likely scenarios that should result in error propagation.
 * 403 Forbidden / Quota Exceeded
 * 503 Unavailable
 * 400 Invalid Argument

pwschuurman:
**What type of PR is this?**
/kind bug

Errors from the GCE images.get API call are not properly wrapped. This prevents error messages from being transparently displayed on a VolumeSnapshotContent, results in driver nil-ptr dereference and can result in Kubernetes sidecars to also restart.

Fixes #1513

```release-note
Properly wrap error from GCE Images.Get() API call, to fix a potential nil-ptr dereference
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/fake-gce.go
+++ b/pkg/gce-cloud-provider/compute/fake-gce.go
@@ -17,6 +17,8 @@ package gcecloudprovider
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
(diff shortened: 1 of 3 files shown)


mattcary:
/lgtm
/approve

Nice find

PR #1532, opened 2023-12-04
Add --fallback-requisite-zones command line argument

### LINKED ISSUE #1531: Regional PD Creation fails when Autopilot cluster has only 1 zone available
Repro Steps:

1. Create a GKE Autopilot Cluster
2. Scale to 0 or 1 nodes
3. Create a Regional Persistent Disk using a StorageClass without an `allowedTopology`

Autopilot cluster should fallback on the GKE cluster's default zones, which are configured on cluster creation.

pwschuurman:
**What type of PR is this?**
/kind bug

Fixes Regional Persistent Disk provisioning on GKE Autopilot clusters when there is only 0 or 1 zones available across nodes in the cluster.

Fixes #1531

This changes the behavior for Regional PD CreateVolume calls that have only one preferred topology. Previously, if a Regional PD was provisioned with a single preferredTopology (zone), and a list of requisite topologies (Zone), the secondary zone would be selected at random. This differed from the logic in [the external provisioner](https://github.com/kubernetes-csi/external-provisioner/blob/b02252374e59fb051258185805ed07fc97766d78/pkg/controller/topology.go#L253), which picks the secondary zone deterministically, based on the primary preferred topology. 

We don't expect this scenario to occur in practice, since the external provisioner will either pass in in at least two preferred topologies, or only one (and currently will fail, unless the `--fallback-requisite-zones` flag is provided). So it is safe to change this behavior as there should not be any real world scenarios relying on the random selection of the secondary zone that would be broken by a change to determinism.

```release-note
Add --fallback-requisite-zones flag to allow disk provisioning to fallback to a default set of zones when there are an insufficient number of zones available in a passed in requisite topology in CreateVolume.
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -65,6 +66,8 @@ var (
 
 	maxConcurrentFormatAndMount = flag.Int("max-concurrent-format-and-mount", 1, "If set then format and mount operations are serialized on each node. This is stronger than max-concurrent-format as it includes fsck and other mount operations")
 	formatAndMountTimeout       = flag.Duration("format-and-mount-timeout", 1*time.Minute, "The maximum duration of a format and mount operation before another such operation will be started. Used only if --serialize-format-and-mount")
+
+	fallbackRequisiteZonesFlag = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
 
 	version string
 )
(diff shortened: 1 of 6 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -66,6 +67,8 @@ var (
 	maxConcurrentFormatAndMount = flag.Int("max-concurrent-format-and-mount", 1, "If set then format and mount operations are serialized on each node. This is stronger than max-concurrent-format as it includes fsck and other mount operations")
 	formatAndMountTimeout       = flag.Duration("format-and-mount-timeout", 1*time.Minute, "The maximum duration of a format and mount operation before another such operation will be started. Used only if --serialize-format-and-mount")
 
+	fallbackRequisiteZonesFlag = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
```
amacaskill, line 70: Where is this going to be set? "fallback-requisite-zone" should be set to the GKE cluster's default zones, which are configured on cluster creation. Is this going to need some options.go logic so that when PDCSI is deployed on the cluster, we set the value based on how the cluster is configured?

amacaskill:
> /cc @amacaskill

This LGTM. All the test cases you added helps make it clear this still works with volume cloning. The only testcase I didn't see was regional -> regional volume cloning with --fallback-requisite-zones specified / used.

pwschuurman:
> > /cc Alexis MacAskill
> 
> This LGTM. All the test cases you added helps make it clear this still works with volume cloning. The only testcase I didn't see was regional -> regional volume cloning with --fallback-requisite-zones specified / used.

Thanks, added an extra test case for regional -> regional volume cloning.

PR #1534, opened 2023-12-08
add enable_storage_pools label to OperationErrorMetric

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature

> /kind flake

Adds enable_storage_pools label to the OperationErrorMetric so we can track CSI operation error rates when storage pools is used. 

Fixes #

```release-note
Adds enable_storage_pools label to the OperationErrorMetric so we can track CSI operation error rates when storage pools is used. 
```

### DIFF
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -20,6 +20,7 @@ package metrics
 import (
 	"testing"
 
+	computealpha "google.golang.org/api/compute/v0.alpha"
 	computebeta "google.golang.org/api/compute/v0.beta"
 	"google.golang.org/api/compute/v1"
 	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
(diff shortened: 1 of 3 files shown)


pkg/metrics/metrics_test.go
```diff
@@ -41,41 +42,69 @@ func CreateDiskWithConfidentialCompute(betaVersion bool, confidentialCompute boo
...
+		{
+			name:                              "test alphaDisk with storage pool projects/my-project/zone/us-central1-a/storagePools/sp1",
+			disk:                              CreateDiskWithStoragePool("projects/my-project/zone/us-central1-a/storagePools/sp1", hyperdiskBalanced),
+			expectedEnableConfidentialCompute: "false",
+			expectedDiskType:                  hyperdiskBalanced,
+			expectedEnableStoragePools:        "true",
+		},
+		{
+			name:                              "test alphaDisk with no storage pool",
+			disk:                              CreateDiskWithStoragePool("", hyperdiskBalanced),
```
sunnylovestiramisu, line 90: Can we add a test case that if the GetDisk call return an error(disk object will be empty), what is the expected output of this metrics?

amacaskill: Good call. Added.

PR #1553, opened 2023-12-22
[release-1.8] Update go modules

sunnylovestiramisu:
Updated the following go modules:
golang.org/x/crypto@v0.17.0
```release-note
NONE
```


msau42:
this branch doesn't use crypto module so maybe it doesn't need fixing?

PR #1558, opened 2024-01-02
change GetDisk error reporting to temporary in CreateVolume codepath

leiyiz:
**What type of PR is this?**

/kind bug

When we do CreateVolume, the csi provisioner expects either "final" or "temporary" errors, if we report final error when disk creation could be ongoing or the disk is already created, then a PVC deletion before successful volume creation could lead to disk leakage as DeleteVolume won't happen.

The temporary errors are defined here: https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/controller.go#L1904

This PR changes all the GCE API calls in `CreateVolume` code path that potentially happens *AFTER* `insertDisk` API calls happened to report a non-final error.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -560,7 +560,9 @@ func (cloud *CloudProvider) insertRegionalDisk(
 		if IsGCEError(err, "alreadyExists") {
 			disk, err := cloud.GetDisk(ctx, project, volKey, gceAPIVersion)
 			if err != nil {
-				return err
+				// failed to GetDisk, however the Disk may already exist
+				// the error code should be non-Final
+				return status.Error(codes.Unavailable, err.Error())
 			}
 			err = cloud.ValidateExistingDisk(ctx, disk, params,
 				int64(capacityRange.GetRequiredBytes()),
@@ -572,16 +574,19 @@ func (cloud *CloudProvider) insertRegionalDisk(
 			klog.Warningf("GCE PD %s already exists, reusing", volKey.Name)
 			return nil
 		}
-		return status.Error(codes.Internal, fmt.Sprintf("unknown Insert disk error: %v", err.Error()))
+		// if the error code is considered "final", RegionDisks.Insert might not be retried
+		return fmt.Errorf("unknown Insert Regional disk error: %w", err)
 	}
 	klog.V(5).Infof("InsertDisk operation %s for disk %s", opName, diskToCreate.Name)
 
 	err = cloud.waitForRegionalOp(ctx, project, opName, volKey.Region)
+	// failed to wait for Op to finish, however, the Op possibly is still running as expected
+	// the error code returned should be non-final
 	if err != nil {
 		if IsGCEError(err, "alreadyExists") {
 			disk, err := cloud.GetDisk(ctx, project, volKey, gceAPIVersion)
 			if err != nil {
-				return err
+				return status.Errorf(codes.Unavailable, "error when getting disk: %v", err.Error())
 			}
 			err = cloud.ValidateExistingDisk(ctx, disk, params,
 				int64(capacityRange.GetRequiredBytes()),
@@ -593,7 +598,7 @@ func (cloud *CloudProvider) insertRegionalDisk(
 			klog.Warningf("GCE PD %s already exists after wait, reusing", volKey.Name)
 			return nil
 		}
-		return fmt.Errorf("unknown Insert disk operation error: %w", err)
+		return status.Errorf(codes.Unavailable, "unknown error when polling the operation: %v", err.Error())
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -572,16 +574,18 @@ func (cloud *CloudProvider) insertRegionalDisk(
 			klog.Warningf("GCE PD %s already exists, reusing", volKey.Name)
 			return nil
 		}
-		return status.Error(codes.Internal, fmt.Sprintf("unknown Insert disk error: %v", err.Error()))
+		return fmt.Errorf("unknown Insert Regional disk error: %w", err)
```
mattcary, line 577: So the idea here is to pass whatever the wrapped err is, which should be correct w/rt to temporary or not, rather than converting to internal?

leiyiz: this is sort of unrelated to the main objective of this PR, I noticed that `insertZonalDisk` function has this same error thrown as `fmt.Errorf("unknown Insert Regional disk error: %w", err)` while this function does the manual code, I think it make sense to align the two and it seems more appropriate to do the former.

let me know what you think?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -558,9 +558,11 @@ func (cloud *CloudProvider) insertRegionalDisk(
 	}
 	if err != nil {
 		if IsGCEError(err, "alreadyExists") {
+			// failed to GetDisk, however the Disk may already exist
```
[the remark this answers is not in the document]
leiyiz: done.

leiyiz:
@mattcary addressed the comments, can you take another look

pkg/gce-cloud-provider/compute/gce-compute.go
mattcary: +1 to alignment

mattcary: and +1 to not changing the code to internal.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -695,7 +699,9 @@ func (cloud *CloudProvider) insertZonalDisk(
 		if IsGCEError(err, "alreadyExists") {
 			disk, err := cloud.GetDisk(ctx, project, volKey, gceAPIVersion)
 			if err != nil {
-				return err
```
mattcary, line 698: Above, errors like from SnapshotIDToProjectKey or a bad snapshot type will end up being a temporary error and we will retry. Do we want to be more accurate and fail permenantly?

pkg/gce-cloud-provider/compute/gce-compute.go
mattcary: Add a comment to the effect that this may or may not be retried depending on the error code.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -324,7 +324,8 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 	existingDisk, err := gceCS.CloudProvider.GetDisk(ctx, gceCS.CloudProvider.GetDefaultProject(), volKey, gceAPIVersion)
 	if err != nil {
 		if !gce.IsGCEError(err, "notFound") {
-			return nil, common.LoggedError("CreateVolume, failed to getDisk when validating: ", err)
+			// failed to GetDisk, however the Disk may already be created, the error code should be non-Final
+			return nil, common.LoggedError("CreateVolume, failed to getDisk when validating: ", status.Error(codes.Unavailable, err.Error()))
 		}
 	}
 	if err == nil {
```
mattcary, line 331: (I hate that you can't comment on lines outside the PR...)

On 345 there's a !ready check that returns an internal error code --- we should probably turn this into a temporary error?

You know, I think this is going to screw up our slo filtering :-/

mattcary: and on line 423

pkg/gce-cloud-provider/compute/gce-compute.go
leiyiz: per offline discussion, it's best to keep this as permanent error

pkg/gce-pd-csi-driver/controller.go
leiyiz: done

pkg/gce-cloud-provider/compute/gce-compute.go
leiyiz: done

pkg/gce-pd-csi-driver/controller.go
mattcary: (offline discussion: our slo filtering will be fine)

mattcary:
/lgtm
/approve

thanks!

PR #1559, opened 2024-01-02
Filter multiattach errors

mattcary:
/kind bug

User misconfiguration causing multiattach errors clouds up our SLO.

```release-note
Filter user misconfigured multiattach errors.
```

/assign @msau42

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -337,6 +337,9 @@ func CodeForError(sourceError error) codes.Code {
 		return codes.Internal
 	}
 
+	if code, err := isUserMultiAttachError(sourceError); err == nil {
+		return code
+	}
 	if code, err := existingErrorCode(sourceError); err == nil {
 		return code
 	}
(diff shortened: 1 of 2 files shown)


pkg/common/utils.go
```diff
@@ -374,6 +377,17 @@ func isContextError(err error) (codes.Code, error) {
 	return codes.Unknown, fmt.Errorf("Not a context error: %w", err)
 }
 
+// isUserMultiAttachError returns an InvalidArgument if the error is
+// multi-attach detected from the API server. If we get this error from the API
+// server, it means that the kubelet doesn't know about the multiattch so it is
```
msau42, line 382: I think it is possible there could be a race condition in K8s that also triggers this.

For example, with StatefulSet, the replacement Pod is created with the same name when the old Pod is deleted. Pod deletion is blocked on pod-volume unmounting, but not node-level unmount or detach. So a replacement Pod can be created before we have successfully detached.

mattcary: Yes, but in that cause the kubelet knows the volume is still attached and so the controller will figure out not to attach? https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/attachdetach/reconciler/reconciler.go#L341

mattcary: (I think that the only time I've seen this error from GCP is when the user has made two static PVs that refer to the same disk --- at least that's the case in the current SLOs that are firing).

msau42: Ah I see, in the race condition I am thinking of, ADC prevents the attach call from getting down to the CSI driver. So filtering the error at the CSI driver level is fine.

PR #1579, opened 2024-01-22
update klog version to latest. change %w unwrap to %v

leiyiz:
**What type of PR is this?**
/kind cleanup

in order to bump klog version, `%w` is no longer supported. `%v` should be used instead

Fixes #

```release-note
Error wrapping for `klog` logging is changed from `%w` to `%v` due to klog version bump.
```


leiyiz:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1575 updates klog but `%w` is no longer supported.

sunnylovestiramisu:
Does this worth make a note in release notes?

leiyiz:
> Does this worth make a note in release notes?

Yeah I think that makes sense

PR #1586, opened 2024-02-02
Update driver to support compute staging

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

The PR updates PDCSI driver to support GCE staging compute for internal testing.

Fixes #
NA

```release-note
NONE
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -17,7 +17,9 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
+	"fmt"
 	"math/rand"
 	"os"
 	"runtime"
@@ -67,13 +69,12 @@ var (
 
 	maxConcurrentFormatAndMount = flag.Int("max-concurrent-format-and-mount", 1, "If set then format and mount operations are serialized on each node. This is stronger than max-concurrent-format as it includes fsck and other mount operations")
 	formatAndMountTimeout       = flag.Duration("format-and-mount-timeout", 1*time.Minute, "The maximum duration of a format and mount operation before another such operation will be started. Used only if --serialize-format-and-mount")
-	computeEnvironment          = flag.String("compute-environment", "prod", "Sets the compute environment")
+	fallbackRequisiteZonesFlag  = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
 
-	fallbackRequisiteZonesFlag = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
-
-	enableStoragePoolsFlag = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
-
-	version string
+	enableStoragePoolsFlag                    = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
+	computeEnvironment        gce.Environment = "production"
+	version                   string
+	allowedComputeEnvironment = []string{"staging", "production"}
 )
 
 const (
@@ -86,6 +87,7 @@ func init() {
 	// Use V(4) for general debug information logging
 	// Use V(5) for GCE Cloud Provider Call informational logging
 	// Use V(6) for extra repeated/polling information
+	enumFlag(&computeEnvironment, "compute-environment", allowedComputeEnvironment, "Operating compute environment")
 	klog.InitFlags(flag.CommandLine)
 	flag.Set("logtostderr", "true")
 }
@@ -157,7 +159,7 @@ func handle() {
 	// Initialize requirements for the controller service
 	var controllerServer *driver.GCEControllerServer
 	if *runControllerService {
-		cloudProvider, err := gce.CreateCloudProvider(ctx, version, *cloudConfigFilePath, *computeEndpoint, *computeEnvironment)
+		cloudProvider, err := gce.CreateCloudProvider(ctx, version, *cloudConfigFilePath, *computeEndpoint, computeEnvironment)
 		if err != nil {
 			klog.Fatalf("Failed to get cloud provider: %v", err.Error())
 		}
@@ -205,4 +207,18 @@ func handle() {
 	gce.WaitForOpBackoff.Cap = *waitForOpBackoffCap
 
 	gceDriver.Run(*endpoint, *grpcLogCharCap, *enableOtelTracing)
+}
+
+func enumFlag(target *gce.Environment, name string, allowedComputeEnvironment []string, usage string) {
+	flag.Func(name, usage, func(flagValue string) error {
+		for _, allowedValue := range allowedComputeEnvironment {
+			if flagValue == allowedValue {
+				*target = gce.Environment(flagValue)
+				return nil
+			}
+		}
+		errMsg := fmt.Sprintf(`must be one of %v`, allowedComputeEnvironment)
+		return errors.New(errMsg)
+	})
+
 }
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"fmt"
 	"math/rand"
+	"net/url"
 	"os"
 	"runtime"
 	"strings"
@@ -40,7 +41,6 @@ import (
 var (
 	cloudConfigFilePath  = flag.String("cloud-config", "", "Path to GCE cloud provider config")
 	endpoint             = flag.String("endpoint", "unix:/tmp/csi.sock", "CSI endpoint")
-	computeEndpoint      = flag.String("compute-endpoint", "", "If set, used as the endpoint for the GCE API.")
 	runControllerService = flag.Bool("run-controller-service", true, "If set to false then the CSI driver does not activate its controller service (default: true)")
 	runNodeService       = flag.Bool("run-node-service", true, "If set to false then the CSI driver does not activate its node service (default: true)")
 	httpEndpoint         = flag.String("http-endpoint", "", "The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled.")
@@ -72,9 +72,10 @@ var (
 	fallbackRequisiteZonesFlag  = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
 
 	enableStoragePoolsFlag                    = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
-	computeEnvironment        gce.Environment = "production"
+	computeEnvironment        gce.Environment = gce.EnvironmentProduction
+	computeEndpoint           url.URL
 	version                   string
-	allowedComputeEnvironment = []string{"staging", "production"}
+	allowedComputeEnvironment = []gce.Environment{gce.EnvironmentStaging, gce.EnvironmentProduction}
 )
 
 const (
@@ -88,6 +89,7 @@ func init() {
 	// Use V(5) for GCE Cloud Provider Call informational logging
 	// Use V(6) for extra repeated/polling information
 	enumFlag(&computeEnvironment, "compute-environment", allowedComputeEnvironment, "Operating compute environment")
+	urlFlag(&computeEndpoint, "compute-endpoint", "Compute endpoint")
 	klog.InitFlags(flag.CommandLine)
 	flag.Set("logtostderr", "true")
 }
@@ -159,7 +161,7 @@ func handle() {
 	// Initialize requirements for the controller service
 	var controllerServer *driver.GCEControllerServer
 	if *runControllerService {
-		cloudProvider, err := gce.CreateCloudProvider(ctx, version, *cloudConfigFilePath, *computeEndpoint, computeEnvironment)
+		cloudProvider, err := gce.CreateCloudProvider(ctx, version, *cloudConfigFilePath, computeEndpoint, computeEnvironment)
 		if err != nil {
 			klog.Fatalf("Failed to get cloud provider: %v", err.Error())
 		}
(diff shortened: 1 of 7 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -159,7 +159,7 @@ const (
 )
 
 var (
-	validResourceApiVersions = map[string]bool{"v1": true, "alpha": true, "beta": true}
+	validResourceApiVersions = map[string]bool{"v1": true, "alpha": true, "beta": true, "staging_v1": true, "staging_beta": true, "staging_alpha": true}
```
sunnylovestiramisu, line 162: @mattcary Is it okay we put this on a public repo? It still requires permission to access the staging env.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +219,39 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
 	computeOpts := []option.ClientOption{option.WithHTTPClient(client)}
 	if computeEndpoint != "" {
-		v1Endpoint := fmt.Sprintf("%s/compute/v1/", computeEndpoint)
-		computeOpts = append(computeOpts, option.WithEndpoint(v1Endpoint))
-	}
-	service, err := compute.NewService(ctx, computeOpts...)
-	if err != nil {
-		return nil, err
+		endpoint := fmt.Sprintf("%s%s", computeEndpoint, computeEnvironmentSuffix)
+		klog.Infof("Got compute endpoint %s", endpoint)
```
sunnylovestiramisu, line 247: Nit, why do we need fmt.Sprintf and klog.Infof? klog.Infof is enough?

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +219,39 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
 	return service, nil
 }
 
-func createCloudService(ctx context.Context, vendorVersion string, tokenSource oauth2.TokenSource, computeEndpoint string) (*compute.Service, error) {
-	svc, err := createCloudServiceWithDefaultServiceAccount(ctx, vendorVersion, tokenSource, computeEndpoint)
-	return svc, err
-}
-
-func createCloudServiceWithDefaultServiceAccount(ctx context.Context, vendorVersion string, tokenSource oauth2.TokenSource, computeEndpoint string) (*compute.Service, error) {
+func getComputeVersion(ctx context.Context, tokenSource oauth2.TokenSource, computeEndpoint string, computeEnvironment string, computeVersion string) ([]option.ClientOption, error) {
```
sunnylovestiramisu, line 235: Can we add unit tests for this func as safe guard?

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -47,8 +48,33 @@ const (
...
 )
 
+type versionMap struct {
+	versionV1    *compute.Service
+	versionBeta  *computebeta.Service
+	versionAlpha *computealpha.Service
+}
+
+var computeVersionMap = map[string]map[string]map[string]string{
+	googleEnv: {
```
sunnylovestiramisu, line 64: Will this map work under other env?

pkg/gce-cloud-provider/compute/gce.go
Sneha-at: fmt.Sprintf is to save the endpoint in variable`endpoint`.

pkg/gce-cloud-provider/compute/gce.go
Sneha-at: Yes this will work with other environments too

pkg/gce-pd-csi-driver/controller.go
Sneha-at: We also have this in gce.go -> computeVersionMap. I can instead define a variable or something similar to be used at both places instead of hardcoded values. 
Note: They will still be hardcoded at variable initialization

pkg/gce-cloud-provider/compute/gce.go
sunnylovestiramisu: +1 endpoint := computeEndpoint+computeEnvironmentSuffix

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -47,8 +48,27 @@ const (
...
 
 	replicaZoneURITemplateSingleZone = "projects/%s/zones/%s" // {gce.projectID}/zones/{disk.Zone}
+	versionV1                        = "v1"
+	versionBeta                      = "beta"
+	versionAlpha                     = "alpha"
+	googleEnv                        = "googleapis"
 )
 
+var computeVersionMap = map[string]map[string]map[string]string{
+	googleEnv: {
```
pwschuurman, line 58: What's the purpose of keying this by "googleapis" if there is only one entry in the map? IMO let's just keep the prod/staging map at the top level, and add in the extra level later if needed.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +213,39 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
 	computeOpts := []option.ClientOption{option.WithHTTPClient(client)}
 	if computeEndpoint != "" {
-		v1Endpoint := fmt.Sprintf("%s/compute/v1/", computeEndpoint)
-		computeOpts = append(computeOpts, option.WithEndpoint(v1Endpoint))
-	}
-	service, err := compute.NewService(ctx, computeOpts...)
-	if err != nil {
-		return nil, err
+		endpoint := fmt.Sprintf("%s%s", computeEndpoint, computeEnvironmentSuffix)
+		klog.Infof("Got compute endpoint %s", endpoint)
```
pwschuurman, line 241: Don't we still need to construct the URI path in the event that the computeEndpoint is empty? I think this will result in the path not being appended in the default case where computeEndpoint flag is not provided. Please add a test to validate.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +213,39 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
-func createCloudServiceWithDefaultServiceAccount(ctx context.Context, vendorVersion string, tokenSource oauth2.TokenSource, computeEndpoint string) (*compute.Service, error) {
+func getComputeVersion(ctx context.Context, tokenSource oauth2.TokenSource, computeEndpoint string, computeEnvironment string, computeVersion string) ([]option.ClientOption, error) {
 	client, err := newOauthClient(ctx, tokenSource)
 	if err != nil {
 		return nil, err
 	}
-
+	computeEnvironmentSuffix, ok := computeVersionMap[googleEnv][computeEnvironment][computeVersion]
+	if !ok {
+		return nil, errors.New("Unable to fetch compute endpoint")
```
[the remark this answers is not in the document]
pwschuurman: Also, rather than having multimap of {env/version} -> path string, you could just build the path programatically. If you use types (instead of strings) for env/version, you can avoid having an error check, since the error state would be unrepresentable from your underlying data. Eg:

```
func getPath(env Environment, version Version) {
  prefix := ""
  if env == EnvironmentStaging {
    prefix = fmt.Sprintf("%s_", env)
  }
  return fmt.Sprintf("compute/%s%s/", prefix, version)
}
```

pkg/gce-cloud-provider/compute/gce.go
pwschuurman: Ah, misread. Yes, the path is already constructed by default if no compute endpoint is provided.

pkg/gce-pd-csi-driver/controller.go
Sneha-at: I have removed the other map in favor of building path programatically so we will need to keep this map as is

pkg/gce-cloud-provider/compute/gce.go
Sneha-at: Updated this logic to build path programmatically.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -47,8 +48,27 @@ const (
 	regionURITemplate = "projects/%s/regions/%s"
 
 	replicaZoneURITemplateSingleZone = "projects/%s/zones/%s" // {gce.projectID}/zones/{disk.Zone}
+	versionV1                        = "v1"
+	versionBeta                      = "beta"
+	versionAlpha                     = "alpha"
+	googleEnv                        = "googleapis"
 )
 
+var computeVersionMap = map[string]map[string]map[string]string{
```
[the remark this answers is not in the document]
Sneha-at: Updated all the occurrences to use custom types, note we are no longer using the map to build the path

pkg/gce-cloud-provider/compute/gce.go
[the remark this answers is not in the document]
Sneha-at: Updated the logic as per recommendation

pkg/gce-cloud-provider/compute/gce.go
Sneha-at: Done

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +201,48 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
 	computeOpts := []option.ClientOption{option.WithHTTPClient(client)}
+
 	if computeEndpoint != "" {
-		v1Endpoint := fmt.Sprintf("%s/compute/v1/", computeEndpoint)
-		computeOpts = append(computeOpts, option.WithEndpoint(v1Endpoint))
+		computeURL, err := url.ParseRequestURI(computeEndpoint)
+		if err != nil {
+			return nil, err
+		}
+		endpoint := computeURL.JoinPath(computeEnvironmentSuffix).String()
```
pwschuurman, line 230: Rather than JoinPath, I think we should ignore the path and just construct it ourselves. This will make the logic backwards compatible for anything that a calling process that runs the driver with `--compute-endpoint` string that includes a path.

```
computeURL.Path = computeEnvironmentSuffix
endpoint := computeURL.String()
```

You may need to update your `getPath()` function to include a leading `/` character when constructing the path.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +201,48 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
+		_, err = url.ParseRequestURI(endpoint)
```
pwschuurman, line 231: Isn't calling JoinPath guaranteed to produce a valid URL? I don't think that this second call to `ParseRequestURI` will ever fail. I think it can be removed.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +201,48 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
+		}
+		computeOpts = append(computeOpts, option.WithEndpoint(endpoint))
 	}
-	service, err := compute.NewService(ctx, computeOpts...)
-	if err != nil {
-		return nil, err
+	return computeOpts, nil
+}
+
+func getPath(env Environment, version Version) string {
```
pwschuurman, line 240: nit: I would call this constructComputeEndpointPath, rather than getPath. The name `getPath` is genericvery generic, and `get` implies we're either fetching a field from an object, or we're retrieving specific data from a higher level resource.

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -183,47 +201,48 @@ func createBetaCloudService(ctx context.Context, vendorVersion string, tokenSour
...
 		return nil, err
 	}
-
+	computeEnvironmentSuffix := getPath(computeEnvironment, computeVersion)
 	computeOpts := []option.ClientOption{option.WithHTTPClient(client)}
+
 	if computeEndpoint != "" {
-		v1Endpoint := fmt.Sprintf("%s/compute/v1/", computeEndpoint)
-		computeOpts = append(computeOpts, option.WithEndpoint(v1Endpoint))
+		computeURL, err := url.ParseRequestURI(computeEndpoint)
```
pwschuurman, line 226: The validation of computeEndpoint could be done at the initialization layer. You could do something very similar to your enumFlag() call, where you validate the computeEndpoint string set a computeURL value (pointer) with either the value, or nil if no computeURL is provided.

This allows you to avoid needing to check for `err` here, and your function on the construction of the URL and path.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1295,15 +1295,6 @@ var _ = Describe("GCE PD CSI Driver", func() {
 
 		klog.Infof("Creating new driver and client for node %s\n", i.GetName())
 
-		// Create new driver and client w/ invalid endpoint
-		tcInvalid, err := testutils.GCEClientAndDriverSetup(i, "invalid-string")
```
sunnylovestiramisu, line 1299: So we drop the invalid endpoint case?

pwschuurman: Yes, IMO this should be treated as input validation for the program rather than the RPC layer. Passing an invalid endpoint should crash the program, rather than cause a delayed RPC error.

cmd/gce-pd-csi-driver/main.go
```diff
@@ -67,12 +69,12 @@ var (
...
 
-	fallbackRequisiteZonesFlag = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
-
-	enableStoragePoolsFlag = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
-
-	version string
+	enableStoragePoolsFlag                    = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
+	computeEnvironment        gce.Environment = "production"
+	version                   string
+	allowedComputeEnvironment = []string{"staging", "production"}
```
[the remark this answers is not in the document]
Sneha-at: Done

pkg/gce-cloud-provider/compute/gce.go
Sneha-at: Done

pwschuurman:
@amacaskill Can you also cherrypick #1609 once that PR is merged? Something like:

```
~/go/src/k8s.io/kubernetes/hack/cherry_pick_pull.sh upstream/release-1.13 1586 1609
```

PR #1633, opened 2024-03-07
Fix nvme path filtering logic for udevadm trigger

pwschuurman:
**What type of PR is this?**
/kind bug

This PR correctly filters out partition paths in `/dev/nvme`. The search path was updated in PR#1485, however this introduced logic which filtered out all `/dev/nvme` paths. The result was that the logic to trigger a `udev` event for a NVMe device did not work as expected.

Fixes regression of search path https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1485/files

A follow-up PR will introduce an e2e test for a machine type that uses NVMe (current tests use SCSI), to prevent this type of regression in the future.

```release-note
Fixes a regression in the NVMe device search path
```

### DIFF
--- a/pkg/deviceutils/device-utils.go
+++ b/pkg/deviceutils/device-utils.go
@@ -36,9 +36,10 @@ const (
 	diskScsiGooglePrefix = "scsi-0Google_PersistentDisk_"
 	diskPartitionSuffix  = "-part"
 	diskSDPath           = "/dev/sd"
-	diskSDPattern        = "/dev/sd*"
+	diskSDGlob           = "/dev/sd*"
 	diskNvmePath         = "/dev/nvme"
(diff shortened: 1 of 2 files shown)


pkg/deviceutils/device-utils.go
```diff
@@ -39,6 +39,7 @@ const (
 	diskSDPattern        = "/dev/sd*"
 	diskNvmePath         = "/dev/nvme"
 	diskNvmePattern      = "^/dev/nvme[0-9]+n[0-9]+$"
+	devNvmePattern       = "/dev/nvme*"
```
mattcary, line 42: Can we use a different name to make it clear that it's a glob and not a regexp? eg devNvmeGlob

PR #1636, opened 2024-03-08
Add support for running tests on confidential VMs that use NVMe

pwschuurman:
**What type of PR is this?**
/kind cleanup

This adds support to run e2e tests in a configuration that uses NVMe devices. Currently the e2e tests default to using `n2` machine family, which uses SCSI by default. Passing in the `--enable-confidential-compute` flag is a cost effective way to use NVMe (an alternative being running against `c3-standard-4` machine type, but this requires additional test refactoring, or skipping tests, as there are restrictions on disk types).

To support toggling with just the binary flag, this also updates the defaults of the test to use `n2d` as well as a Ubuntu guest OS image instead of Debian.

Support for running "\[NVMe\]" tagged tests are presubmit will be added in https://github.com/kubernetes/test-infra/pull/32214. These can also be run locally with:

```
./test/run-e2e-local.sh --enable-confidential-compute
```

```release-note
NONE
```

### DIFF
--- a/test/e2e/utils/utils.go
+++ b/test/e2e/utils/utils.go
@@ -293,11 +293,13 @@ func ValidateLogicalLinkIsDisk(instance *remote.InstanceInfo, link, diskName str
 
 	devFsPath, err := instance.SSH("find", link, "-printf", "'%l'")
 	if err != nil {
-		return false, fmt.Errorf("failed to find symbolic link for %s. Output: %v, errror: %v", link, devFsPath, err.Error())
+		// Skip over if there is no matching symlink.
+		return false, nil
 	}
 	if len(devFsPath) == 0 {
 		return false, nil
 	}
+
 	if sdx := sdRegex.FindString(devFsPath); len(sdx) != 0 {
 		fullDevPath := path.Join("/dev/", string(sdx))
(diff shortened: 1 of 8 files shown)


test/e2e/utils/utils.go
```diff
@@ -275,11 +275,13 @@ func ValidateLogicalLinkIsDisk(instance *remote.InstanceInfo, link, diskName str
 
 	devFsPath, err := instance.SSH("find", link, "-printf", "'%l'")
 	if err != nil {
-		return false, fmt.Errorf("failed to find symbolic link for %s. Output: %v, errror: %v", link, devFsPath, err.Error())
```
Sneha-at, line 278: Why do we skip instead of erroring here?

pwschuurman: This function is called in a loop, based on pre-defined device paths. These device paths are generated here: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/e26e399e3d623b6c6b339201a1860159fcb9fbc7/pkg/deviceutils/device-utils.go#L116

If a device path that was generated isn't actually valid (eg: find fails to get a link), we should skip over the path, and move on to the next. So in that way, we don't actually want to fail here, we just want to return false to the caller, to indicate that the provided "link" is not a disk.

PR #1658, opened 2024-04-04
Add support for checking if a device is being used by a filesystem

### LINKED ISSUE #1657: NodeUnstage: Check if block device filesystem is in use
Currently the PDCSI driver performs an unmount on the globalmount during NodeUnstage, and success signals that ControllerUnpublish (GCE disk detach) can proceed. Normally successful mount is an indicator that there is no other consumer of the device on the OS. There are some sitautions where this may not be true:
1) Mount Namespace Capture: If a mount namespace is held open by a container process, due to an open file descriptor to `/proc/<pid>/ns/mnt`, a private mount point to the block device may exist in a container's mount namespace. Typically this only occurs if there is some other process that has root access to the OS and causes this to occur.
2) If there is some other user on the system which creates a private bind mount from the globalmount. The unmount and removal of the globalmount does not private to private bind mounts, resulting in a successful unmount in PDCSI, despite the device still being in use.

On linux, this can be resolved by checking if `/sys/fs/<filesystem>/<devX>` path exists. If the path exists, the device is still mounted and in use by the OS, and an error should be returned in NodeUnstage

pwschuurman:
**What type of PR is this?**
/kind bug

Summary: If a PD block device is still being used by the filesystem (misconfiguration, mount namespace capture), return an error during NodeUnstage.

Currently the PDCSI driver performs an unmount on the globalmount during NodeUnstage, and success signals that ControllerUnpublish (GCE disk detach) can proceed. Normally successful mount is an indicator that there is no other consumer of the device on the OS. There are some sitautions where this may not be true:
1) Mount Namespace Capture: If a mount namespace is held open by a container process, due to an open file descriptor to `/proc/<pid>/ns/mnt`, a private mount point to the block device may exist in a container's mount namespace. Typically this only occurs if there is some other process that has root access to the OS and causes this to occur.
2) If there is some other user on the system which creates a private bind mount from the globalmount. The unmount and removal of the globalmount does not private to private bind mounts, resulting in a successful unmount in PDCSI, despite the device still being in use.

Fixes #1657

```release-note
Check if volume's block device is still in use during NodeUnstage
```

### DIFF
--- a/pkg/deviceutils/device-utils.go
+++ b/pkg/deviceutils/device-utils.go
@@ -26,6 +26,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
+	"k8s.io/mount-utils"
 	pathutils "k8s.io/utils/path"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/resizefs"
 )
@@ -91,6 +92,13 @@ type DeviceUtils interface {
 
 	// Resize returns whether or not a device needs resizing.
 	Resize(resizer resizefs.Resizefs, devicePath string, deviceMountPath string) (bool, error)
+
+	// IsDeviceFilesystemInUse returns if a device path with the specified fstype
+	// TODO: Mounter is passed in in order to call GetDiskFormat()
+	// This is currently only implemented in mounter_linux, not mounter_windows.
+	// Refactor this interface and function call up the stack to the caller once it is
+	// implemented in mounter_windows.
+	IsDeviceFilesystemInUse(mounter *mount.SafeFormatAndMount, devicePath, devFsPath string) (bool, error)
 }
 
 type deviceUtils struct {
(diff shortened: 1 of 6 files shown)


mattcary:
/lgtm

Good idea to test in e2e. Probably we should have done that for disable device...

pwschuurman:
> @pwschuurman: #1658 failed to apply on top of branch "release-1.13":
> 
> ```
> Applying: Add support for checking if a device is being used by a filesystem
> Using index info to reconstruct a base tree...
> M	pkg/gce-pd-csi-driver/node.go
> M	test/e2e/tests/single_zone_e2e_test.go
> Falling back to patching base and 3-way merge...
> Auto-merging test/e2e/tests/single_zone_e2e_test.go
> Auto-merging pkg/gce-pd-csi-driver/node.go
> CONFLICT (content): Merge conflict in pkg/gce-pd-csi-driver/node.go
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> Patch failed at 0001 Add support for checking if a device is being used by a filesystem
> When you have resolved this problem, run "git am --continue".
> If you prefer to skip this patch, run "git am --skip" instead.
> To restore the original branch and stop patching, run "git am --abort".
> ```

Cherry pick #1805

pwschuurman:
Cherrypick for `release-1.12`: #1845

PR #1664, opened 2024-04-08
Record original error code to operation_errors metric for temporary errors 

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When CreateVolume is executed, the csi provisioner expects either "final" or "temporary" errors. If we report final error when disk creation could be ongoing or the disk is already created, then a PVC deletion before successful volume creation could lead to disk leakage as DeleteVolume won't happen.

To prevent disk leakage in PDCSI driver, we changed the error codes returned from GetDisk, InsertDisk, waitForZonalOp / waitForRegionalOp inside of CreateVolume to return the Unavailable / Aborted temporary error code  (see [PR](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1558)). The consequence of this most of the interesting errors we get from CreateVolume, all return Unavailable error code, which is filtered out from SLO. This PR reports the original error code (returned by GetDisk, InsertDisk, waitForZonalOp / waitForRegionalOp) to the operation_errors metric, but continues to return the "temporary" errors (Unavailable/Aborted) to the csi provisioner. 

Fixes #

```release-note
Record original error code for temporary errors for kubernetes.io/master/pdcsi/csidriver_operation_errors metric.
```

### DIFF
--- a/pkg/metrics/metrics_test.go
+++ b/pkg/metrics/metrics_test.go
@@ -18,9 +18,19 @@ limitations under the License.
 package metrics
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/metrics/metrics_test.go
```diff
@@ -115,3 +125,70 @@ func TestGetMetricParameters(t *testing.T) {
...
+		{
+			name:          "TemporaryError that wraps fmt.Errorf, which wraps googleapi error",
+			operationErr:  common.NewTemporaryError(codes.Aborted, fmt.Errorf("got error: %w", &googleapi.Error{Code: http.StatusBadRequest, Message: "User error with bad request"})),
+			wantErrorCode: "InvalidArgument",
+		},
+		{
+			name:          "TemporaryError that wraps status error",
+			operationErr:  common.NewTemporaryError(codes.Aborted, status.Error(codes.InvalidArgument, "User error with bad request")),
+			wantErrorCode: "InvalidArgument",
+		},
```
leiyiz, line 184: maybe add a case where temp error wraps the multi-attach error? is that a possible scenario?

leiyiz (commented):
/approve

pkg/metrics/metrics_test.go
amacaskill: Not totally sure? But I think its good to add.

amacaskill:
/hold cancel 
Added the test case.

pkg/metrics/metrics_test.go
amacaskill: Done

leiyiz (commented):
/lgtm

leiyiz (commented):
/lgtm

PR #1667, opened 2024-04-16
Remove support for GCE Alpha Disk Insert API

amacaskill:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This is a follow up PR to [PR 1666](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1666), which Migrated Hyperdisk/CHD/StoragePools features to the GCE v1 disk API. Now that no features are using the GCE alpha disk API, this PR removes support for Alpha Disks. It also adds v1 support to CloudDisk (/gce-cloud-provider/compute/cloud-disk.go) for the Hyperdisk/CHD/StoragePools features (d *CloudDisk) GetEnableConfidentialCompute()` 

Fixes #

```release-note
Remove support for GCE Alpha Disk Insert API.
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/cloud-disk.go
+++ b/pkg/gce-cloud-provider/compute/cloud-disk.go
@@ -18,15 +18,13 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	computealpha "google.golang.org/api/compute/v0.alpha"
 	computebeta "google.golang.org/api/compute/v0.beta"
 	computev1 "google.golang.org/api/compute/v1"
 )
 
 type CloudDisk struct {
-	disk      *computev1.Disk
-	betaDisk  *computebeta.Disk
-	alphaDisk *computealpha.Disk
+	disk     *computev1.Disk
+	betaDisk *computebeta.Disk
 }
 
 type CloudDiskType string
@@ -43,12 +41,6 @@ func CloudDiskFromBeta(disk *computebeta.Disk) *CloudDisk {
 	}
 }
 
-func CloudDiskFromAlpha(disk *computealpha.Disk) *CloudDisk {
-	return &CloudDisk{
-		alphaDisk: disk,
-	}
-}
-
 func (d *CloudDisk) LocationType() meta.KeyType {
 	var zone, region string
 	switch {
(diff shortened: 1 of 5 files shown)


pkg/gce-cloud-provider/compute/cloud-disk.go
```diff
@@ -134,8 +115,6 @@ func (d *CloudDisk) GetPDType() string {
 		pdType = d.disk.Type
 	case d.betaDisk != nil:
 		pdType = d.betaDisk.Type
-	case d.alphaDisk != nil:
```
sunnylovestiramisu, line 137: How will this impact the endpoint in TPC?

pkg/gce-cloud-provider/compute/cloud-disk.go
```diff
@@ -18,15 +18,13 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
-	computealpha "google.golang.org/api/compute/v0.alpha"
 	computebeta "google.golang.org/api/compute/v0.beta"
 	computev1 "google.golang.org/api/compute/v1"
 )
 
 type CloudDisk struct {
```
sunnylovestiramisu, line 25: Can we add a comment here like "if you want to add alpha API please refer to this PR" ?

amacaskill: I added this to cloud provider interface which is what stores the alpha service. I didn;t think we needed it in both places. What do you think?

```
// CloudProvider only supports GCE v1/beta Disk APIs. See
// https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1524
// for how to add GCE alpha Disk support.
```

sunnylovestiramisu: Oh I missed this, yeah one place is fine.

pkg/gce-cloud-provider/compute/cloud-disk.go
sunnylovestiramisu: I think just that in TPC it will also only use beta/v1 which should not be a problem

PR #1680, opened 2024-04-23
Fix CVE-2023-45288

dannawang0221:
**What type of PR is this?**
>

 /kind bug

Fixes #

```release-note
Fix CVE-2023-45288
```


vendor/modules.txt
```diff
@@ -885,7 +885,7 @@ k8s.io/kube-openapi/pkg/spec3
 k8s.io/kube-openapi/pkg/util/proto
 k8s.io/kube-openapi/pkg/validation/spec
 # k8s.io/mount-utils v0.31.0-alpha.0
-## explicit; go 1.22
+## explicit; go 1.22.0
```
sunnylovestiramisu, line 888: Can do you do go mod tidy and go mod vendor to clean this up?

vendor/modules.txt
```diff
@@ -885,7 +885,7 @@ k8s.io/kube-openapi/pkg/spec3
 k8s.io/kube-openapi/pkg/util/proto
 k8s.io/kube-openapi/pkg/validation/spec
 # k8s.io/mount-utils v0.31.0-alpha.0
```
sunnylovestiramisu, line 887: Hmm not related to this PR but this is not right, why are we depending on the alpha version.

dannawang0221: It comes from dependabot https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/commit/cde24e7aad20a314d1c83eeedb67debbc3f1e929 Seems like we should revert it?

vendor/modules.txt
dannawang0221: run `go mod tidy` will change go.mod back to 1.22.0

PR #1689, opened 2024-04-24
Remove device disable call

mattcary:
/kind bug

The device disable call has been failing because a newline needed to be added to the "offline" state written to sysfs. But in testing that, we've discovered disabling devices can cause problems. From a comment that I've added into this PR:

> Once a device is disabled, it's unusable, and can't be enabled unless the serial number is known. But the serial number cannot be read from the device as it's disabled. If a device is disabled in NodeUnstage, and then NodeStage is called without a NodeUnpublish & publish sequence, the disabled state of the device will cause NodeStage to fail. So this can only be used if we track the serial numbers of disabled devices in a persisent way that survives driver restarts.

```release-note
Remove disable device call on node unstage. This call was producing spurious error messages and was not effective.
```

@pwschuurman

### DIFF
--- a/pkg/deviceutils/device-utils_linux.go
+++ b/pkg/deviceutils/device-utils_linux.go
@@ -24,9 +24,16 @@ import (
 	"k8s.io/mount-utils"
 )
 
+// DisableDevice asks the kernel to disable a device via /sys.
+// NB: this can be dangerous to use. Once a device is disabled, it's unusable, and can't be enabled
+// unless the serial number is known. But the serial number cannot be read from the device as it's
+// disabled. If a device is disabled in NodeUnstage, and then NodeStage is called without a
+// NodeUnpublish & publish sequence, the disabled state of the device will cause NodeStage to fail.
+// So this can only be used if we track the serial numbers of disabled devices in a persisent way
+// that survives driver restarts.
 func (_ *deviceUtils) DisableDevice(devicePath string) error {
 	deviceName := filepath.Base(devicePath)
-	return os.WriteFile(fmt.Sprintf("/sys/block/%s/device/state", deviceName), []byte("offline"), 0644)
+	return os.WriteFile(fmt.Sprintf("/sys/block/%s/device/state", deviceName), []byte("offline\n"), 0644)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/deviceutils/device-utils_linux.go
```diff
@@ -24,9 +24,16 @@ import (
...
 )
 
+// DisableDevice asks the kernel to disable a device via /sys.
+// NB: this can be dangerous to use. Once a device is disabled, it's unusable, and can't be enabled
+// unless the serial number is known. But the serial number cannot be read from the device as it's
+// disabled. If a device is disabled in NodeUnstage, and then NodeStage is called without a
+// NodeUnpublish & publish sequence, the disabled state of the device will cause NodeStage to fail.
+// So this can only be used if we track the serial numbers of disabled devices in a persisent way
+// that survives driver restarts.
 func (_ *deviceUtils) DisableDevice(devicePath string) error {
```
pwschuurman, line 34: Why not just remove the entire function from deviceUtils interface? If we leave it in, this function isn't used and the code path isn't even being tested. If we really want to add it back in later, we can always look back at this PR revision to understand what was attempted before.

mattcary: I snuck in the newline fix. Can we submit this to get it in the repo, then I'll delete it in a followup CL? I also like the comment being attached to the method as it's hard to link up the PR to the code.

pwschuurman: Sure, SG

pkg/deviceutils/device-utils_linux.go
```diff
@@ -24,9 +24,16 @@ import (
 	"k8s.io/mount-utils"
 )
 
+// DisableDevice asks the kernel to disable a device via /sys.
+// NB: this can be dangerous to use. Once a device is disabled, it's unusable, and can't be enabled
+// unless the serial number is known. But the serial number cannot be read from the device as it's
+// disabled. If a device is disabled in NodeUnstage, and then NodeStage is called without a
+// NodeUnpublish & publish sequence, the disabled state of the device will cause NodeStage to fail.
+// So this can only be used if we track the serial numbers of disabled devices in a persisent way
```
pwschuurman, line 32: s/persisent/persistent

mattcary:
/retest

I up'd the timeout on the e2e tests, let's see if that fixed it.

PR #1695, opened 2024-05-02
Remove unused DisableDevice method

mattcary:
The DisableDevice call is no longer used. The comment around the device explains why so that if we ever decide to put it back in we'll have context.

/assign @pwschuurman 

/kind cleanup

```release-note
None
```

### DIFF
--- a/pkg/deviceutils/device-utils.go
+++ b/pkg/deviceutils/device-utils.go
@@ -86,10 +86,6 @@ type DeviceUtils interface {
 	// exists on the machine, or an empty string if none exists
 	VerifyDevicePath(devicePaths []string, deviceName string) (string, error)
 
-	// DisableDevice performs necessary disabling prior to a device being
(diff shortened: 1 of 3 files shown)


pwschuurman:
Can you also update the function names here to remove references to `disableDevice`: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/17e261efaa9f6d7be3ce4cbedc9e42cc0dddba2e/pkg/gce-pd-csi-driver/node.go#L454

I think the [warning message here](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/17e261efaa9f6d7be3ce4cbedc9e42cc0dddba2e/pkg/gce-pd-csi-driver/node.go#L457) should read something like:

```
Unable to check if device associated with volumeId %s is in use. Ignoring error and continuing with unstage: %v
```

mattcary:
Done, ptal

PR #1696, opened 2024-05-03
Use instances.list to determine PublishedNodeIds in ListVolume API

pwschuurman:
**What type of PR is this?**
  /kind bug
  
  The `disks.list` API is currently used in ListVolumes RPC. This looks up all disks within a project/region. The response includes a `items[].users[]` field, which contains the list of all VMs that are attached to a given disk. Thehe `items[].users[]` array has an implicit limit of 50 VMs. If a disk is attached to more than 50 instances, only 50 instances will be returned in `items[].users[]`. This causes problems with the [`external-attacher`](https://github.com/kubernetes-csi/external-attacher/blob/5f016e3d49b561d4e72375468684c6e19e56044a/pkg/controller/csi_handler.go#L134) reconciliation logic, since these disks appear to not be attached. When this occurs, the `external-attacher` will re-issue unnecessary ControllerPublishVolume calls for these disk/VM pairs. This causes additional RPC calls to the compute API (2x per ControllerPublish for `instances.get` and `disks.get`), and increases load on the PDCSI controller. The user facing effect is that GCP quota can be consumed unnecessarily.
  
The `instances.list` API has a `items[].disks[]` field which indicates all attached disks for an instance.  By moving to `instances.list` API to determine which VMs a disk is attached to, this gets around the 50 instance limit. Each VM can attach up to 128 disks (127 excluding the boot disk), so the `instances.list` API does not limit responses.
  
By making this new API call, the PDCSI controller will receive more data from the GCE API (from `instances.list` and `instances.get`). This has the potential to increase the latency of the ListVolume RPC call. This PR attempts to offset cost of this additional (paginated) API call by introducing two enhancements:
   *  This PR adds a new command line flag to filter disks that are returned in `instances.list` API call. This allows a driver to limit disks with particular features. For example, in Kubernetes, the driver should only be concerned about determining if disks are attached to VMs associated with the Kubernetes cluster the driver is running in. To limit total responses from `instances.list`, an administrator can pre-tag node VMs with a label (eg: the Kubernetes cluster name) and run the driver with the `--instances-list-filters` flag, targeting this label.
   * This PR limits the response fields returned from the GCE API for `disks.list` and `instances.list` API calls. Only a subset of the available response fields are actually used by the PDCSI controller client. This should reduce the payload response size, reducing the memory burst footprint of ListVolumes API calls (since partial results are cached in memory), and response latency of each paginated call, as less data needs to be transfered from the GCE server endpoint.
  
  ```release-note
The --instances-list-filters command line flag allows the driver to filter on response properties when listing VMs from the GCE API in ListVolumes RPC.
The --use-instance-api-to-list-volumes-published-nodes command line flag allows the driver to use the instances.list API instead of disks.list API when determining PublishedNodeIs in ListVolumes RPC.
  ```

**Testing**:

Validated existing behavior with:

```
./test/run-e2e-local.sh --enable-confidential-compute --use-instance-api-to-list-volumes-published-nodes=false
```

Validated new behavior:

```
./test/run-e2e-local.sh --enable-confidential-compute --use-instance-api-to-list-volumes-published-nodes
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -33,6 +33,7 @@ import (
 	"golang.org/x/oauth2"
 	computebeta "google.golang.org/api/compute/v0.beta"
 	computev1 "google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -106,7 +107,8 @@ type GCECompute interface {
 	GetDiskTypeURI(project string, volKey *meta.Key, diskType string) string
 	WaitForAttach(ctx context.Context, project string, volKey *meta.Key, diskType, instanceZone, instanceName string) error
 	ResizeDisk(ctx context.Context, project string, volKey *meta.Key, requestBytes int64) (int64, error)
-	ListDisks(ctx context.Context) ([]*computev1.Disk, string, error)
+	ListDisks(ctx context.Context, fields []googleapi.Field) ([]*computev1.Disk, string, error)
+	ListInstances(ctx context.Context, fields []googleapi.Field) ([]*computev1.Instance, string, error)
 	// Regional Disk Methods
 	GetReplicaZoneURI(project string, zone string) string
 	// Instance Methods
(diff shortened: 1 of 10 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -151,6 +153,7 @@ func (cloud *CloudProvider) ListDisks(ctx context.Context) ([]*computev1.Disk, s
 	nextPageToken := "pageToken"
 	for nextPageToken != "" {
 		rDiskList, err := rlCall.Do()
+		rlCall.Fields(fields...)
```
mattcary, line 156: Shouldn't this be before the Do()?

pwschuurman: Ah, yes. Updated

PR #1708, opened 2024-05-14
Properly unwrap gce-compute error code.

hime:
**What type of PR is this?**
>  
> /kind api-change  

/kind bug  

> /kind cleanup  
> /kind design  
> /kind documentation  
> /kind failing-test  
> /kind feature  
> /kind flake  

A new way of wrapping errors was introduced in [compute engine API](https://cloud.google.com/compute/docs/reference/rest/v1) (eg. "cloud.service.Disks.Insert") where some of these errors are wrapped with an additional layer of googleapi.Error. This additional wrap does not contain the http error code in its error code field, but instead contains a default value of Unknown. This is problematic because one of the ways we extract errors is by checking if the error can be a status (signifying a temporary error instead). The way the error is structured allows it to be classified as temporary, which causes our code to return the Unknown error present in the outermost layer, before any other error checks occur.

This has been causing issues in how we report errors coming from compute engine, since we do not currently unwrap them properly, resulting in `Unknown` error code being reported in most scenarios. This PR ensures we are unwrapping the embedded error code properly while making sure all the other error types we encounter continue to process errors as expected.

Fixes #

```release-note
Error codes extracted from errors part of compute engine api are now exposed with correct http errors instead of being classified as Unknown.

```

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
+	"github.com/googleapis/gax-go/v2/apierror"
 	"golang.org/x/time/rate"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/grpc/codes"
@@ -423,7 +424,6 @@ func CodeForError(sourceError error) codes.Code {
 	if sourceError == nil {
 		return codes.Internal
 	}
-
 	if code, err := isUserMultiAttachError(sourceError); err == nil {
 		return code
 	}
@@ -436,12 +436,7 @@ func CodeForError(sourceError error) codes.Code {
 	if code, err := isConnectionResetError(sourceError); err == nil {
 		return code
 	}
-
-	var apiErr *googleapi.Error
-	if !errors.As(sourceError, &apiErr) {
-		return codes.Internal
-	}
-	if code, ok := userErrorCodeMap[apiErr.Code]; ok {
+	if code, err := isGoogleAPIError(sourceError); err == nil {
 		return code
 	}
 
@@ -492,14 +487,62 @@ func isUserMultiAttachError(err error) (codes.Code, error) {
 	return codes.Unknown, fmt.Errorf("Not a user multiattach error: %w", err)
 }
 
+// existingErrorCode returns the existing gRPC Status error code for the given error, if one exists,
+// or an error if one doesn't exist. Since github.com/googleapis/gax-go/v2/apierror now wraps googleapi
+// errors (returned from GCE API calls), and sets their status error code to Unknown, we now have to
+// make sure we only return existing error codes from errors that are either TemporaryErrors, or errors
+// that do not wrap googleAPI errors. Otherwise, we will return Unknown for all GCE API calls that
+// return googleapi errors.
 func existingErrorCode(err error) (codes.Code, error) {
 	if err == nil {
 		return codes.Unknown, fmt.Errorf("null error")
 	}
-	if status, ok := status.FromError(err); ok {
-		return status.Code(), nil
+	var tmpError *TemporaryError
+	// This explicitly checks our error is a temporary error before extracting its
+	// status, as there can be other errors that can qualify as statusable
+	// while not necessarily being temporary.
+	if errors.As(err, &tmpError) {
+		if status, ok := status.FromError(err); ok {
+			return status.Code(), nil
+		}
 	}
+	// We want to make sure we catch other error types that are statusable.
+	// (eg. grpc-go/internal/status/status.go Error struct that wraps a status)
+	var googleErr *googleapi.Error
+	if !errors.As(err, &googleErr) {
+		if status, ok := status.FromError(err); ok {
+			return status.Code(), nil
+		}
+	}
+
 	return codes.Unknown, fmt.Errorf("no existing error code for %w", err)
+}
+
+// isGoogleAPIError returns the gRPC status code for the given googleapi error by mapping
+// the googleapi error's HTTP code to the corresponding gRPC error code. If the error is
+// wrapped in an APIError (github.com/googleapis/gax-go/v2/apierror), it maps the wrapped
+// googleAPI error's HTTP code to the corresponding gRPC error code. Returns an error if
+// the given error is not a googleapi error.
+func isGoogleAPIError(err error) (codes.Code, error) {
+	var googleErr *googleapi.Error
+	if !errors.As(err, &googleErr) {
+		return codes.Unknown, fmt.Errorf("error %w is not a googleapi.Error", err)
+	}
+	var sourceCode int
+	var apiErr *apierror.APIError
+	if errors.As(err, &apiErr) {
+		// When googleapi.Err is used as a wrapper, we return the error code of the wrapped contents.
+		sourceCode = apiErr.HTTPCode()
+	} else {
+		// Rely on error code in googleapi.Err when it is our primary error.
+		sourceCode = googleErr.Code
+	}
+	// Map API error code to user error code.
+	if code, ok := userErrorCodeMap[sourceCode]; ok {
+		return code, nil
+	}
+
+	return codes.Unknown, fmt.Errorf("googleapi.Error %w does not map to any known errors", err)
 }
 
 func LoggedError(msg string, err error) error {
--- a/pkg/common/utils_test.go
+++ b/pkg/common/utils_test.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"github.com/google/go-cmp/cmp"
+	"github.com/googleapis/gax-go/v2/apierror"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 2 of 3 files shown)


mattcary:
Please add a release not -- it will make backporting of this change a lot easier. Thanks!

pkg/common/utils_test.go
```diff
@@ -1220,6 +1228,14 @@ func TestCodeForError(t *testing.T) {
 			inputErr: &googleapi.Error{Code: http.StatusBadRequest, Message: "User error with bad request"},
 			expCode:  codes.InvalidArgument,
 		},
+		{
+			name: "googleapi.Error that wraps apierror.APIError, that wraps googleapi.Error",
```
amacaskill, line 1232: I thought the errors we receive are "github.com/googleapis/gax-go/v2/apierror" APIError, which wrap a googleAPI error? When will "googleapi.Error that wraps apierror.APIError, that wraps googleapi.Error" happen?

Can you add other test cases, for when api.APIError wraps other types of errors that aren't googleapi errors?

pkg/common/utils.go
```diff
@@ -479,12 +473,43 @@ func existingErrorCode(err error) (codes.Code, error) {
...
+		if status, ok := status.FromError(err); ok {
+			return status.Code(), nil
+		}
+	}
+	var googleErr *googleapi.Error
+	if !errors.As(err, &googleErr) {
+		if status, ok := status.FromError(err); ok {
+			return status.Code(), nil
+		}
 	}
```
amacaskill, line 487: Why do we need `!errors.As(err, &googleErr)`?

pkg/common/utils.go
```diff
@@ -479,12 +473,43 @@ func existingErrorCode(err error) (codes.Code, error) {
...
+
 	return codes.Unknown, fmt.Errorf("no existing error code for %w", err)
 }
 
+func isAPIError(err error) (codes.Code, error) {
+	var googleErr *googleapi.Error
+	if errors.As(err, &googleErr) {
```
amacaskill, line 494: nit: this entire function is in an if statement which checks that the error is a googleAPI error. We could do 
```
	var googleErr *googleapi.Error
	if !errors.As(err, &googleErr) {
		return codes.Internal
	}
```
first then return, to reduce nested if statments.

pkg/common/utils.go
```diff
@@ -479,12 +473,43 @@ func existingErrorCode(err error) (codes.Code, error) {
...
+		var sourceCode int
+		var apiErr *apierror.APIError
+		if errors.As(googleErr.Unwrap(), &apiErr) {
+			sourceCode = apiErr.HTTPCode()
+		} else {
+			sourceCode = googleErr.Code
```
amacaskill, line 500: Can you add comments to this? When will we hit each case.

pkg/common/utils.go
hime: We want to catch any other errors that qualify as status-able. This is because there is yae (yet another error) that we can get back from compute engine:

https://github.com/grpc/grpc-go/blob/f7d3d3eecbeecfdc61451426545dea92981666eb/internal/status/status.go#L174

There's no direct way to access this type because its under internal, so i thought this would be the best next thing.

pkg/common/utils_test.go
hime: You are right! We are not necessarily wrapping a third error here. The way the error gets created is by passing in googleAPI error to a function. I think we benefit for adding more test cases to deal with the different ways the APIError can be structured.

hime: Added more test cases.

pkg/common/utils.go
```diff
@@ -479,12 +474,49 @@ func existingErrorCode(err error) (codes.Code, error) {
 	if err == nil {
```
amacaskill, line 474: Can you add another comment explaining what existingErrorCode does as a whole? I still think its unclear why we 
need to do !errors.As(err, &googleErr)? I think this would be a good comment: 

existingErrorCode returns the existing gRPC Status error code for the given error, if one exists, or Unknown if one doesn't exist. Since github.com/googleapis/gax-go/v2/apierror now wraps googleapi errors (returned from GCE API calls), and sets their status error code to Unknown, we now have to make sure we only return existing error codes from errors that are either TemporaryErrors, or errors that do not wrap googleAPI errors. Otherwise, we will return Unknown for all GCE API calls that return googleapi errors.

pkg/common/utils.go
```diff
@@ -479,12 +474,49 @@ func existingErrorCode(err error) (codes.Code, error) {
...
+	var tmpError *TemporaryError
+	// This explicitly checks our error is a temporary error before extracting its
+	// status, as there can be other errors that can qualify as statusable
+	// while not necessarily being temporary.
+	if errors.As(err, &tmpError) {
+		if status, ok := status.FromError(err); ok {
+			return status.Code(), nil
+		}
+	}
+	// We want to make sure we catch other temporary error types that are statusable.
```
amacaskill, line 486: Nit: Remove the word temporary. I think this is easy to confuse with the temporary error above, and we don't mean temporary like TemporaryError here.

 We want to make sure we catch other temporary error types that are statusable. -> We want to make sure we catch other error types that are statusable.

pkg/common/utils.go
```diff
@@ -479,12 +474,49 @@ func existingErrorCode(err error) (codes.Code, error) {
...
+	if !errors.As(err, &googleErr) {
+		if status, ok := status.FromError(err); ok {
+			return status.Code(), nil
+		}
 	}
+
 	return codes.Unknown, fmt.Errorf("no existing error code for %w", err)
 }
 
+func isAPIError(err error) (codes.Code, error) {
```
amacaskill, line 498: I think isGoogleAPIError is probably a better name. 

Can you also add a function comment: 
isGoogleAPIError returns the gRPC status code for the given googleapi error by mapping the googleapi error's HTTP code to the corresponding gRPC error code.  If the error is wrapped in an APIError (github.com/googleapis/gax-go/v2/apierror), it maps the wrapped googleAPI error's HTTP code to the corresponding gRPC error code. Returns an error if the given error is not a googleapi error.

pkg/common/utils.go
```diff
@@ -479,12 +474,49 @@ func existingErrorCode(err error) (codes.Code, error) {
...
+	var googleErr *googleapi.Error
+	if !errors.As(err, &googleErr) {
+		return codes.Unknown, fmt.Errorf("error %w is not a googleapi.Error", err)
+	}
+	var sourceCode int
+	var apiErr *apierror.APIError
+	if errors.As(googleErr.Unwrap(), &apiErr) {
```
amacaskill, line 505: Do you need to do googleErr.Unwrap? Shouldn't this be  `errors.As(err, &apiErr) {` ?

apiError.HTTPCode() should give you the underlying googleapi HTTP error code, or -1 if it is not a googleapi error. 

```
// HTTPCode returns the underlying HTTP response status code. This method returns
// `-1` if the underlying error is a [google.golang.org/grpc/status.Status]. To
// check gRPC error codes use [google.golang.org/grpc/status.Code].
func (a *APIError) HTTPCode() int {
	if a.httpErr == nil {
		return -1
	}
	return a.httpErr.Code
}
```
I think

pkg/common/utils_test.go
```diff
@@ -1220,6 +1228,29 @@ func TestCodeForError(t *testing.T) {
...
+		{
+			name: "googleapi.Error that wraps apierror.APIError of http kind",
+			inputErr: getAPIWrappedError(&googleapi.Error{
+				Code:    404,
+				Message: "data requested not found error",
+			}, false /* wrap */),
+			expCode: codes.NotFound,
+		},
+		{
+			name: "googleapi.Error that wraps apierror.APIError of status kind",
```
amacaskill, line 1248: Can you also add a testcase for:
1. apierror.APIError wraps googleAPI error. 
2. apierror.APIError wraps status error. 

I know you have test cases for googleapi that wraps 1, and 2, but I think adding these without the outer wrapping of googleapi error is a good idea.

pkg/common/utils_test.go
```diff
@@ -1220,6 +1228,29 @@ func TestCodeForError(t *testing.T) {
...
+			expCode: codes.PermissionDenied,
+		},
+		{
+			name: "googleapi.Error that wraps apierror.APIError of http kind",
+			inputErr: getAPIWrappedError(&googleapi.Error{
+				Code:    404,
+				Message: "data requested not found error",
+			}, false /* wrap */),
+			expCode: codes.NotFound,
+		},
```
amacaskill, line 1246: Isn't googleapi.Error the same as http kind? If so, can we change http kind to googleapi.Error to be consistent. If they are the same, then I think this is the same test case as the one above.

pkg/common/utils.go
hime: I see that errors.As is able to take care of unwrapping case. Done.

pkg/common/utils.go
```diff
@@ -492,16 +487,64 @@ func isUserMultiAttachError(err error) (codes.Code, error) {
 	return codes.Unknown, fmt.Errorf("Not a user multiattach error: %w", err)
 }
 
+// existingErrorCode returns the existing gRPC Status error code for the given error, if one exists,
+// or Unknown if one doesn't exist. Since github.com/googleapis/gax-go/v2/apierror now wraps googleapi
```
amacaskill, line 491: Sorry one last nit. Can you please replace  "or Unknown if one doesn't exist." with "or an error if one doesn't exist". 

The error is the important thing. We ignore the Unknown code that is returned with the error.

PR #1717, opened 2024-05-20
update error messages for ControllerExpandVolume / CreateSnapshot of multi-zone PV

hungnguyen243:
update error messages for ControllerExpandVolume / CreateSnapshot of multi-zone PV.

Updated messages:
+ CreateSnapshot: "Snapshots are not supported with the `multi-zone` PV volumeHandle feature."
+ ControllerExpandVolume: "Resize operation is not supported with the `multi-zone` PVC volumeHandle feature. Please re-create the disk from source if you want a larger size."

### DIFF
--- a/test/e2e/tests/setup_e2e_test.go
+++ b/test/e2e/tests/setup_e2e_test.go
@@ -43,6 +43,7 @@ var (
 	imageURL        = flag.String("image-url", "projects/debian-cloud/global/images/family/debian-11", "OS image url to get image from")
 	runInProw       = flag.Bool("run-in-prow", false, "If true, use a Boskos loaned project and special CI service accounts and ssh keys")
 	deleteInstances = flag.Bool("delete-instances", false, "Delete the instances after tests run")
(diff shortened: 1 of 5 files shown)


Sneha-at:
/ok-to-test
@pwschuurman please LGTM.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1200,6 +1200,12 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 
 func (gceCS *GCEControllerServer) createPDSnapshot(ctx context.Context, project string, volKey *meta.Key, snapshotName string, snapshotParams common.SnapshotParameters) (*csi.Snapshot, error) {
 	volumeID, err := common.KeyToVolumeID(volKey, project)
+
+	volumeIsMultiZone := isMultiZoneVolKey(volKey)
```
Sneha-at, line 1204: Please move this after the volumeKey validation to line#1212. We would like to ensure volKey is valid before checking the multi-zone support.

PR #1718, opened 2024-05-21
Improve error messages for ControllerExpandVolume / CreateSnapshot of multi-zone PV.

hungnguyen243:
#### What type of PR is this?
/kind cleanup
#### What this PR does / why we need it:
Improve error messages for ControllerExpandVolume / CreateSnapshot of multi-zone PV.

Before:
CreateSnapshot: "Failed to create snapshot: failed to take snapshot of the volume projects/psch-gke-dev/zones/multi-zone/disks/hdml-llama2-70b-hf: "rpc error: code = Unknown desc = CreateSnapshot, failed 
to getDisk: googleapi: Error 400: Invalid value for field 'zone': 'multi-zone'. Unknown zone., invalid""
ControllerExpandVolume: "resize volume "my-disk-pv" by resizer "pd.csi.storage.gke.io" failed: rpc error: code = Unknown desc = ControllerExpandVolume failed to resize disk: failed to get disk: googleapi: Error 400: In
valid value for field 'zone': 'multi-zone'. Unknown zone., invalid"

After:
CreateSnapshot: "Snapshots are not supported with the `multi-zone` PV volumeHandle feature"
ControllerExpandVolume: "Resize operation is not supported with the `multi-zone` PVC volumeHandle feature. Please re-create the disk from source if you want a larger size."
#### Which issue(s) this PR fixes:
Fixes #

#### Does this PR introduce a user-facing change?
```release-note
Update logging on multi-zone feature support for volume snapshot and resize
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -1160,7 +1160,7 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 
 	volumeIsMultiZone := isMultiZoneVolKey(volKey)
 	if gceCS.multiZoneVolumeHandleConfig.Enable && volumeIsMultiZone {
-		return nil, fmt.Errorf("Snapshots are not supported with the `multi-zone` PV volumeHandle feature.")
+		return nil, status.Errorf(codes.InvalidArgument, "Snapshots are not supported with the multi-zone PV volumeHandle feature")
 	}
 
 	if acquired := gceCS.volumeLocks.TryAcquire(volumeID); !acquired {
@@ -1521,7 +1521,7 @@ func (gceCS *GCEControllerServer) ControllerExpandVolume(ctx context.Context, re
 
 	volumeIsMultiZone := isMultiZoneVolKey(volKey)
 	if gceCS.multiZoneVolumeHandleConfig.Enable && volumeIsMultiZone {
-		return nil, fmt.Errorf("Resize operation is not supported with the `multi-zone` PVC volumeHandle feature. Please re-create the disk from source if you want a larger size.")
+		return nil, status.Errorf(codes.InvalidArgument, "ControllerExpandVolume is not supported with the multi-zone PVC volumeHandle feature. Please re-create the volume %v from source if you want a larger size", volumeID)
 	}
 
 	sourceDisk, err := gceCS.CloudProvider.GetDisk(ctx, project, volKey, gce.GCEAPIVersionV1)
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -1160,7 +1160,7 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 
 	volumeIsMultiZone := isMultiZoneVolKey(volKey)
 	if gceCS.multiZoneVolumeHandleConfig.Enable && volumeIsMultiZone {
-		return nil, status.Errorf(codes.InvalidArgument, "Snapshots are not supported with the multi-zone PV volumeHandle feature")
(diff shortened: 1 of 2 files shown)


amacaskill:
I checked the e2e snapshot failure, the snapshot operation hadn't returned success before we tried to delete the disk, so that's why we got the operation error: https://screenshot.googleplex.com/7aHcpQdMhw4sBr5

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1158,6 +1158,11 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 		return nil, status.Errorf(codes.InvalidArgument, "CreateSnapshot Volume ID is invalid: %v", err.Error())
 	}
 
+	volumeIsMultiZone := isMultiZoneVolKey(volKey)
+	if gceCS.multiZoneVolumeHandleConfig.Enable && volumeIsMultiZone {
+		return nil, fmt.Errorf("Snapshots are not supported with the `multi-zone` PV volumeHandle feature.")
```
pwschuurman, line 1163: These should return a  proper status.Error (eg: InvalidArgument)

pwschuurman: I would also omit the backticks. This will be printed to the terminal, so it won't be markdown formatted

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1504,12 +1511,19 @@ func (gceCS *GCEControllerServer) ControllerExpandVolume(ctx context.Context, re
...
 	if err != nil {
 		if gce.IsGCENotFoundError(err) {
 			return nil, status.Errorf(codes.NotFound, "ControllerExpandVolume could not find volume with ID %v: %v", volumeID, err.Error())
 		}
 		return nil, common.LoggedError("ControllerExpandVolume error repairing underspecified volume key: ", err)
 	}
+
+	volumeIsMultiZone := isMultiZoneVolKey(volKey)
+	if gceCS.multiZoneVolumeHandleConfig.Enable && volumeIsMultiZone {
+		return nil, fmt.Errorf("Resize operation is not supported with the `multi-zone` PVC volumeHandle feature. Please re-create the disk from source if you want a larger size.")
```
pwschuurman, line 1524: I think naming this "ControllerExpandVolume" instead of "Resize operation" is actually better. It describes the RPC, as "Resize" is more of a Container Orchestrator term.

pwschuurman, line 1524: Use "volume" instead of "disk". Also, print out the volumeID in the error message

pwschuurman (commented):
Please add some tests that validate passing in a "multi-zone" volumeID returns the appropriate error (eg: InvalidArgument) for these two RPCs.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1158,6 +1158,11 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 		return nil, status.Errorf(codes.InvalidArgument, "CreateSnapshot Volume ID is invalid: %v", err.Error())
 	}
 
+	volumeIsMultiZone := isMultiZoneVolKey(volKey)
+	if gceCS.multiZoneVolumeHandleConfig.Enable && volumeIsMultiZone {
+		return nil, status.Errorf(codes.InvalidArgument, "Snapshots are not supported with the multi-zone PV volumeHandle feature")
```
pwschuurman, line 1163: Can you also emit the volumeID here?

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -255,6 +255,77 @@ func TestCreateSnapshotArguments(t *testing.T) {
 	}
 }
 
+func TestUnsupporteddMultiZoneCreateSnapshot(t *testing.T) {
```
pwschuurman, line 258: s/TestUnsupporteddMultiZoneCreateSnapshot/TestUnsupportedMultiZoneCreateSnapshot

PR #1733, opened 2024-06-04
Add support for "multi-zone" volume handle provisioning and deletion

pwschuurman:
**What type of PR is this?**
/kind feature

Add support for `multi-zone` volume handle provisioning. Volumes created with the `enable-multi-zone-provisioning: true` parameter will be created with multiple underlying disks, one per specified zone. The underlying volume returned will specify the keyword `multi-zone` in place of the `zone` field. See https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1616 for more details on the syntax.
 1) If there are zones specified in the requisite topologies, one disk will be created per zone
 2) If there are no zones specified in the requisite topologies, one disk will be created per zone in all zones in the driver's default region (where the particular disk type is supported).

Multi-zone volumes can be created from three sources:
1) PD images (ROX mode)
2) PD Snapshots (ROX mode)
3) Empty disks (RWO mode)

Deletion will inspect all disks that match the volume handle name, for all zones in the driver's default region.

This adds some additional e2e tests, that are not yet runnable, due to API support being in alpha. These have been run on a locally project that has alpha support, and can be replicated by running:

```
export RUN_MULTI_ZONE_TESTS=true
ginkgo --v "test/e2e/tests" -- --project "${PROJECT}" --service-account "${IAM_NAME}" --v=6 --logtostderr --ginkgo.focus "'multi-zone'" --extra-driver-flags=--set-access-mode-gce-api-version=alpha
```

```release-note
Add support for provisioning multi-zone volume handles
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -195,6 +195,8 @@ const (
 	resourceProject    = "projects"
 
 	listDisksUsersField = googleapi.Field("items/users")
+
+	readOnlyManyAccessMode = "READ_ONLY_MANY"
 )
 
 var (
@@ -214,7 +216,8 @@ var (
 		"items/selfLink",
 		"nextPageToken",
 	}
-	listDisksFieldsWithUsers = append(listDisksFieldsWithoutUsers, "items/users")
+	listDisksFieldsWithUsers      = append(listDisksFieldsWithoutUsers, "items/users")
+	disksWithModifiableAccessMode = []string{"hyperdisk-ml"}
 )
 
 func isDiskReady(disk *gce.CloudDisk) (bool, error) {
@@ -279,6 +282,15 @@ func useVolumeCloning(req *csi.CreateVolumeRequest) bool {
 }
 
 func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
+	response, err := gceCS.createVolumeInternal(ctx, req)
+	if err != nil && req != nil {
+		klog.V(4).Infof("CreateVolume succeeded for volume %v", req.Name)
+	}
+
+	return response, err
+}
+
+func (gceCS *GCEControllerServer) createVolumeInternal(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	var err error
 	diskTypeForMetric := metrics.DefaultDiskTypeForMetric
 	enableConfidentialCompute := metrics.DefaultEnableConfidentialCompute
@@ -288,17 +300,16 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
 	}()
 	// Validate arguments
 	volumeCapabilities := req.GetVolumeCapabilities()
-	name := req.GetName()
 	capacityRange := req.GetCapacityRange()
-	if len(name) == 0 {
+	if len(req.GetName()) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "CreateVolume Name must be provided")
 	}
 	if volumeCapabilities == nil || len(volumeCapabilities) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "CreateVolume Volume capabilities must be provided")
 	}
 
-	capBytes, err := getRequestCapacity(capacityRange)
-	if err != nil {
+	// Validate request capacity early
+	if _, err := getRequestCapacity(capacityRange); err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "CreateVolume Request Capacity is invalid: %v", err.Error())
 	}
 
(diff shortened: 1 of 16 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -331,11 +338,154 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
+			createDiskErrs = append(createDiskErrs, updateAccessModeErr)
+			continue
+		}
+
+		createdDisks = append(createdDisks, disk)
+	}
+	if len(createDiskErrs) == 1 {
+		return nil, common.LoggedError("Failed to create disk: ", createDiskErrs[0])
+	} else if len(createDiskErrs) > 1 {
+		// Return a non-final error here, since one of the errors could be a non-final.
```
mattcary, line 452: It might be a common case that len(createDiskErrs) == len(zones) and all the errors are the same, and final. eg some bad project configuration. Endlessly retrying that might be a problem?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -549,6 +709,83 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
...
+		err := gceCS.CloudProvider.DeleteDisk(ctx, project, zonalVolKey)
+		if err != nil {
+			deleteDiskErrs = append(deleteDiskErrs, gceCS.CloudProvider.DeleteDisk(ctx, project, volKey))
+		}
+	}
+
+	if len(deleteDiskErrs) == 1 {
+		return nil, common.LoggedError("Failed to delete disk: ", err)
+	} else if len(deleteDiskErrs) > 1 {
+		// Return a non-final error here, since one of the errors could be a non-final.
```
mattcary, line 770: In contrast, endlessly trying to delete disks is probably desired behavior.

pkg/gce-pd-csi-driver/controller.go
pwschuurman: Hmm, good call. Let me add some error merging logic here.

pwschuurman: Updated to check if the error is one of the retryable codes.

pkg/gce-pd-csi-driver/controller.go
pwschuurman: Updated to check if the error is one of the retryable codes.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -331,11 +338,149 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
+		volKey := meta.ZonalKey(req.GetName(), zone)
+		klog.V(4).Infof("Creating single zone disk for zone %q and volume: %v", zone, volKey)
+		disk, err := gceCS.createSingleDisk(ctx, req, params, volKey, []string{zone})
+		if err != nil {
+			createDiskErrs = append(createDiskErrs, err)
+			continue
+		}
+
+		readonly, _ := getReadOnlyFromCapabilities(req.GetVolumeCapabilities())
+		updateAccessModeErr := gceCS.updateAccessModeIfNecessary(ctx, volKey, disk, readonly)
```
msau42, line 441: Does this need to be set at creation time or is attach time sufficient? I'm wondering about the case where someone may have two pvs, one as rwo for updating and one rox.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -331,11 +338,149 @@ func (gceCS *GCEControllerServer) CreateVolume(ctx context.Context, req *csi.Cre
...
+		return nil, err
+	}
+	if acquired := gceCS.volumeLocks.TryAcquire(volumeID); !acquired {
+		return nil, status.Errorf(codes.Aborted, common.VolumeOperationAlreadyExistsFmt, volumeID)
+	}
+	defer gceCS.volumeLocks.Release(volumeID)
+
+	createDiskErrs := []error{}
+	createdDisks := make([]*gce.CloudDisk, 0, len(zones))
+	for _, zone := range zones {
```
msau42, line 431: Is it possible on a retry that the list of zones could change?

test/e2e/utils/utils.go
```diff
@@ -62,9 +63,10 @@ func GCEClientAndDriverSetup(instance *remote.InstanceInfo, driverConfig DriverC
 		fmt.Sprintf("--extra-labels=%s=%s", DiskLabelKey, DiskLabelValue),
 		"--max-concurrent-format-and-mount=20", // otherwise the serialization times out the e2e test.
 		"--multi-zone-volume-handle-enable",
-		"--multi-zone-volume-handle-disk-types=pd-standard",
+		"--multi-zone-volume-handle-disk-types=pd-standard,hyperdisk-ml",
```
msau42, line 66: I thought it was pd-ssd instead of pd-standard?

pkg/gce-pd-csi-driver/controller.go
pwschuurman: Yes, possible if support changes from the API side.

Right now, we're filtering down the user specified list of zones into the set of compatible zones. What do you think about using this as a check (and throwing an error if one of the provided zones is not supported?). This will cause the auto-scaling scenario and scale-from-zero scenarios to not work as well in regions where not all zones are supported for `hyperdisk-ml`. But we could also notify the user with an error message like:

`Disk type abc is not supported in zones [x y z]. If using dynamic topology awareness, please explicitly specify the set of zones`.

test/e2e/utils/utils.go
pwschuurman: This is just a testing configuration. I don't think anyone would want to enable --multi-zone-volume-handle-disk-types on `pd-standard`. I can update the tests?

pkg/gce-pd-csi-driver/controller.go
pwschuurman: Good catch, updated the accessMode to be set at creation time. When I was testing previously against the alpha, I didn't want to update the `disks.insert` call to use alpha (as it would affect other tests), so I just updated the accessMode after disk creation.

pkg/gce-pd-csi-driver/controller.go
msau42: I was thinking more from a k8s perspective. Maybe some nodes scaled down and that particular zone is not available anymore.

pwschuurman: Updated this logic to check for existing disks (with the same name/type, for all zones in the region) when determining the set of zones

PR #1736, opened 2024-06-05
Update docker file to get lvm configuration

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Updates the LVM configuration steps in the Dockerfile to support running in read only FS. To achieve this, it updates the entrypoint of the container.

Fixes #

```release-note
NONE 
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,8 +55,7 @@ COPY --from=debian /sbin/fsck /sbin/fsck
 COPY --from=debian /sbin/fsck* /sbin/
 COPY --from=debian /sbin/fsck.xfs /sbin/fsck.xfs
 # Add dependencies for LVM
-COPY --from=debian /etc/lvm /etc/lvm
-COPY --from=debian /etc/lvm* /etc/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -162,5 +156,5 @@ RUN /print-missing-deps.sh
 
 # Final build stage, create the real Docker image with ENTRYPOINT
 FROM output-image
-
-ENTRYPOINT ["/gce-pd-csi-driver"]
+COPY --from=builder /go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/initialize-driver.sh /initialize-driver.sh
+ENTRYPOINT ["/initialize-driver.sh"]
```
mattcary, line 160: This is fine for now. I think in the future we'll want to remove the entrypoint so that the deployment yaml has to be explicit about whether to use the initialize script or call the driver directly.

mattcary:
Force merging as there's some infra problem with the e2e tests.

PR #1749, opened 2024-06-13
Improve documentation of development project setup

mattcary:
/kind documentation

Improve documentation of developer setup, based on recent on-boarding of a new developer.

```release-note
None
```

/assign @pwschuurman 
/cc travisyx

### DIFF
--- a/deploy/common.sh
+++ b/deploy/common.sh
@@ -28,3 +28,30 @@ function ensure_kustomize()
   ensure_var PKGDIR
   "${PKGDIR}/deploy/kubernetes/install-kustomize.sh"
 }
+
+# This is a heuristic list of roles that appear to be necessary to run the full developer workflow.
+function check_dev_roles()
+{
+  local needed_roles="
+    roles/artifactregistry.admin
+    roles/compute.admin
+    roles/container.admin
+    roles/iam.roleAdmin
+    roles/iam.serviceAccountAdmin
+    roles/iam.serviceAccountKeyAdmin
+    roles/iam.serviceAccountUser
+    roles/resourcemanager.projectIamAdmin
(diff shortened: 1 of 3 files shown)


deploy/common.sh
```diff
@@ -28,3 +28,30 @@ function ensure_kustomize()
...
+    roles/iam.serviceAccountKeyAdmin
+    roles/iam.serviceAccountUser
+    roles/resourcemanager.projectIamAdmin
+    roles/serviceusage.serviceUsageAdmin
+  "
+  local have_roles=$(gcloud projects get-iam-policy "${PROJECT}")
+
+  missing_roles=
+  for r in $needed_roles; do
+    echo $have_roles | grep -q $r || missing_roles="$missing_roles $r"
```
pwschuurman, line 50: I think this will compare against **all** roles (even ones that aren't assigned to the current authenticated user).

I think a better approach may be to list the actual the current logged in user has, and then diff them with the set of required permissions

Something like:

```
for r in $needed_roles; do
  missing_permissions=$(comm --nocheck-order -13  <(gcloud projects get-iam-policy psch-gke-dev --flatten="bindings[].members" --filter="bindings.members=user:$(gcloud config list account --format 'value(core.account)')" --format="table(bindings.role)" | tail -n+2 | xargs -I {} gcloud iam roles describe {} --flatten="includedPermissions[]" --format="table(includedPermissions)" | tail -n+2) <(gcloud iam roles describe $r --flatten="includedPermissions[]" --format="table(includedPermissions)" | tail -n+2))
  if [ -n "$missing_permissions" ]; then
    ...
```

As much as like playing bash golf, I think this approach is pretty brittle, so I'd lean away from even introducing this check...

PR #1801, opened 2024-08-08
Add ControllerModifyVolume functionality

travisyx:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

**What this PR does / why we need it**

This PR implements ControllerModifyVolume. The code changes made allow for hyperdisks which support dynamic provisioning to update the IOPS and Throughput by patching the PVC with a different VolumeAttributesClass. For more information, see [here](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3751-volume-attributes-class).

**Which issue(s) this PR fixes**:

Fixes #1693

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
The ControllerModifyVolume functionality has been added to the CSI driver, allowing for dynamic provisioning of metadata for hyperdisks for an alpha cluster with the VolumeAttributesClass enabled.
```

PR #1814, opened 2024-08-28
Upgrade sanity tests to v5.3.0 and CSI Spec to v1.10.0

travisyx:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR upgrades sanity tests to v5.3.0, adding ControllerModifyVolume sanity tests. This PR also fixes unit tests that fail when comparing unexported fields in CSI structs.

Fixes #

I am creating this PR as a predecessor to [PR #1801](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1801). The ControllerModifyVolume functionality is added in that PR.

```release-note
NONE
```


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -108,6 +108,8 @@ type GCEControllerServer struct {
 	multiZoneVolumeHandleConfig MultiZoneVolumeHandleConfig
 
 	listVolumesConfig ListVolumesConfig
+
+	csi.UnimplementedControllerServer
```
amacaskill, line 112: Please just return Unimplemented for the CSI Spec functions that we are missing. t Example what we do in our other driver: https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/master/pkg/csi_driver/controller_unimpl.go#L26

Same for Node / identity service

travisyx: Resolved offline that it shouldn't be a pointer given that the methods are value methods. Added the comments to the three unimplemented fields

PR #1826, opened 2024-09-17
Add ControllerModifyVolume E2E tests

travisyx:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR adds ControllerModifyVolume E2E tests. This allows us to directly test the various APIs to check for any breaking changes in the future with the clients used.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/tests/single_zone_e2e_test.go
+++ b/test/e2e/tests/single_zone_e2e_test.go
@@ -17,8 +17,10 @@ package tests
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


test/run-e2e.sh
```diff
@@ -5,4 +5,4 @@ set -x
 
 readonly PKGDIR=sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 
-go test --timeout 30m --v "${PKGDIR}/test/e2e/tests" --run-in-prow=true --delete-instances=true --logtostderr $@
+go test --timeout 45m --v "${PKGDIR}/test/e2e/tests" --run-in-prow=true --delete-instances=true --logtostderr $@
```
amacaskill, line 8: previously these tests were taking 27 min ish to run, and now they are taking 44 min. Why do these tests make the tests run so much longer?

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1549,6 +1560,84 @@ var _ = Describe("GCE PD CSI Driver", func() {
 		Entry("with missing multi-zone label", multiZoneTestConfig{diskType: standardDiskType, readOnly: true, hasMultiZoneLabel: false, wantErrSubstring: "points to disk that is missing label \"goog-gke-multi-zone\""}),
 		Entry("with unsupported disk-type pd-extreme", multiZoneTestConfig{diskType: extremeDiskType, readOnly: true, hasMultiZoneLabel: true, wantErrSubstring: "points to disk with unsupported disk type"}),
 	)
+
+	DescribeTable("Should update metadata when providing valid metadata",
```
amacaskill, line 1564: If we cherry pick PRs back to previous release branches (before your change), is there a way to skip these test cases ? Otherwise I think these will fail against PRs merged to old release branches. In the test log, I see 
```
Ran 43 of 51 Specs in 1471.037 seconds
SUCCESS! -- 43 Passed | 0 Failed | 2 Pending | 6 Skipped
```

So I think there is a way to skip somewhere

test/run-e2e.sh
travisyx: The changes from ControllerModifyVolume take 5-6 minutes to be reflected. I added tests for both IOPS/throughput, just IOPS, and just throughput, causing the tests to take longer

test/e2e/tests/single_zone_e2e_test.go
travisyx: I changed the table to be pending, so this should skip for older versions. Is this sufficient?

test/run-e2e.sh
amacaskill: Do these e2e tests not run in parallel?

test/e2e/tests/single_zone_e2e_test.go
amacaskill: How does that make it skip for older versions? I see your tests now are pending even for this PR. I'm assuming that means they didn't run at all. What are the tests we are skipping now, and can we make those be skipped based on release branch?

travisyx: The other tests that were skipped are the tests in multi_zone_e2e_test.go. They get skipped if an environment variable isn't set. It doesn't seem like doing the same for these tests makes sense as they just wouldn't run on the tests for this PR. I set the tests to read the latest tag information and base the test skip on if the version is before v1.15.0.

amacaskill: Actually, based on a past new feature / new e2e test (https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/commit/70148a9b6a0f55837a6b2c1e153c43976dc9b9e0#diff-2547c5a5595d73ed4720c414c96d15d919e53abf7441f91f1b2c7250196a2606) I don't think this is an issue I guess?

travisyx: So basing test skips off an environment variable is fine? I guess that would make sense especially given that the tests were verified to be passing from an earlier PR run

test/run-e2e.sh
travisyx: No, they don't run in parallel by default

PR #1846, opened 2024-10-16
Add StatusConflict http kind to userErrorCodeMap.

halimsam:
**What type of PR is this?**
/kind bug

This PR accounts for http error code 409 (Status conflict) and inputting it into the user error code map as FailedPrecondition. This will be used to filter out unactionable SLO triggering alerts.
Fixes #

for internal bug.
no

```release-note
NONE
```
/assign mattcary

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -98,6 +98,7 @@ var (
 		http.StatusBadRequest:      codes.InvalidArgument,
 		http.StatusTooManyRequests: codes.ResourceExhausted,
 		http.StatusNotFound:        codes.NotFound,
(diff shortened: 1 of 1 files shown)


pkg/common/utils.go
```diff
@@ -98,6 +98,7 @@ var (
 		http.StatusBadRequest:      codes.InvalidArgument,
 		http.StatusTooManyRequests: codes.ResourceExhausted,
 		http.StatusNotFound:        codes.NotFound,
+		http.StatusConflict:        codes.FailedPrecondition,
```
amacaskill, line 101: Please add a unit test for this

PR #1849, opened 2024-10-23
feat: Provide hostname for zonal disk insert API

Rishita-Golla:
CL contains changes to pass in hostname for zonal disk.insert API as a part of location hint field. These changes are enabled behind a feature flag.

**What type of PR is this?**
> /kind feature

To improve performance when provisioning PDs, we can optimize their placement by using the hostname of the VM. In case of delayed binding, since the VM is already known, we can use this information to pass hostname as a parameter to the driver's create volume requests. For PD creation, the hostname can be included in location hint field in the [disks.insert](https://cloud.google.com/compute/docs/reference/rest/v1/disks/insert) call. This ensure the PD is created as close as possible to the VM, minimizing latency and cross-cell attachments.

Fixes #

Changes for regional disks will be part of future CLs.

```release-note
NONE
```


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -872,6 +878,24 @@ func (gceCS *GCEControllerServer) DeleteVolume(ctx context.Context, req *csi.Del
...
+	}
+
+	pvc, err := kubeClient.CoreV1().PersistentVolumeClaims(pvcNamespace).Get(ctx, pvcName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get persistent volume claims: %v", err)
+		// Don't log the error as create volume request can be continued without hostname
+		return ""
+	}
+	if hostName, ok := pvc.Annotations[annSelectedNode]; ok {
+		return hostName
```
mattcary, line 894: Add some logging when we found the selected node -- it will make it easier to confirm that things are working as expected.

pkg/gce-pd-csi-driver/controller_test.go
```diff
@@ -1596,12 +1596,12 @@ func (cloud *FakeCloudProviderInsertDiskErr) AddDiskForErr(volKey *meta.Key, err
 	cloud.insertDiskErrors[volKey.String()] = err
 }
```
mattcary, line 1598: It would be nice to add some sort of test that we can actually fetch the PVC correctly. Otherwise we're relying only on the k8s integration tests, and it will be harder to catch a regression.

I'll have to think on how to do that, though. The local api server testing setup I usually use will be hard to use in this context. Maybe you can poke around k/k to see if there's a lightweight mock?

Rishita-Golla: True! I wanted to discuss about this, hence marked the PR as draft. 
We could set up a mock client, but I'm not sure how useful/ how often it will be used in this repo. Will check in k/k if I find something.

pkg/gce-pd-csi-driver/controller.go
Rishita-Golla: Added.

pkg/gce-pd-csi-driver/controller_test.go
mattcary: I think something where we can confirm the expected PVC query happens, and the returned name is correctly plumbed into the gcp insert request, would be enough.

Rishita-Golla: I'm thinking using [fake package](https://pkg.go.dev/k8s.io/client-go/kubernetes/fake) is a good idea. We can create a client with PVC resource and checked returned hostname. Let me know your thoughts on this.

mattcary: Yup, I think that should do the trick.

Be sure to have tests using this, and also when it's nil (ie the hostname flag is not set)

Rishita-Golla: Changes are not tested in staging yet, updated the CL to include tests with fake-client and its imports.

PR #1853, opened 2024-10-30
return the correct path if paths are missing

kranurag7:
as of now we are always returning nvmeIdPath even if it exists. This commit fixes the same.

**What type of PR is this?**

/kind bug

Fixes #

```release-note

```


mattcary:
good catch, thank you.

/approve
/lgtm

PR #1857, opened 2024-11-04
switch to use snapshots.insert instead of disks.createsnapshot

leiyiz:
**What type of PR is this?**
/kind bug

If we use disks.CreateSnapshot, there could be a race condition where snapshot.Status is "READY" but the snapshot creation is not actually done. Snapshots.Insert API avoids this problem.

Fixes #

- snapshot.insert will try to grab resource lock too according to Xin.
- CMEK implicitly will be populated from PD if left blank, which keeps the previous behavior

```release-note
use `snapshots.insert` instead of `disks.createsnapshot` when creating pd snapshots.
pdcsi release including this change **MUST** upgrade csi-snapshotter sidecar to v7.0.0+.
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -1264,9 +1264,8 @@ func opIsDone(op *computev1.Operation) (bool, error) {
 
 func (cloud *CloudProvider) GetInstanceOrError(ctx context.Context, instanceZone, instanceName string) (*computev1.Instance, error) {
 	klog.V(5).Infof("Getting instance %v from zone %v", instanceName, instanceZone)
-	svc := cloud.service
 	project := cloud.project
-	instance, err := svc.Instances.Get(project, instanceZone, instanceName).Do()
+	instance, err := cloud.service.Instances.Get(project, instanceZone, instanceName).Do()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -1308,20 +1306,38 @@ func (cloud *CloudProvider) CreateSnapshot(ctx context.Context, project string,
...
+	}
 	switch volKey.Type() {
 	case meta.Zonal:
 		if description == "" {
 			description = "Snapshot created by GCE-PD CSI Driver"
 		}
-		return cloud.createZonalDiskSnapshot(ctx, project, volKey, snapshotName, snapshotParams, description)
 	case meta.Regional:
 		if description == "" {
 			description = "Regional Snapshot created by GCE-PD CSI Driver"
```
pwschuurman, line 1323: It doesn't look like you're setting description on the request. Also I'd invert the description empty string check, and the volKey Type check.

pwschuurman (commented):
/lgtm
/approve

pkg/gce-cloud-provider/compute/gce-compute.go
leiyiz: what do you mean by
> Also I'd invert the description empty string check, and the volKey Type check.

Thanks for the description catch.

PR #1864, opened 2024-11-07
Do not use beta API for hyperdisk in multi-writer mode.

### LINKED ISSUE #1863: Hyperdisk creation multi-writer (accesMode = ReadWriteMany) fails with `Invalid value for field 'resource.multiWriter': 'true'`
It was observed in https://issues.redhat.com/browse/STOR-2043 that dynamic provisioning of disk in multi-writer mode on C3 instance type with hyperdisk-balanced disk failed with below error:

```
Warning ProvisioningFailed 31m (x15 over 57m) pd.csi.storage.gke.io_wduan-1030a-g-vdnqh-master-0_dc46e9f6-1725-4f15-922c-99f0e281e102
failed to provision volume with StorageClass "balanced-storage": rpc error: code = InvalidArgument
desc = CreateVolume failed: rpc error: code = InvalidArgument desc = CreateVolume failed to create single zonal disk
pvc-0b65d680-636b-46c6-876d-d3a6c412c3ef: failed to insert zonal disk: unknown Insert disk error:
googleapi: Error 400: Invalid value for field 'resource.multiWriter': 'true'. Cannot specify the multi writer field for
'hyperdisk-balanced' disks, please use access mode instead., invalid%!v(MISSING)
```

I think the bug is [here](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L654C5-L654C16) the driver is still calling the beta API when multi writer is set to true.

[this](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L705) sets betaDiskToCreate.MultiWriter = multiWriter the boolean flag is only available in [v0.beta](https://pkg.go.dev/google.golang.org/api/compute/v0.beta) APIs and not in [v1](https://pkg.go.dev/google.golang.org/api/compute/v1) API.

The fix is to call v1 APIs which support accessMode = ReadWriteMany. Will create a PR.

karkunpavan:
**What type of PR is this?**
/kind bug

Hyper disk multi-writer support is now in [compute/v1](https://pkg.go.dev/google.golang.org/api/compute/v1) but the persistent disk csi driver still calls [v0.beta](https://pkg.go.dev/google.golang.org/api/compute/v0.beta) which leads a error while creating a hyperdisk with multi-writer mode. Error:

```
Warning ProvisioningFailed 31m (x15 over 57m) pd.csi.storage.gke.io_wduan-1030a-g-vdnqh-master-0_dc46e9f6-1725-4f15-922c-99f0e281e102 failed to provision volume with StorageClass
"balanced-storage": rpc error: code = InvalidArgument desc = CreateVolume failed: rpc error: code = InvalidArgument desc = CreateVolume failed to create single zonal disk pvc-0b65d680-636b-46c6-876d-d3a6c412c3ef: failed to insert zonal disk: unknown Insert disk error:
googleapi: Error 400: Invalid value for field 'resource.multiWriter': 'true'.
Cannot specify the multi writer field for 'hyperdisk-balanced' disks,
please use access mode instead., invalid%!v(MISSING)
```

To fix this we will call the v1 API which accepts `accessMode = ReadWriteMany`. 

Fixes #1863

- Discussed this issue and the fix with Matthew
- Making this change only for disk types `hyperdisk*`, multi-writer support for other disk types is not the focus of this fix.

```release-note
NONE
```

### DIFF
--- a/CHANGELOG/CHANGELOG-1.12.md
+++ b/CHANGELOG/CHANGELOG-1.12.md
@@ -1,3 +1,45 @@
+# v1.12.13 - Changelog since v1.12.11
+
+## Changes by Kind
+
+### Bug
+- Update Go version and dependencies to fix CVE-2024-24790,CVE-2024-24789 by @Sneha-at in #1851
+
+# v1.12.12 - Changelog since v1.12.11
+
+## Changes by Kind
+
+### Bug
+- Automated cherry pick of #1658: Add support for checking if a device is being used by a filesystem by @pwschuurman in #1845
+
+
+# v1.12.11 - Changelog since v1.12.10
+
+## Changes by Kind
+
+### Bug
+- [release-1.12] Update debian image from bullseye to bookworm to fix CVEs by @k8s-infra-cherrypick-robot in #1735
+- Reverting the Dockerfile debian image from bookworm to bullseye due to regression by @Sneha-at in #1774
+- [release-1.12] Return Unavailable for 'connection reset by peer' errors by @k8s-infra-cherrypick-robot in #1813
+- Automated cherry pick of #1708: Properly unwrap gce-compute error code. by @hime in #1840
+
+# v1.12.10 - Changelog since v1.12.9
+
+## Changes by Kind
+
+### Bug
+- Fix CVE-2023-45288 by @dannawang0221 in #1682
+
+
+# v1.12.9 - Changelog since v1.12.8
+
+## Changes by Kind
+
+### Bug
+
+- Change GetDisk error reporting to temporary in CreateVolume codepath ([#1600])https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1600), [@k8s-infra-cherrypick-robot](https://github.com/k8s-infra-cherrypick-robot))
+- [release-1.12] Fix nvme path filtering logic for udevadm trigger by @k8s-infra-cherrypick-robot in #1647
+
 # v1.12.8 - Changelog since v1.12.7
 
 ## Changes by Kind
--- a/CHANGELOG/CHANGELOG-1.13.md
+++ b/CHANGELOG/CHANGELOG-1.13.md
@@ -1,3 +1,46 @@
+# v1.13.8 - Changelog since v1.13.7
+
+## Changes by Kind
+
+### Uncategorized
+
+- [release-1.13] Properly unwrap gce-compute error code. by @hime in #1839
+
+
+# v1.13.7 - Changelog since v1.13.6
+
+## Changes by Kind
+
+### Uncategorized
+
+- [release-1.13] Reassign error returned from validateStoragePools so InvalidArgument is recorded by @k8s-infra-cherrypick-robot in #1721
+- [release-1.13] Return Unavailable for 'connection reset by peer' errors by @k8s-infra-cherrypick-robot in #1724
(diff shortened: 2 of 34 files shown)


karkunpavan:
Detailed code changes as requested. Updated the PR with all of these code changes.

1. In [insertZonalDisk()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L764) PD CSI driver currently calls beta APIs when multi-writer is set. We need to add a switch based on disk type. For hyperdisk use v1 API and for persistent disks use beta APIs.

1. Similar changes needed in [insertRegionalDisk()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L636C29-L636C47) 

1. AccessMode flag is already set [here](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L843) and no changes needed to explicitly set it for hyperdisks

1. [GetMultiWriter()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/cloud-disk.go#L216) currently thinks that v1 APIs do not support multi-writer and this needs to be fixed. We should return a True when AccessMode == READ_WRITE_MANY

1. Remove [comments](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-cloud-provider/compute/gce-compute.go#L556) about hyperdisk not supporting multi-writer

karkunpavan:
/retest-required 

The failures are SSH timeouts which seem to be unrelated to my changes.

gtomitsuka:
@mattcary @leiyiz any potential follow-up on why the test could be failing?

this is pretty important to my company, would let us move from `standard-rwx` with 1TB min capacity to something much more fine-grained. happy to support with anything

mattcary:
> @mattcary @leiyiz any potential follow-up on why the test could be failing?
> 
> this is pretty important to my company, would let us move from `standard-rwx` with 1TB min capacity to something much more fine-grained. happy to support with anything

This is not a replacement for standard-rwx. It enables multiwriter on a block device, but this does not mean you have a multiwriter filesystem. A distributed filesystem is hard to make, and multiwriter devices is only one piece of the puzzle. ext4 and xfs are not distributed.

gtomitsuka:
I understand, my case specifically is that we have Performance pods (= one pod per node) running ML models on GKE, and we'd like to enable rolling updates for it while keeping the models persistent to reduce overhead.

Right now, this rules us out from either doing rolling updates or using hyperdisk, due to a "true" distributed FS requirement, which has very high overhead.
Our understanding is that this is the k8s equivalent to using a cannon to catch a fish and in practice a much simpler solution would suffice, since writes are extremely rare and no more than one pod would be writing at any given time (we can also guarantee this using external locks if needed).

Am I understanding this incorrectly?

karkunpavan:
Closing this pull request, will raise a new one with planned changes

PR #1876, opened 2024-11-26
Refactor metric defer() statements to gRPC metric interceptor

pwschuurman:
Migrate metric defer() statements to gRPC metric interceptor. This allows for more accurate error code reporting of the internal `operation_errors` metric when gRPC handling functionality is refactored

> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Currently the ControllerServer reports error metrics through a `defer()` function pattern. This captures the output of captured variables when the function returns. If captured references are not updated (eg: a statement is refactored, and the `err` variable is not updated), the error code reported may not be accurate. This can result in poor error classification when this metric is used for reporting (as all error codes may be classified as `INTERNAL`).

```release-note
Improved field accuracy for `csidriver/operation_errors` metric.
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -21,7 +21,6 @@ import (
 	"math/rand"
 	neturl "net/url"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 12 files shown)


amacaskill:
@pwschuurman Can you please add a release note? 

The release note in the PR is how changes show up in the release notes/changelog that is generated when you press the generate release notes button when cutting a release. I think this is useful to know what version had what change, especially for SLO changes when we are debugging to see if whatever cluster triggered the alert, has your fix.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -332,10 +325,7 @@ func (gceCS *GCEControllerServer) createVolumeInternal(ctx context.Context, req
 	// Apply Parameters (case-insensitive). We leave validation of
 	// the values to the cloud provider.
 	params, err := gceCS.parameterProcessor().ExtractAndDefaultParameters(req.GetParameters(), gceCS.Driver.extraVolumeLabels, gceCS.Driver.extraTags)
-	diskTypeForMetric = params.DiskType
-	enableConfidentialCompute = strconv.FormatBool(params.EnableConfidentialCompute)
-	hasStoragePools := len(params.StoragePools) > 0
-	enableStoragePools = strconv.FormatBool(hasStoragePools)
+	metrics.UpdateRequestMetadataFromParams(ctx, params)
```
amacaskill, line 328: Why can't we put this as the first line in createVolumeInternal? 

We are returning InvalidArgument errors in several cases before we call this. I think that means this won't  be reported to the metric, correct? We don't really count those InvalidArgument requests to calculate any success rates, but this will make the PDCSI controller stackdriver logs, not match what the metric actually reports which sometimes makes it a little more confusing on debugging the alerts.

pwschuurman: SG, moved up the parameter extraction.

PR #1878, opened 2024-11-27
Update ./hack/verify-docker-deps.sh script to validate dependencies on build platform

pwschuurman:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
/kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Verify presubmit is currently failing (due to running on QEMU using target platform). This changes from using `ldd` on the target platform to using `readelf` on executables from the target image to validate dependencies exist in the base image.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1879, opened 2024-11-27
Remove make from verify-all in order to reduce presubmit flakes

mattcary:
/kind cleanup

```release-note
None
```

/assign @amacaskill

### DIFF
--- a/hack/verify-all.sh
+++ b/hack/verify-all.sh
@@ -23,5 +23,3 @@ PKG_ROOT=$(git rev-parse --show-toplevel)
 "${PKG_ROOT}/hack/verify-gofmt.sh"
 "${PKG_ROOT}/hack/verify-govet.sh"
 "${PKG_ROOT}/hack/verify-docker-deps.sh"
-
-make -C "${PKG_ROOT}" all


pwschuurman:
@mattcary Do you have an example of a verify presubmit flake where `make` is failing?

From what I've seen the verify presubmit started failing ~2 weeks ago, and times out running the verify-docker-deps command (running on ARM architecture with QEMU). I think #1878 should fix this problem.

mattcary:
> @mattcary Do you have an example of a verify presubmit flake where `make` is failing?
> 
> From what I've seen the verify presubmit started failing ~2 weeks ago, and times out running the verify-docker-deps command (running on ARM architecture with QEMU). I think #1878 should fix this problem.

verify-docker-deps is run during the docker bulid which is run as part of the make?

#1878 seems like a good idea, so we should do it too, but since we also build as part of the e2e I'm not sure what the value of running make as part of the verify is.

Having verify focus on the go code seems like a nice understandable and contained step to do. Fiddling with the build process seems like a different thing.

pwschuurman:
> Having verify focus on the go code seems like a nice understandable and contained step to do. Fiddling with the build process seems like a different thing.

Gotcha, agreed. Yeah, I think it's reasonable to remove.

mattcary:
Hmm, although with #1878 we will no longer verify dependencies as part of presubmit. That's a bug.

mattcary:
Ah, so actually I'm timing out in the verify docker deps which also runs make. So even more reason not to run make again?

mattcary:
I picked up Peter's fix and will retest.

/retest

PR #1896, opened 2024-12-20
Add validation for negative provisioned-throughput-on-create parameter

cemakd:
**What type of PR is this?**
/kind bug

If the StorageClass is of type hyperdisk-throughput and the parameter `provisioned-throughput-on-create` has a negative value, we don't include the negative value when calling the GCE API to provision the volume. Instead a sensible default is used injected by GCE API, leading to a confusing UX.

Fixes #

```release-note
Using a negative value for provisioned-throughput-on-create parameter of hyperdisk-throughput type StorageClass will now lead to a failure when creating a new volume.
```


leiyiz (commented):
/lgtm
/approve

PR #1897, opened 2025-01-02
Use correct path in error message for udev tooling

mattcary:
/kind bug

The current error message uses the wrong path. #1853 would have fixed it, but it was not submitted for some reason.

```release-note
Use correct tool path in udev error message.
```

/assign @msau42


mattcary:
/retest

Infra flake

PR #1899, opened 2025-01-02
[WIP] Fix Hyperdisk ControllerExpandVolume Edge Cases

sunnylovestiramisu:
**What type of PR is this?**
/kind bug

A [hyperdisk](https://cloud.google.com/compute/docs/disks/hyperdisks) with size 4Gi needs exactly 2000 IOPS, and 5Gi needs exactly 2500 IOPS. Disks 6Gi and more need a minimum of 3000 IOPS. When resizing a hyperdisk, there is a need to update IOPS as well for some corner cases. Currently ControllerExpandVolume is failing for these edge cases.

The GCE disk update API does not support resizing async PD today. Even currently the PDCSI driver does not support resizing sync PDs, to make it a two way door, the proposal is:

- Use GCE **Resize** API for **PD** disks ControllerVolumeExpand
- Use GCE **Update** API for **Hyperdisks** ControllerVolumeExpand

Testing:

1. Use kubetest clusters([machine type support for Hyperdisk](https://cloud.google.com/compute/docs/disks/hyperdisks#machine-type-support)) ```NUM_NODES=1 NODE_SIZE="n4-standard-4" MASTER_SIZE="n4-standard-4"  NODE_DISK_TYPE="hyperdisk-balanced" kubetest --up``` 
2. Deploy PDCSI driver with this change
3. Create PVC
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-demo
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 4Gi
  storageClassName: hyperdisk-balanced-sc
---
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: hyperdisk-balanced-sc
parameters:
  type: hyperdisk-balanced
  provisioned-iops-on-create: "2000"
  provisioned-throughput-on-create: "200Mi"
allowVolumeExpansion: true
---
kind: Pod
apiVersion: v1
metadata:
  name: pod-demo
spec:
  volumes:
    - name: pvc-demo-vol
      persistentVolumeClaim:
       claimName: pvc-demo
  containers:
    - name: pod-demo
      image: nginx
      resources:
        limits:
          cpu: 10m
          memory: 80Mi
        requests:
          cpu: 10m
          memory: 80Mi
      ports:
        - containerPort: 80
          name: "http-server"
      volumeMounts:
        - mountPath: "/usr/share/nginx/html"
          name: pvc-demo-vol
```

Fixes #

```release-note
Fix Hyperdisk ControllerExpandVolume Edge Cases for 4,5 and 6 Gi Resize
```


mattcary:
Can you expand on the release notes with the specific sizes (the 4, 5 and 6 G thresholds that you mention above)? It's handy to have those specifics in the release notes that go out with the new version.

pkg/common/parameters.go
```diff
@@ -134,6 +134,7 @@ type ParameterProcessor struct {
 type ModifyVolumeParameters struct {
 	IOPS       *int64
 	Throughput *int64
+	SizeGb     *int64
```
mattcary, line 137: As discussed offline, doing a size change in modify volume will cause problems because there's no node rpc to update the filesystem?

mattcary: Oh, I see, you're just using this in controller expand volume.

I think we at least need a comment that this is only used in that RPC so there's no future confusion.

sunnylovestiramisu: I also need to change the current implementation of the GCE Update Disk, there is a restriction right now that the request has to pass in one of iops or throughput, while it can take sizeGb only. And we will move the iops/throughput check in the actual ControllerExpandVolume func.

sunnylovestiramisu: Hmm it turns out to be more complicated, because currently we have resizeZonalDisk + resizeRegionalDisk, but for update we only have updateZonalDisk.

sunnylovestiramisu:
Will open another PR because the solution has changed.

/close

PR #1910, opened 2025-01-21
Add handling for TPC to GetRegionFromZones

hanz-vora:
Add handling for TPC to GetRegionFromZones

**Does this PR introduce a user-facing change?**
```
Bug fix preventing PDCSI driver from being compatible with TPC
```


pkg/common/utils.go
```diff
@@ -180,17 +180,22 @@ func NodeIDToZoneAndName(id string) (string, string, error) {
...
 
 func GetRegionFromZones(zones []string) (string, error) {
+	const tpcPrefix = "u"
 	regions := sets.String{}
 	if len(zones) < 1 {
 		return "", fmt.Errorf("no zones specified")
 	}
 	for _, zone := range zones {
 		// Zone expected format {locale}-{region}-{zone}
+		// TPC zone expected format is u-{}-{}-{}, e.g. u-europe-central2-a. See go/tpc-region-naming.
```
pwschuurman, line 190: Consider using a regex extract, to encode the "u-" prefix as optional.

PR #1911, opened 2025-01-22
Update kubetest2 version

pwschuurman:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Update `kubetest2` to a newer version for CI testgrid. This bumps kt2 to a version from December 16, 2024: https://github.com/kubernetes-sigs/kubetest2/commit/22d5b1410bef09ae679fa5813a5f0d196b6079de

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1914, opened 2025-01-25
Add XMLName back to TestSuite struct

pwschuurman:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Test runners that rely on parsing pdcsi_junit.xml are not working. This was changed in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1911 when the XMLName was removed for an output struct.

Fixes #

```release-note
NONE
```


pwschuurman:
Testsuite passing, except for two snapshot tests. These will be fixed in a subsequent PR

PR #1915, opened 2025-01-29
Add support for HdHA disk provisioning

tonyzhc:
**What type of PR is this?**
> /kind feature

This PR adds support for the PDCSI driver to provision Hyperdisk-balanced High Availability (HdHA) disks. This allows users to use the latest generation of regional disks from GCE. The feature is gated behind flag `allow-hdha-provisioning`.

Fixes b/391964605

One potential gap I observe is that we have to return the raw GCE error if users call `ControllerPublishVolume` on a HdHA disk in read only mode, because we cannot determine the disk type without actually fetching the disk. I think this is mostly fine since the GCE error is pretty clear (" Error 400: The disk cannot be attached with READ_ONLY mode."), but please let me know if I should add some more explicit error message.

```release-note
Add support for provisioning Hyperdisk-balanced High Availability disks.
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -72,6 +72,7 @@ var (
 	formatAndMountTimeout       = flag.Duration("format-and-mount-timeout", 1*time.Minute, "The maximum duration of a format and mount operation before another such operation will be started. Used only if --serialize-format-and-mount")
 	fallbackRequisiteZonesFlag  = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
 	enableStoragePoolsFlag      = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
+	enableHdHAFlag              = flag.Bool("allow-hdha-provisioning", false, "If set to true, will allow the driver to provision Hyperdisk-balanced High Availability disks")
 
 	multiZoneVolumeHandleDiskTypesFlag = flag.String("multi-zone-volume-handle-disk-types", "", "Comma separated list of allowed disk types that can use the multi-zone volumeHandle. Used only if --multi-zone-volume-handle-enable")
 	multiZoneVolumeHandleEnableFlag    = flag.Bool("multi-zone-volume-handle-enable", false, "If set to true, the multi-zone volumeHandle feature will be enabled")
@@ -222,7 +223,7 @@ func handle() {
 		}
 		initialBackoffDuration := time.Duration(*errorBackoffInitialDurationMs) * time.Millisecond
 		maxBackoffDuration := time.Duration(*errorBackoffMaxDurationMs) * time.Millisecond
-		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider, initialBackoffDuration, maxBackoffDuration, fallbackRequisiteZones, *enableStoragePoolsFlag, multiZoneVolumeHandleConfig, listVolumesConfig, provisionableDisksConfig)
+		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider, initialBackoffDuration, maxBackoffDuration, fallbackRequisiteZones, *enableStoragePoolsFlag, multiZoneVolumeHandleConfig, listVolumesConfig, provisionableDisksConfig, *enableHdHAFlag)
 	} else if *cloudConfigFilePath != "" {
 		klog.Warningf("controller service is disabled but cloud config given - it has no effect")
 	}
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -423,8 +423,8 @@ func ValidateDiskParameters(disk *CloudDisk, params common.DiskParameters) error
 	}
 
 	locationType := disk.LocationType()
-	if (params.ReplicationType == "none" && locationType != meta.Zonal) || (params.ReplicationType == "regional-pd" && locationType != meta.Regional) {
-		return fmt.Errorf("actual disk replication type %v did not match expected param %s", locationType, params.ReplicationType)
+	if (params.ReplicationType == "none" && locationType != meta.Zonal) || (params.IsRegional() && locationType != meta.Regional) {
+		return fmt.Errorf("actual replication type %v did not match expected param %s and %s", locationType, params.ReplicationType, params.DiskType)
 	}
 
 	if !KmsKeyEqual(
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -104,6 +104,10 @@ type GCEControllerServer struct {
 	// If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools.
 	enableStoragePools bool
 
+	// If set to true, the CSI Driver will allow Hyperdisk-balanced High Availability disks
+	// to be provisioned.
+	enableHdHA bool
+
 	multiZoneVolumeHandleConfig MultiZoneVolumeHandleConfig
 
 	listVolumesConfig ListVolumesConfig
@@ -167,10 +171,10 @@ type workItem struct {
 
 // locationRequirements are additional location topology requirements that must be respected when creating a volume.
 type locationRequirements struct {
-	srcVolRegion         string
-	srcVolZone           string
-	srcReplicationType   string
-	cloneReplicationType string
+	srcVolRegion    string
+	srcVolZone      string
+	srcIsRegional   bool
+	cloneIsRegional bool
 }
 
 // PDCSIContext is the extracted VolumeContext from controller requests.
@@ -206,7 +210,8 @@ const (
 
 	listDisksUsersField = googleapi.Field("items/users")
 
-	readOnlyManyAccessMode = "READ_ONLY_MANY"
+	gceReadOnlyManyAccessMode  = "READ_ONLY_MANY"
+	gceReadWriteManyAccessMode = "READ_WRITE_MANY"
 )
 
 var (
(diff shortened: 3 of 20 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -72,6 +72,7 @@ var (
 	formatAndMountTimeout       = flag.Duration("format-and-mount-timeout", 1*time.Minute, "The maximum duration of a format and mount operation before another such operation will be started. Used only if --serialize-format-and-mount")
 	fallbackRequisiteZonesFlag  = flag.String("fallback-requisite-zones", "", "Comma separated list of requisite zones that will be used if there are not sufficient zones present in requisite topologies when provisioning a disk")
 	enableStoragePoolsFlag      = flag.Bool("enable-storage-pools", false, "If set to true, the CSI Driver will allow volumes to be provisioned in Storage Pools")
+	enableHdHAFlag              = flag.Bool("allow-hdha-provisioning", false, "If set to true, will allow the driver to provisioin Hyperdisk-balanced High Availability disks")
```
amacaskill, line 75: provisioin-> provision

amacaskill:
Integration test failed, but I think this is the snapshot race condition. https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_gcp-compute-persistent-disk-csi-driver/1915/pull-gcp-compute-persistent-disk-csi-driver-kubernetes-integration/1886853026778451968 

/retest

pkg/gce-pd-csi-driver/gce-pd-driver_test.go
```diff
@@ -54,7 +54,7 @@ func controllerServerForTest(cloudProvider gce.GCECompute) *GCEControllerServer
 		SupportsThroughputChange: []string{"hyperdisk-balanced", "hyperdisk-throughput", "hyperdisk-ml"},
 	}
 
-	return NewControllerServer(gceDriver, cloudProvider, errorBackoffInitialDuration, errorBackoffMaxDuration, fallbackRequisiteZones, enableStoragePools, multiZoneVolumeHandleConfig, listVolumesConfig, provisionableDisksConfig)
+	return NewControllerServer(gceDriver, cloudProvider, errorBackoffInitialDuration, errorBackoffMaxDuration, fallbackRequisiteZones, enableStoragePools, multiZoneVolumeHandleConfig, listVolumesConfig, provisionableDisksConfig, true)
```
amacaskill, line 57: nit: for readability please add inline comment so we can easily see what this is: /* enableHdHA */  true

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1547,6 +1548,10 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 		}
 		return nil, common.LoggedError("CreateSnapshot, failed to getDisk: ", err)
 	}
+	isHdHA := disk.GetPDType() == common.ParameterHdHADiskType
+	if isHdHA && disk.GetAccessMode() == gceReadWriteManyAccessMode {
```
amacaskill, line 1552: If we make RWX as a command line flag, then we can enable support of this by flipping the flag in the component yaml, then creating new component CL. Without command line flag, to turn this on, we need to do entire OSS release process ([go/pdcsi-oss-release-process](http://goto.google.com/pdcsi-oss-release-process) before creating component CL.

amacaskill: If this isn't going to be supported, or we need other changes to support it, then how it is now,  is the correct way to  do it!

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1561,6 +1566,9 @@ func (gceCS *GCEControllerServer) CreateSnapshot(ctx context.Context, req *csi.C
 			return nil, err
 		}
 	case common.DiskImageType:
+		if isHdHA {
```
amacaskill, line 1569: Looks like this also isn't supported for regional pd: https://docs.google.com/document/d/1cJOw1tbD6JB6e3ePeskgNdZxRG2gQGHC-6aZ_x1qBpM/edit?disco=AAABAIXTNuM. See: https://cloud.google.com/compute/docs/disks/regional-persistent-disk#restrictions

Please add that as well. Looks like we missed it first time around

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -412,19 +412,19 @@ func (cloud *CloudProvider) ValidateExistingDisk(ctx context.Context, resp *Clou
...
-func ValidateDiskParameters(disk *CloudDisk, params common.DiskParameters) error {
+func ValidateDiskParameters(disk *CloudDisk, params common.DiskParameters, allowHdHA bool) error {
 	if disk.GetPDType() != params.DiskType {
 		return fmt.Errorf("actual pd type %s did not match the expected param %s", disk.GetPDType(), params.DiskType)
 	}
 
 	locationType := disk.LocationType()
-	if (params.ReplicationType == "none" && locationType != meta.Zonal) || (params.ReplicationType == "regional-pd" && locationType != meta.Regional) {
-		return fmt.Errorf("actual disk replication type %v did not match expected param %s", locationType, params.ReplicationType)
+	if (params.ReplicationType == "none" && locationType != meta.Zonal) || (params.IsRegional(allowHdHA) && locationType != meta.Regional) {
```
amacaskill, line 426: I think we want to explicitly error somewhere if someone uses HDHA when it is not supported. I don't see where you are doing that, but please correct me if I am wrong. It would either be in ValidateDiskParameters, or in ExtractAndDefaultParameters. See what we did for Storage Pools.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -361,7 +361,7 @@ func (gceCS *GCEControllerServer) createVolumeInternal(ctx context.Context, req
 	if _, err := getMultiWriterFromCapabilities(volumeCapabilities); err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "VolumeCapabilities is invalid: %v", err.Error())
 	}
-	err = validateStoragePools(req, params, gceCS.CloudProvider.GetDefaultProject())
+	err = validateStoragePools(req, params, gceCS.CloudProvider.GetDefaultProject(), gceCS.enableHdHA)
```
amacaskill, line 364: please add a unit test for this to TestValidateStoragePools

amacaskill: This is an open source repo, so we have a lot of contributors who are unaware of previous  features. This will help prevent someone who is unaware of your feature, from breaking the functionality.

tonyzhc:
Can confirm that the integration test is failing because of the race condition between createSnapshot and attach/detachDisk (b/345837974): https://screenshot.googleplex.com/BeTmv9fojEWedUF.

pkg/gce-pd-csi-driver/controller.go
tonyzhc: I think from https://cloud.google.com/compute/docs/disks/hyperdisks#limitations HdHA RWX is actually not allowed to create any kinds of backup (whether snapshot or image), so I am explicitly checking that here.

amacaskill: Looks like that is a limitation for all hyperdisk types. Can you please change the if check to check for all hyperdisk types? 

isHyperdisk := disk.GetPDType() starts with "hyperdisk-"
if isHyperdisk && disk.GetAccessMode() == gceReadWriteManyAccessMode {

pkg/gce-cloud-provider/compute/gce-compute.go
amacaskill: Now that you throw an error in ExtractAndDefaultParameters, I don't think we need allowHdHA in this function anymore correct? ExtractAndDefaultParameters does the validation regarding whether or not the feature is enabled, so if we get to this point, enableHDHA should always be true right?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -2165,13 +2165,21 @@ func pickZonesFromTopology(top *csi.TopologyRequirement, numZones int, locationT
 			}
 
 			// For zonal -> regional disk cloning, one of the replicated zones must match the source zone.
-			if locationTopReq.srcReplicationType == replicationTypeNone {
+			if !locationTopReq.srcIsRegional {
```
amacaskill, line 2168: nit: I think this is more readable as "locationTopReq.srcReplicationType == replicationTypeNone" because the comment says for  zonal -> regional cloning... You are changing this to be for !zonal -> regional cloning, which is just extra logic to do in your head.

pkg/gce-cloud-provider/compute/gce-compute.go
tonyzhc: I see that this function is used in `ValidateDiskCapabilities` RPC as well. I'm not sure how that RPC is called, but would it be better to keep this variable just in case we call this function in flows other than creating the disk?

pkg/gce-pd-csi-driver/controller.go
amacaskill: Makes sense. Thanks

amacaskill:
Please fix the latest comments I left here and over chat, then I will take one last look. I think its pretty close at this point!

pkg/gce-pd-csi-driver/controller.go
tonyzhc: Done.

pkg/gce-cloud-provider/compute/gce-compute.go
tonyzhc: Removed enabledHdHA.

test/run-e2e-local.sh
```diff
@@ -16,4 +16,4 @@ if hostname | grep -q c.googlers.com ; then
   CLOUDTOP_HOST=--cloudtop-host
 fi
 
-ginkgo --v "test/e2e/tests" -- --project "${PROJECT}" --service-account "${IAM_NAME}" "${CLOUDTOP_HOST}" --v=6 --logtostderr $@
+ginkgo --focus "HdHA" --v "test/e2e/tests" -- --project "${PROJECT}" --service-account "${IAM_NAME}" "${CLOUDTOP_HOST}" --v=6 --logtostderr $@
```
amacaskill, line 19: I think we can revert this change

test/run-e2e.sh
```diff
@@ -5,7 +5,7 @@ set -x
 
 readonly PKGDIR=sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 
-TIMEOUT=30m
+TIMEOUT=40m
```
amacaskill, line 8: We can also revert this change, since this doesn't make the test pass

test/run-e2e-local.sh
tonyzhc: Removed.

test/run-e2e.sh
tonyzhc: Removed.

PR #1918, opened 2025-02-03
Force timeout nodeunstagevolume

davis-haba:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Adds a 30 second timeout to NodeUnstageVolume. If a device has been in use for longer than 30 seconds, the next NodeUnstageVolume attempt to succeed. 

Fixes #

2 CMEK E2E tests are failing locally, but I believe this is due to test setup issues, as this change is entirely unrelated to CMEK.

```release-note
NONE
```


mattcary:
/ok-to-test

I'll let @pwschuurman comment on the PR.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -81,6 +85,8 @@ const (
 	fsTypeExt3                 = "ext3"
 
 	readAheadKBMountFlagRegexPattern = "^read_ahead_kb=(.+)$"
+
+	deviceInUseTimeout = 30
```
pwschuurman, line 89: Can you make this configurable with a feature flag? We may need to reduce this if 30 seconds is too high.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -459,11 +492,15 @@ func (ns *GCENodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
 		var targetErr *ignoreableError
```
pwschuurman, line 492: I know you didn't name this variable, but can  you change the name of `targetErr` to `ignoreableErr`? I think that's a better name that makes this a bit more clear.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -94,6 +100,33 @@ func getDefaultFsType() string {
 		return defaultLinuxFsType
 	}
 }
+
+// checkDeviceErrorTimeout returns true an error was encounted for the specified deviceName,
```
pwschuurman, line 104: s/encounted/encountered

pkg/gce-pd-csi-driver/node.go
pwschuurman: Also, consider making this a duration, so we have better type safety.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -51,6 +51,10 @@ type GCENodeServer struct {
 	// for that same volume (as defined by VolumeID) return an Aborted error
 	volumeLocks *common.VolumeLocks
 
+	// deviceInUseErrors keeps tracks of device names and a timestamp for when an error is
+	// encounted for that device
+	deviceInUseErrors map[string]time.Time
```
pwschuurman, line 56: There may be concurrent CSI API calls. Can you migrate this to an atomic data structure, or add a mechanism to serialize data access?

See VolumeLocks for inspiration: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/99542b07e40b599c4f9106cd96ff1ebeb577351a/pkg/common/volume_lock.go#L31

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1437,6 +1437,82 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+
+		// Unstage Disk. This should record a "deviceInUse" error
+		err = client.NodeUnstageVolume(volID, stageDir)
+		Expect(err).ToNot(BeNil(), "Expected failure during unstage")
+		Expect(err).To(MatchError(ContainSubstring(("is still in use"))))
+
+		// Wait 35s (30s timeout + 5s buffer) and try again
+		time.Sleep(35 * time.Second)
+		err = client.NodeUnstageVolume(volID, stageDir)
+		Expect(err).To(BeNil(), "Failed to unpublish after 30s in-use timeout for volume: %s, stageDir: %s", volID, stageDir)
```
pwschuurman, line 1508: I think this should succeed, right?

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1437,6 +1437,82 @@ var _ = Describe("GCE PD CSI Driver", func() {
...
+			unmountAndRmPrivateBindMount()
+		}()
+
+		// Unstage Disk. This should record a "deviceInUse" error
+		err = client.NodeUnstageVolume(volID, stageDir)
+		Expect(err).ToNot(BeNil(), "Expected failure during unstage")
+		Expect(err).To(MatchError(ContainSubstring(("is still in use"))))
+
+		// Wait 35s (30s timeout + 5s buffer) and try again
+		time.Sleep(35 * time.Second)
```
pwschuurman, line 1506: Ideally we would be able to advance the clock... but we can also reduce the timeout to something shorter (say ~10s via command line argument to PDCSI binary) to reduce the runtime of the test.

pkg/gce-pd-csi-driver/node.go
davis-haba: Done. Also moved deviceInUseErrors struct to new file: pkg/gce-pd-csi-driver/device_error_map.go

test/e2e/tests/single_zone_e2e_test.go
davis-haba: Yea, unless I'm missing something it does succeed. Expects err to be nil.

davis-haba:
@pwschuurman PTAL

The E2E test presubmits have been failing on setup, but I've ran the test locally and they pass. I'll continue to retest the presubmits.

cmd/gce-pd-csi-driver/main.go
```diff
@@ -63,6 +63,7 @@ var (
 	waitForOpBackoffJitter    = flag.Float64("wait-op-backoff-jitter", 0.0, "Jitter for wait for operation backoff")
 	waitForOpBackoffSteps     = flag.Int("wait-op-backoff-steps", 100, "Steps for wait for operation backoff")
 	waitForOpBackoffCap       = flag.Duration("wait-op-backoff-cap", 0, "Cap for wait for operation backoff")
+	deviceInUseTimeout        = flag.Duration("device-in-use-timeout", 30, "Max time to wait for a device to be unsed when attempting to unstage")
```
pwschuurman, line 66: This is meant to be 30s, right? I think an integer is interpreted as a 30 nanoseconds.

Also, what do you think about adding a feature flag to default disable the device in use check? As this feature adds a memory datastructure it may impact total memory usage. It may be useful to be able to disable the device in use check entirely.

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// deviceErrMap is an atomic data datastructure for recording deviceInUseError times
+// for specified devices
+type deviceErrMap struct {
+	mux               sync.Mutex
+	deviceInUseErrors map[string]time.Time
+}
+
+var DeviceInUseTimeout = time.Second * 30
```
pwschuurman, line 31: Can you set this as a field of `deviceErrMap` and plumb it through constructor instantiation? I'd rather have this set as a field on the Node server, rather than initialized as a global variable. Global variables limit the ability of this logic to be unit tested, and limit concurrency patterns if the driver is refactored.

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// markDeviceError updates the internal `deviceInUseErrors` map to denote an error was encounted
+// for the specified deviceName at the current time. If an error had previously been recorded, the
+// time will not be updated.
+func (devErrMap *deviceErrMap) markDeviceError(deviceName string) {
+	devErrMap.mux.Lock()
+	defer devErrMap.mux.Unlock()
+
+	if devErrMap.deviceInUseErrors == nil {
```
pwschuurman, line 54: Can you just set `devErrMap.deviceInUseErrors` in a constructor?

test/e2e/utils/utils.go
```diff
@@ -68,6 +68,7 @@ func GCEClientAndDriverSetup(instance *remote.InstanceInfo, driverConfig DriverC
 		"--use-instance-api-to-list-volumes-published-nodes",
 		"--supports-dynamic-iops-provisioning=hyperdisk-balanced,hyperdisk-extreme",
 		"--supports-dynamic-throughput-provisioning=hyperdisk-balanced,hyperdisk-throughput,hyperdisk-ml",
+		"--device-in-use-timeout=10", // Set lower than the usual value to expidite tests
```
pwschuurman, line 71: s/expidite/expedite

pwschuurman: Also, since this is a duration, it should be `--device-in-use-timeout=10s`

cmd/gce-pd-csi-driver/main.go
pwschuurman, line 66: s/unsed/unused

pkg/gce-pd-csi-driver/device_error_map.go
davis-haba: Yea good suggestion. I ended up plumping it through to the `deviceInUseErrors` struct, as the logic in NodeServer itself doesn't really care about it.

cmd/gce-pd-csi-driver/main.go
davis-haba: Done. All deviceInUse.* functions will short-circuit if the feature is disabled.

cmd/gce-pd-csi-driver/main.go
```diff
@@ -64,6 +64,9 @@ var (
 	waitForOpBackoffSteps     = flag.Int("wait-op-backoff-steps", 100, "Steps for wait for operation backoff")
 	waitForOpBackoffCap       = flag.Duration("wait-op-backoff-cap", 0, "Cap for wait for operation backoff")
 
+	enableDeviceInUseTimeout = flag.Bool("enable-device-in-use-timeout", true, "If set to true, ignores device in use errors when attempting to unstage a device if it has been stuck for longer than 'device-in-use-timeout'")
```
pwschuurman, line 67: I think the boolean flag should bypass the "device in use" check entirely, and be renamed `enable-device-in-use-check-on-node-unstage`. I don't see much value in having a boolean for using the timeout or not (eg: if a user really wants a disable the timeout, they can set `device-in-use-timeout` to infinity). I think the value of the switch more for users where the device in use check is blocking NodeUnstage due to the way `/sys/fs/` is reacting, and wants to skip in entirely, for all disks.

In retrospect, this is something that I should have added in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1658, to allow the feature to be turned off.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -456,14 +470,18 @@ func (ns *GCENodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
 	}
 
 	if err := ns.confirmDeviceUnused(volumeID); err != nil {
-		var targetErr *ignoreableError
-		if errors.As(err, &targetErr) {
+		var ignoreableErr *ignoreableError
+		if errors.As(err, &ignoreableErr) {
 			klog.Warningf("Unabled to check if device for %s is unused. Device has been unmounted successfully. Ignoring and continuing with unstaging. (%v)", volumeID, err)
+		} else if ns.deviceInUseErrors.checkDeviceErrorTimeout(volumeID) {
```
pwschuurman, line 476: There are some edge cases if a NodeUnstage that gets skipped due to force detach, or manual intervention. I don't know how often these could occur, and I think generally they're just theoretical problems:
1) A pre-existing old deviceInUse, resulting in the device in use signal being ignored. I'm less concerned about this one, as worst case it wouldn't block a user's workload, and this is a best effort deviceInUse check. We could solve this with a TTL cache, but I don't see an existing reference implementation we could easily use.
2) Device entries growing unbounded, if there are new unique volume IDs being added every now and then, and NodeUnstage gets skipped. In GCE the maximum number of concurrent disks is 128, so we could set a maximum cache entry to something larger, say 256 or 512 to bound the growth. This would allow for a fresh set of ~128 concurrent disks with room for some overlap. We could solve this with a LRU cache (github.com/hashicorp/golang-lru).

cmd/gce-pd-csi-driver/main.go
davis-haba: Ah, I see what you mean. Fixed --the `confirmDeviceUnused` in node.go will now always return nil if the flag is true.

pkg/gce-pd-csi-driver/node.go
davis-haba: I implemented the LRU cache you suggested to address point 2.

It looks the library you mentioned also has a cache that will auto-expire keys (e.g. TTL): https://github.com/hashicorp/golang-lru?tab=readme-ov-file#expirable-lru-cache-example

Do we want to use this? What would a reasonable TTL be? maybe `deviceInUseTimeout * 2`?

pwschuurman: > Do we want to use this? What would a reasonable TTL be? maybe deviceInUseTimeout * 2?

Yeah, that seems reasonable.

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+// checkDeviceErrorTimeout returns true an error was encountered for the specified deviceName,
+// where the error happened at least `deviceInUseTimeout` seconds ago.
+func (devErrMap *deviceErrMap) checkDeviceErrorTimeout(deviceName string) bool {
+	devErrMap.mux.Lock()
+	defer devErrMap.mux.Unlock()
+
+	lastErrTime, exists := devErrMap.cache.Get(deviceName)
```
pwschuurman, line 54: Wouldn't the cache hold the first error time, rather than the last? Maybe rename this to indicate something like "first time error encountered".

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,82 @@
...
+	return exists && currentTime().Sub(lastErrTime).Seconds() >= devErrMap.timeout.Seconds()
```
pwschuurman, line 55: Might be more readable to compare `Time` objects with `After()`

```
expirationTime := lastErrTime.Add(devErrMap.timeout)
return exists && currentTime().After(expirationTime)
```

pkg/gce-pd-csi-driver/node.go
```diff
@@ -456,14 +472,18 @@ func (ns *GCENodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
 	}
```
pwschuurman, line 473: I'd put the `EnableDeviceInUseCheck` check here, and wrap everything. That way you don't have references to confirmDevice or deviceInUse outside of that check (eg: this calls `ns.deviceInUseErrors.deleteDevice` even when the flag is false).

pkg/gce-pd-csi-driver/node.go
davis-haba: Thanks. I added the exirable cache. 
`c := expirable.NewLRU[string, time.Time](maxDeviceCacheSize, nil, timeout*2)`

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/hashicorp/golang-lru/v2/expirable"
+	"k8s.io/klog/v2"
+)
+
+// maxDeviceCacheSize specifies the maximum number if in-use devices to cache.
```
pwschuurman, line 25: Can you comment noting that 256 was selected, as 128 is the max PDs per VM? (eg: in case this changes in the future). https://cloud.google.com/compute/docs/disks/persistent-disks#pdnumberlimits

pkg/gce-pd-csi-driver/node.go
davis-haba: Sure, done.

pkg/gce-pd-csi-driver/device_error_map.go
davis-haba: Done

pkg/gce-pd-csi-driver/device_error_map.go
davis-haba: Done

pkg/gce-pd-csi-driver/device_error_map.go
davis-haba: Yes, good call. Renamed to `firstEncounteredErrTime`.

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,83 @@
...
+	c := expirable.NewLRU[string, time.Time](maxDeviceCacheSize, nil, timeout*2)
+
+	return &deviceErrMap{
+		cache:   c,
+		timeout: timeout,
+	}
+}
+
+// checkDeviceErrorTimeout returns true an error was encountered for the specified deviceName,
+// where the error happened at least `deviceInUseTimeout` seconds ago.
```
pwschuurman, line 50: I'd update this comment, now that the variable is renamed (and is a Duration).

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,83 @@
...
+func (devErrMap *deviceErrMap) checkDeviceErrorTimeout(deviceName string) bool {
```
pwschuurman, line 51: Nit on the naming. I'd prefer this to be updated to an action oriented function name, like `deviceErrorExpired()`

pkg/gce-pd-csi-driver/device_error_map.go
```diff
@@ -0,0 +1,83 @@
...
+	devErrMap.mux.Lock()
+	defer devErrMap.mux.Unlock()
+
+	firstEncounteredErrTime, exists := devErrMap.cache.Get(deviceName)
+	if !exists {
```
pwschuurman, line 56: Can you add a comment explaining why `false` is returned even if the device does not exist in the map? (eg: this would indicate the first time the error is encountered).

pkg/gce-pd-csi-driver/node.go
```diff
@@ -455,13 +471,19 @@ func (ns *GCENodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
...
-		var targetErr *ignoreableError
-		if errors.As(err, &targetErr) {
-			klog.Warningf("Unabled to check if device for %s is unused. Device has been unmounted successfully. Ignoring and continuing with unstaging. (%v)", volumeID, err)
-		} else {
-			return nil, status.Errorf(codes.Internal, "NodeUnstageVolume for volume %s failed: %v", volumeID, err)
+	if ns.enableDeviceInUseCheck {
+		if err := ns.confirmDeviceUnused(volumeID); err != nil {
+			var ignoreableErr *ignoreableError
+			if errors.As(err, &ignoreableErr) {
+				klog.Warningf("Unabled to check if device for %s is unused. Device has been unmounted successfully. Ignoring and continuing with unstaging. (%v)", volumeID, err)
```
pwschuurman, line 478: s/Unabled/Unable

pwschuurman (approved):

PR #1901, opened 2025-01-10
changes to support hyperdisk multi-writer mode

karkunpavan:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Hyper disk multi-writer support is now in [compute/v1](https://pkg.go.dev/google.golang.org/api/compute/v1) but the persistent disk csi driver still calls [v0.beta](https://pkg.go.dev/google.golang.org/api/compute/v0.beta) which leads a error while creating a hyperdisk with multi-writer mode.

Error:
```
Warning ProvisioningFailed 31m (x15 over 57m) pd.csi.storage.gke.io_wduan-1030a-g-vdnqh-master-0_dc46e9f6-1725-4f15-922c-99f0e281e102 failed to provision volume with StorageClass
"balanced-storage": rpc error: code = InvalidArgument desc = CreateVolume failed: rpc error: code = InvalidArgument desc = CreateVolume failed to create single zonal disk pvc-0b65d680-636b-46c6-876d-d3a6c412c3ef: failed to insert zonal disk: unknown Insert disk error:
googleapi: Error 400: Invalid value for field 'resource.multiWriter': 'true'.
Cannot specify the multi writer field for 'hyperdisk-balanced' disks,
please use access mode instead., invalid%!v(MISSING)
```

To fix this here are the changes done:
1. Add AccessMode param to [DiskParameters struct](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/96338cae8f2a0f04d68f21af525556fb1a5fcc5e/pkg/common/parameters.go#L73)
2. Initialize the AccessMode field to the input value in ExtractAndDefaultParameters(). We do not have an explicit default - that will be decided by the PD layer depending on disk type.
3. The parameter is read by controller in [createVolumeInternal()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/96338cae8f2a0f04d68f21af525556fb1a5fcc5e/pkg/gce-pd-csi-driver/controller.go#L307)
4. The accessMode before calling createSingleZoneDisk() and createRegionalDisk() is currently set to "", the change fixes [createsingledisk()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/96338cae8f2a0f04d68f21af525556fb1a5fcc5e/pkg/gce-pd-csi-driver/controller.go#L599) to read from the parameter passed by the user.
[...]

PR #1919, opened 2025-02-05
Feature: Changes to support Hyperdisk Multi-Writer mode by moving to Beta APIs.

sjswerdlow:
**What type of PR is this?**
/kind bug

Hyper disk multi-writer support is now in [compute/v1](https://pkg.go.dev/google.golang.org/api/compute/v1) but the persistent disk csi driver still calls [v0.beta](https://pkg.go.dev/google.golang.org/api/compute/v0.beta) which leads a error while creating a hyperdisk with multi-writer mode.

Error:

```Warning ProvisioningFailed 31m (x15 over 57m) pd.csi.storage.gke.io_wduan-1030a-g-vdnqh-master-0_dc46e9f6-1725-4f15-922c-99f0e281e102 failed to provision volume with StorageClass
"balanced-storage": rpc error: code = InvalidArgument desc = CreateVolume failed: rpc error: code = InvalidArgument desc = CreateVolume failed to create single zonal disk pvc-0b65d680-636b-46c6-876d-d3a6c412c3ef: failed to insert zonal disk: unknown Insert disk error:
googleapi: Error 400: Invalid value for field 'resource.multiWriter': 'true'.
Cannot specify the multi writer field for 'hyperdisk-balanced' disks,
please use access mode instead., invalid%!v(MISSING)
```

Fixes: #1863 

Implementation of Proposal V2 from Matt. Changes done on top of https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1901.

```release-note
This PR introduces READ_WRITE_MANY AccessMode also know as Multi-writer support for hyperdisks. Refer [here](https://cloud.google.com/compute/docs/disks/sharing-disks-between-vms#hd-multi-writer) for details about the multi-writer mode support, disk and machine types that support it.

Users can enable this mode by setting `access-mode: "READ_WRITE_MANY"` for the corresponding disk section to attach it in multi-writer mode.
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/fake-gce.go
+++ b/pkg/gce-cloud-provider/compute/fake-gce.go
@@ -212,11 +212,6 @@ func (cloud *FakeCloudProvider) ValidateExistingDisk(ctx context.Context, resp *
 			params.DiskType, respType[len(respType)-1])
 	}
 
-	// We are assuming here that a multiWriter disk could be used as non-multiWriter
-	if multiWriter {
-		return fmt.Errorf("disk already exists with incompatible capability. Need MultiWriter. Got non-MultiWriter")
-	}
-
 	klog.V(4).Infof("Compatible disk already exists")
 	return ValidateDiskParameters(resp, params)
 }
--- a/pkg/gce-cloud-provider/compute/fake-gce.go
+++ b/pkg/gce-cloud-provider/compute/fake-gce.go
@@ -79,6 +79,10 @@ func CreateFakeCloudProvider(project, zone string, cloudDisks []*CloudDisk) (*Fa
 		mockDiskStatus: "READY",
 	}
 	for _, d := range cloudDisks {
+		if d.LocationType() == meta.Regional {
+			fcp.disks[meta.RegionalKey(d.GetName(), d.GetRegion()).String()] = d
+			continue
+		}
 		diskZone := d.GetZone()
 		if diskZone == "" {
 			diskZone = zone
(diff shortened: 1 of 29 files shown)


pkg/gce-cloud-provider/compute/fake-gce.go
```diff
@@ -212,11 +212,6 @@ func (cloud *FakeCloudProvider) ValidateExistingDisk(ctx context.Context, resp *
 			params.DiskType, respType[len(respType)-1])
 	}
 
-	// We are assuming here that a multiWriter disk could be used as non-multiWriter
-	if multiWriter && !resp.GetMultiWriter() {
```
mattcary, line 216: I think we need to keep this check, to validate that we aren't trying to use a non-multiwriter disk when we need a multiwriter one?

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -407,11 +395,6 @@ func (cloud *CloudProvider) ValidateExistingDisk(ctx context.Context, resp *Clou
 			reqBytes, common.GbToBytes(resp.GetSizeGb()), limBytes)
 	}
 
-	// We are assuming here that a multiWriter disk could be used as non-multiWriter
```
mattcary, line 410: Ditto, we should keep the multiwriter check.

Hmm, this won't verify that a hyperdisk is multiwriter, however, since we aren't passing in the access mode. It looks like it's just called from the one place in pk/gce-pd-csi-driver/controller.go, so should be easy to add accessMode to the method (this makes it parallel to the insert disk call. Part of me thinks we should push this detail of how multiwriter is declared down into cloud disk, but we can defer that for now, HdHA might clarify the question).

amacaskill: HdHA PR has already been merged by the way: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1915

mattcary:
/lgtm
/approve

Thanks for working though this!

mattcary:
The integration test errors are known problems with the snapshot tests. Force-merging.

PR #1940, opened 2025-02-14
switch to use snapshots.insert instead of disks.createsnapshot

pwschuurman:
**What type of PR is this?**
/kind bug

If we use disks.CreateSnapshot, there could be a race condition where snapshot.Status is "READY" but the snapshot creation is not actually done. Snapshots.Insert API avoids this problem.

Fixes #

- snapshot.insert will try to grab resource lock too according to Xin.
- CMEK implicitly will be populated from PD if left blank, which keeps the previous behavior

```release-note
use `snapshots.insert` instead of `disks.createsnapshot` when creating pd snapshots. This requires usage of `csi-snapshottter` with minimum version 7.0.0
```

### DIFF
--- a/deploy/kubernetes/images/stable-master/image.yaml
+++ b/deploy/kubernetes/images/stable-master/image.yaml
@@ -31,7 +31,7 @@ metadata:
   name: imagetag-csi-snapshotter
 imageTag:
   name: registry.k8s.io/sig-storage/csi-snapshotter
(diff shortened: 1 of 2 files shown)


pwschuurman:
Fixed in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1857

PR #1945, opened 2025-02-20
Derive access mode from the CreateVolume request instead of parameters

tonyzhc:
**What type of PR is this?**
>
> /kind bug

Previously https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1919 introduces multi-writer mode support for hyperdisks, but the access mode was derived through the parameters list. This is not the standard Kubernetes way of configuring access mode; rather we should derive it from the Volume Capabilities list in the CreateVolume request. This would confuse customers if they provide conflicting information in StorageClass vs PVC.

Fixes b/397726676.

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -209,9 +209,6 @@ const (
 	resourceProject    = "projects"
 
 	listDisksUsersField = googleapi.Field("items/users")
-
-	gceReadOnlyManyAccessMode  = "READ_ONLY_MANY"
-	gceReadWriteManyAccessMode = "READ_WRITE_MANY"
 )
 
 var (
@@ -532,12 +529,12 @@ func (gceCS *GCEControllerServer) updateAccessModeIfNecessary(ctx context.Contex
 		return nil
 	}
 	project := gceCS.CloudProvider.GetDefaultProject()
-	if disk.GetAccessMode() == gceReadOnlyManyAccessMode {
+	if disk.GetAccessMode() == common.GCEReadOnlyManyAccessMode {
 		// If the access mode is already readonly, return
 		return nil
 	}
 
-	return gceCS.CloudProvider.SetDiskAccessMode(ctx, project, volKey, gceReadOnlyManyAccessMode)
+	return gceCS.CloudProvider.SetDiskAccessMode(ctx, project, volKey, common.GCEReadOnlyManyAccessMode)
 }
 
 func (gceCS *GCEControllerServer) createSingleDeviceDisk(ctx context.Context, req *csi.CreateVolumeRequest, params common.DiskParameters) (*csi.CreateVolumeResponse, error) {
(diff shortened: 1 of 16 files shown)


mattcary:
Thanks for sorting this out, @tonyzhc ! 

one minor comment above

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -253,6 +261,24 @@ func getReadOnlyFromCapabilities(vcs []*csi.VolumeCapability) (bool, error) {
 	return false, nil
 }
 
+func hyperdiskAccessModeFrom(vcs []*csi.VolumeCapability) (string, error) {
```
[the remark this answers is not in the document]
tonyzhc: Done.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -599,11 +609,23 @@ func (gceCS *GCEControllerServer) createSingleDeviceDisk(ctx context.Context, re
...
 	capacityRange := req.GetCapacityRange()
 	capBytes, _ := getRequestCapacity(capacityRange)
-	multiWriter, _ := getMultiWriterFromCapabilities(req.GetVolumeCapabilities())
 	readonly, _ := getReadOnlyFromCapabilities(req.GetVolumeCapabilities())
-	accessMode := params.AccessMode
+	accessMode := ""
+	multiWriter := false
+	if common.IsHyperdisk(params.DiskType) {
+		if !disksWithUnsettableAccessMode[params.DiskType] {
+			if am, err := getHyperdiskAccessModeFromCapabilities(req.GetVolumeCapabilities()); err != nil {
```
mattcary, line 617: Just making sure I understand, this check is actually redundant, as we should prevent unsettable access modes from getting set up in createVolumeInternal?

tonyzhc: We need this because we could still derive "READ_WRITE_SINGLE" from the volume capabilities, which is a valid mode for these disks; createVolumeInternal only throws error if the volume capabilities contain multi-node access modes (namely "READ_WRITE_MANY" and "READ_ONLY_MANY").

mattcary: So if this is an unsettable type (HdT etc), then we'll silently ignore req.GetVolumeCapabilities (or at least the access mode from them)? I'm not sure that's right, I think we should error if disksWithUnsettableAccessMode && am != RW_SINGLE.

Or am I getting confused about something?

tonyzhc: That check is performed somewhere else (see other comment). Let me know if you think there's improvement I can make here to make the code more understandable.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -377,8 +378,17 @@ func (gceCS *GCEControllerServer) createVolumeInternal(ctx context.Context, req
...
+	if readonly && params.DiskType == common.DiskTypeHdHA {
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid access mode for disk type %s", common.DiskTypeHdHA)
+	}
+
+	// Hyperdisk-throughput and hyperdisk-extreme do not support attaching to multiple VMs.
+	isMultiAttach, err := getMultiAttachementFromCapabilities(volumeCapabilities)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "CreateVolume failed to parse volume capabilities: %v", err)
+	}
+	if isMultiAttach && disksWithUnsettableAccessMode[params.DiskType] {
```
tonyzhc, line 390: @mattcary this is where we error if disksWithUnsettableAccessMode && am != RW_SINGLE

pkg/gce-pd-csi-driver/controller.go
mattcary: I think we should return an error if disksWithUnsettableAccessMode[params.DiskType] == true here, even if it's redundant with the check above. It will save us if there is a future refactor that changes the order of validation.

tonyzhc: Good point, done.

pkg/gce-pd-csi-driver/controller.go
mattcary: ack.

tonyzhc:
Looks like the integration tests are failing with the same volume snapshot flakes.

mattcary:
The windows test isn't required, so this was able to merge. The main e2e passed (the snapshot tests are just flakey IIUC)

PR #1950, opened 2025-02-24
Update RAID logic and post-RAID validation to integrate Datacache support for GKE nodes

hungnguyen243:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Update RAIDing logic to fetch requested local SSD count for Datacache from GKE node label, set up  RAIDing for Datacache with the corresponding local SSD number, and validate RAID-ing config afterwards.

Fixes #

```release-note
Update RAID logic and post-RAID validation to integrate Data Cache support for GKE nodes
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -27,12 +27,8 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/strings/slices"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/deviceutils"
 	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
@@ -102,9 +98,7 @@ var (
 )
 
 const (
-	driverName          = "pd.csi.storage.gke.io"
-	dataCacheLabel      = "datacache-storage-gke-io"
-	dataCacheLabelValue = "enabled"
+	driverName = "pd.csi.storage.gke.io"
 )
 
 func init() {
@@ -350,29 +344,74 @@ func urlFlag(target **url.URL, name string, usage string) {
 	})
 }
 
-func setupDataCache(ctx context.Context, nodeName string) error {
-	klog.V(2).Infof("Setting up data cache for node %s", nodeName)
-	if nodeName != common.TestNode {
-		cfg, err := rest.InClusterConfig()
-		if err != nil {
-			return err
+func fetchLssdsForRaiding(lssdCount int) ([]string, error) {
+	allLssds, err := driver.FetchAllLssds()
+	if err != nil {
+		return nil, fmt.Errorf("Error listing all LSSDs %v", err)
+	}
+
+	raidedLssds, err := driver.FetchRaidedLssds()
+	if err != nil {
+		return nil, fmt.Errorf("Error listing RAIDed LSSDs %v", err)
+	}
+
+	unRaidedLssds := []string{}
+	for _, l := range allLssds {
+		if !slices.Contains(raidedLssds, l) {
+			unRaidedLssds = append(unRaidedLssds, l)
 		}
-		kubeClient, err := kubernetes.NewForConfig(cfg)
-		if err != nil {
-			return err
-		}
-		node, err := kubeClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
-		if err != nil {
-			// We could retry, but this error will also crashloop the driver which may be as good a way to retry as any.
-			return err
-		}
-		if val, found := node.GetLabels()[dataCacheLabel]; !found || val != dataCacheLabelValue {
-			klog.V(2).Infof("Datacache not enabled for node %s; node label %s=%s and not %s", nodeName, dataCacheLabel, val, dataCacheLabelValue)
-			return nil
+		if len(unRaidedLssds) == lssdCount {
+			break
 		}
 	}
-	klog.V(2).Info("Raiding local ssds to setup data cache")
-	if err := driver.RaidLocalSsds(); err != nil {
+
+	LSSDsWithEmptyMountPoint, err := driver.FetchLSSDsWihtEmptyMountPoint()
+	if err != nil {
+		return nil, fmt.Errorf("Error listing LSSDs with empty mountpoint: %v", err)
+	}
+
+	// We need to ensure the disks to be used for Datacache are both unRAIDed & not containing mountpoints for ephemeral storage already
+	availableLssds := slices.Filter(nil, unRaidedLssds, func(e string) bool {
+		return slices.Contains(LSSDsWithEmptyMountPoint, e)
+	})
+
+	if len(availableLssds) == 0 {
+		return nil, fmt.Errorf("No LSSDs available to set up caching")
+	}
+
+	if len(availableLssds) < lssdCount {
+		return nil, fmt.Errorf("Not enough LSSDs available to set up caching. Available LSSDs: %v, wanted LSSDs: %v", len(availableLssds), lssdCount)
+	}
+	return availableLssds, nil
+}
+
+func setupDataCache(ctx context.Context, nodeName string) error {
+	isAlreadyRaided, err := driver.IsRaided()
+	if err != nil {
+		klog.V(2).Infof("Errored while scanning for available LocalSSDs err:%v; continuing Raiding", err)
+	} else if isAlreadyRaided {
+		klog.V(2).Infof("Local SSDs are already RAIDed. Skipping Datacache setup.")
+		return nil
+	}
+
+	lssdCount := common.LocalSSDCountForDataCache
+	if nodeName != common.TestNode {
+		var err error
+		lssdCount, err = driver.GetDataCacheCountFromNodeLabel(ctx, nodeName)
+		if lssdCount == 0 {
+			klog.Infof("Datacache is not enabled on node %v", nodeName)
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+	}
+	lssdNames, err := fetchLssdsForRaiding(lssdCount)
+	if err != nil {
+		klog.Fatalf("Failed to get sufficient SSDs for Datacache's caching setup: %v", err)
+	}
+	klog.V(2).Infof("Raiding local ssds to setup data cache: %v", lssdNames)
+	if err := driver.RaidLocalSsds(lssdNames); err != nil {
 		return fmt.Errorf("Failed to Raid local SSDs, unable to setup data caching, got error %v", err)
 	}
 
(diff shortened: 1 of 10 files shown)


sunnylovestiramisu:
Please update release note in the PR description to:

```
Update RAID logic and post-RAID validation to integrate Data Cache support for GKE nodes
```

The script we use to create a change log for the release is reading these release notes to automatically generate the change list.

pkg/common/constants.go
```diff
@@ -44,4 +44,11 @@ const (
 	ContexLocalSsdCacheSize = "local-ssd-cache-size"
 	// Node name for E2E tests
 	TestNode = "test-node-csi-e2e"
+
+	// Default LSSD count for datacache E2E tests
+	LocalSSDCountForDataCache = 2
+
+	// Node label for datacache
```
sunnylovestiramisu, line 51: Add a comment saying "these labels are only available on GKE" because we have other customers deploy PDCSI driver on GCE.

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -18,25 +20,37 @@ const (
...
+func fetchRAIDedLocalSsdPath() (string, error) {
+	args := []string{
+		"--detail",
+		"--scan",
+	}
+	info, err := common.RunCommand("grep", []string{raidedLocalSsdName}, "mdadm", args...)
+	if err != nil || len(info) == 0 {
+		return "", fmt.Errorf("Error getting RAIDed device path for Datacache %v, output:%v ===============", err, string(info))
+	}
+	infoString := strings.TrimSpace(string(info))
```
sunnylovestiramisu, line 34: Provide a sample info string as part of the a comment will help later someone else comes to read the code.

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -174,24 +191,161 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+	}
+	if val, found := node.GetLabels()[fmt.Sprintf(common.NodeLabelPrefix, common.DataCacheLssdCountLabel)]; found {
+		dataCacheCount, err := strconv.Atoi(val)
+		if err != nil {
+			return -1, fmt.Errorf("Error getting Datacache's LSSD count from node label: %v", err)
+		}
+		klog.Infof("Number of local SSDs requested for Datacache: %v", dataCacheCount)
+		return dataCacheCount, nil
+	}
+	return 0, fmt.Errorf("Cannot get Datacache's LSSD count from node label")
```
sunnylovestiramisu, line 251: There are nodes do not have data cache disk configured, these nodes should not return an error?

Do we have unit tests coverage for these cases?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -174,24 +191,161 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+			return fmt.Errorf("Local SSDs are not setup for caching")
+		}
+		return nil
+	}
+	klog.Infof("Data cache is not enabled for PVC")
+	return nil
+}
+
+func GetDataCacheCountFromNodeLabel(ctx context.Context, nodeName string) (int, error) {
+	if nodeName == common.TestNode {
```
sunnylovestiramisu, line 225: This will only affect customers using PDCSI driver on GCE with the same name, correct? And the mitigation is tell these customers do not name their node "test-node-csi-e2e".

hungnguyen243: Ah this is actually duplicate logic as we already checked if nodeName != common.TestNode in the caller before invoking this function. I'll just remove this.

cmd/gce-pd-csi-driver/main.go
```diff
@@ -350,32 +344,77 @@ func urlFlag(target **url.URL, name string, usage string) {
...
+		var err error
+		lssdCount, err = driver.GetDataCacheCountFromNodeLabel(ctx, nodeName)
 		if err != nil {
-			// We could retry, but this error will also crashloop the driver which may be as good a way to retry as any.
 			return err
 		}
-		if val, found := node.GetLabels()[dataCacheLabel]; !found || val != dataCacheLabelValue {
-			klog.V(2).Infof("Datacache not enabled for node %s; node label %s=%s and not %s", nodeName, dataCacheLabel, val, dataCacheLabelValue)
+		if lssdCount == 0 {
+			klog.Infof("Data Cache is not enabled on node %v", nodeName)
```
Sneha-at, line 405: "Data Cache is not enabled on node %v so skipping caching setup", nodeName"

cmd/gce-pd-csi-driver/main.go
```diff
@@ -350,32 +344,77 @@ func urlFlag(target **url.URL, name string, usage string) {
...
+	if err != nil {
+		klog.Fatalf("Failed to get sufficient SSDs for Data Cache's caching setup: %v", err)
+	}
+	klog.V(2).Infof("Raiding local ssds to setup Data Cache: %v", lssdNames)
+	if err := driver.RaidLocalSsds(lssdNames); err != nil {
+		return fmt.Errorf("Failed to Raid local SSDs, unable to setup Data Cache, got error %v", err)
 	}
 
-	klog.V(2).Infof("Datacache enabled for node %s", nodeName)
+	klog.V(2).Infof("Data Cache enabled for node %s", nodeName)
```
Sneha-at, line 418: Let's update this to
```
"LSSD caching setup for the data cache enabled node %v", nodeName
```

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -18,25 +20,37 @@ const (
...
+	}
+	info, err := common.RunCommand("grep", []string{raidedLocalSsdName}, "mdadm", args...)
+	if err != nil || len(info) == 0 {
+		return "", fmt.Errorf("Error getting RAIDed device path for Data Cache %v, output:%v", err, string(info))
+	}
+	infoString := strings.TrimSpace(string(info))
+	infoSlice := strings.Split(infoString, " ")
+
+	// We want to get the second element in the array, which is the path to the RAIDed device
+	return infoSlice[1], nil
```
Sneha-at, line 38: @sunnylovestiramisu do you think we can add any regex for any other check here to ensure we always pick up the path. 
Command output:
```
$ sudo mdadm --detail --scan
ARRAY /dev/md/kubelet_ephemeral_storage metadata=1.2 name=kubelet_ephemeral_storage UUID=xxxx
ARRAY /dev/md126 metadata=1.2 name=csi-driver-data-cache UUID=xxxx
```

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -174,24 +191,159 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+		isAlreadyRaided, err := IsRaided()
+		if err != nil {
+			return fmt.Errorf("Local SSDs are not setup for caching; got error: %v", err)
+		}
+		if !isAlreadyRaided {
+			return fmt.Errorf("Local SSDs are not setup for caching")
+		}
+		return nil
+	}
+	klog.Infof("Data Cache is not enabled for PVC")
```
Sneha-at, line 220: Can we add the values for these 2 parameters too in the error message so the customer knows what's missing.
```
klog.V(2).Infof("Data Cache is not enabled for PVC, set data cache size and mode in storage class to enable caching")
```

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -174,24 +191,159 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+	if err != nil {
+		// We could retry, but this error will also crashloop the driver which may be as good a way to retry as any.
+		return 0, err
+	}
+	if val, found := node.GetLabels()[fmt.Sprintf(common.NodeLabelPrefix, common.DataCacheLssdCountLabel)]; found {
+		dataCacheCount, err := strconv.Atoi(val)
+		if err != nil {
+			return 0, fmt.Errorf("Error getting Data Cache's LSSD count from node label: %v", err)
+		}
+		klog.Infof("Number of local SSDs requested for Data Cache: %v", dataCacheCount)
```
Sneha-at, line 245: @sunnylovestiramisu Do we have any recommended logging level for all info logs?

cmd/gce-pd-csi-driver/main.go
sunnylovestiramisu: whichever string we pick for the feature 'data cache' or 'Data cache', keep it consistent across the driver, we do not have the "=====" anymore will rely on this string to trace.

pkg/gce-pd-csi-driver/cache.go
sunnylovestiramisu: Since we are greping the "raidedLocalSsdName" will always get 1 array(hopefully):
ARRAY /dev/md126 metadata=1.2 name=csi-driver-data-cache UUID=xxxx

The path itself should be in some format /dev/md* ?

pkg/gce-pd-csi-driver/cache.go
sunnylovestiramisu: https://kingye.me/study-go/docs/advanced/pkg/klog/ This is what I usually use to refer to, but I will say for us, look at the existing pdcsi logs, do they specify log levels? We should keep it consistent with existing logs.

sunnylovestiramisu (commented):
Review for data cache

pkg/gce-pd-csi-driver/cache.go
hungnguyen243: I think if the grep 'csi-driver-data-cache' is able to fetch the line, the path should be correct since how else is a RAID array created with the same name but at some random path? A possible error is when the RAID fails and so the grep will not return anything, which is caught before we process the infoString.

Sneha-at: It might as well be `/dev/csi-driver-data-cache` for our E2E tests and in certain scenarios :( like if ephemeral storage is not used. @hungnguyen243 to confirm

hungnguyen243: For our E2E tests, it will be /dev/md/csi-driver-datacache (I double checked on this), so adding a regex check should be fine, but as I mentioned above, it shouldn't be necessary?

pkg/gce-pd-csi-driver/node.go
```diff
@@ -337,14 +337,18 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
 		if len(nodeId) == 0 {
 			return nil, status.Error(codes.InvalidArgument, "NodeStageVolume Node ID must be provided")
 		}
 		devFsPath, err := filepath.EvalSymlinks(devicePath)
 		if err != nil {
 			klog.Errorf("filepath.EvalSymlinks(%q) failed when trying to create volume group: %v", devicePath, err)
 		}
+		configError := ValidateDataCacheConfig(req.GetPublishContext()[common.ContextDataCacheMode], req.GetPublishContext()[common.ContextDataCacheSize], ctx, nodeId)
+		if configError != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Error validate configuration for Data Cache: %v", err.Error()))
```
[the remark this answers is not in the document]
hungnguyen243: Synced offline; will resolve this with another PR.

pkg/gce-pd-csi-driver/cache.go
hungnguyen243: Changed to V(4) logging.

pkg/gce-pd-csi-driver/cache.go
sunnylovestiramisu: I am not a fan of regex check without test coverage anyway, and for this one it seems not necessary to add really.

sunnylovestiramisu:
/lgtm
/approve

We need @mattcary to force merge because the windows test is not fixed yet.

PR #1952, opened 2025-02-25
Mount error metrics branch

hajiler:
**What type of PR is this?**
/kind feature

This PR adds metrics capability for the PDCSI node server. Metrics enable us to define SLO/Alerts for the node server. This PR specifically adds a metric for mount errors. 

Fixes #

```release-note
adds metrics for node server including:
- 'process_start_time': start time of the container for use in metric scraping
- 'node_mount_errors': error types which occur during volume mounting
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"k8s.io/component-base/metrics"
 	"k8s.io/klog/v2"
+	"k8s.io/mount-utils"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 )
 
@@ -53,6 +54,15 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		},
 		[]string{"driver_name", "method_name", "grpc_status_code", "disk_type", "enable_confidential_storage", "enable_storage_pools"})
+
+	mountErrorMetric = metrics.NewCounterVec(&metrics.CounterOpts{
+		Subsystem:      "node",
+		Name:           "mount_errors",
+		Help:           "Node server file system mounting errors",
+		StabilityLevel: metrics.ALPHA,
+	},
+		[]string{"error_type"},
+	)
 )
 
 type MetricsManager struct {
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -53,6 +54,15 @@ var (
...
 		},
 		[]string{"driver_name", "method_name", "grpc_status_code", "disk_type", "enable_confidential_storage", "enable_storage_pools"})
+
+	mountErrorMetric = metrics.NewCounterVec(&metrics.CounterOpts{
+		Subsystem:      "node",
+		Name:           "mount_errors",
+		Help:           "Node server file system mounting errors",
+		StabilityLevel: metrics.ALPHA,
+	},
+		[]string{"error_type"},
```
sunnylovestiramisu, line 64: Do we have a SLO design doc for this metrics? We probably also want info of "driver_name", "grpc_status_code", "disk_type" etc. extra information?

sunnylovestiramisu: The thing is we need success calls and error calls to calculate the right SLO, so we probably need to emit both.

msau42: Check out existing csi/k8s volume metrics to follow similar patterns.

pkg/metrics/metrics.go
```diff
@@ -101,6 +115,19 @@ func (mm *MetricsManager) RecordOperationErrorMetrics(
 	klog.Infof("Recorded PDCSI operation error code: %q", errCode)
 }
 
+func (mm *MetricsManager) RecordMountErrorMetric(fs_format string, err error) {
+	mntErr := &mount.MountError{}
+	if errors.As(err, mntErr) {
+		mountErrorMetric.WithLabelValues(pdcsiDriverName, fs_format, string(mntErr.Type)).Inc()
```
sunnylovestiramisu, line 121: Is this error an count? How do we calculate the SLOs if we do not have the success rate as well?

sunnylovestiramisu:
Has this change been end to end tested? For example on a GCE cluster, you can intentionally return an error at the mount step to trigger the emission.

pkg/metrics/metrics.go
```diff
@@ -101,6 +115,21 @@ func (mm *MetricsManager) RecordOperationErrorMetrics(
 	klog.Infof("Recorded PDCSI operation error code: %q", errCode)
 }
 
+func (mm *MetricsManager) RecordMountErrorMetric(fs_format string, err error) {
```
sunnylovestiramisu, line 118: Can we have some unit tests for this function?

In the past errors.As had some weird bugs of not being able to parse the error type as we expected. 

Other than this, LGTM.

hajiler: Got it, done.

sunnylovestiramisu:
For other reviewers if you want them to take a look as well
/hold

PR #1954, opened 2025-02-25
Fix chunksize bug for large cache size in data cache

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Data cache : Setting up caching would error out for large cache size as LVM has a limit on number of chunks that could be used, this PR fixes that issue by calculating dynamic cache size

Fixes #
Internal error, unreleased feature

```release-note
NONE
```

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -713,6 +713,13 @@ func ValidateDataCacheMode(s string) error {
 	return fmt.Errorf("invalid data-cache-mode %s. Only \"writeback\" and \"writethrough\" is a valid input", s)
 }
 
+func ValidateNonNegativeInt(n int64) error {
+	if n <= 0 {
+		return fmt.Errorf("Input should be set to > 0, got %d", n)
+	}
+	return nil
+}
+
 // NewLimiter returns a token bucket based request rate limiter after initializing
 // the passed values for limit, burst (or token bucket) size. If opted for emptyBucket
 // all initial tokens are reserved for the first burst.
--- a/pkg/gce-pd-csi-driver/cache.go
+++ b/pkg/gce-pd-csi-driver/cache.go
@@ -17,13 +17,18 @@ import (
 )
 
 const (
-	cacheSuffix              = "csi-fast"
-	mainLvSuffix             = "csi-main"
-	raidedLocalSsdName       = "csi-driver-data-cache"
-	raidMode                 = "0"
-	maxAllowedChunks   int64 = 1000000 // This is the max allowed chunks for LVM
-	GiB                int64 = 1024 * 1024 * 1024
-	KiB                int64 = 1024
+	cacheSuffix                = "csi-fast"
+	mainLvSuffix               = "csi-main"
+	raidedLocalSsdName         = "csi-driver-data-cache"
+	raidMode                   = "0"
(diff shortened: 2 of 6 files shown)


pkg/gce-pd-csi-driver/cache.go
```diff
@@ -204,10 +217,16 @@ func cleanupCache(volumeId string, nodeId string) error {
 
 	volumeGroupName := getVolumeGroupName(nodeId)
 	if !checkVgExists(volumeGroupName) {
+		klog.V(4).Infof("Volume group %s not found, no cache clean up needed", volumeGroupName)
```
sunnylovestiramisu, line 220: I saw that some logs are V(2) and some are V(4), how did we decide this level?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -366,3 +396,18 @@ func isCachingSetup(mainLvName string) (error, bool) {
 	}
 	return nil, false
 }
+
```
sunnylovestiramisu, line 399: This can be unit tested?

Sneha-at: Done, we do not have invalid test cases as the input in pre-validated and I don't think it's needed to throw an error in this function.

pkg/gce-pd-csi-driver/cache.go
Sneha-at: Updated to match other logs - V(4)

pkg/gce-pd-csi-driver/cache_test.go
```diff
@@ -0,0 +1,57 @@
...
+			cacheSize:    "100Ki",
+			expChunkSize: "160KiB", //range defined in fetchChunkSizeKiB - min 160 KiB
+		},
+		{
+			name:         "invalid cacheSize",
+			cacheSize:    "fdfsdKi",
+			expChunkSize: "160KiB", //range defined in fetchChunkSizeKiB - min 160 KiB
+			expErr:       true,
+		},
+		// cacheSize is validated in storage class parameter so assuming invalid cacheSize (like negative, 0) would not be passed to the function
```
sunnylovestiramisu, line 40: Does cacheSize in StorageClass has the >=0 check? In the past there were bugs in Hyperdisk SC parameters that actually take in < 0 params. Though an error will be returned from GCE.

pkg/common/utils.go
```diff
@@ -713,6 +713,13 @@ func ValidateDataCacheMode(s string) error {
 	return fmt.Errorf("invalid data-cache-mode %s. Only \"writeback\" and \"writethrough\" is a valid input", s)
 }
 
+func ValidateDataCacheSize(size int64) error {
```
sunnylovestiramisu, line 716: This can be a more generic function like ValidateNonNegativeInt, it can be used in a few other places. "Input should be set to > 0, got %d".

sunnylovestiramisu: And unit tests.

mattcary:
The k8s integration failure is only the known snapshot flake, I'll force-merge

PR #1965, opened 2025-02-28
CVE-2022-1996

mattcary:
```release-note
Fix CVE-2022-1996
```


mattcary:
The suggested upgrade does not compile, abandoning in favor of automated dependabot run.

PR #1975, opened 2025-03-05
Relax volumeContentSource restriction for ROX multi-zone dynamic volume creation

pwschuurman:
This enables empty ROX multi-zone PVCs to be created, to allow for static hydration of disks

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

**What this PR does / why we need it**: Currently ROX multi-zone volumes can only be created from a volume source (eg: snapshot/disk image). This limits the flexibility of a user that wants to create a ROX multi-zone volume which is empty, and then write the same data individually to each created disk. This allows all disks to share the same name (as they are part of a "multi-zone" volume), but have data populated without needing to pre-create a snapshot/disk image (as copying data directly to each individual disk in parallel can be faster than creating and restoring from a snapshot).

Fixes #

```release-note
Relax multi-zone volume ROX restriction when creating a volume without a volumeContentSource.
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -232,7 +232,7 @@ var (
 		"nextPageToken",
 	}
 	listDisksFieldsWithUsers      = append(listDisksFieldsWithoutUsers, "items/users")
-	disksWithModifiableAccessMode = []string{"hyperdisk-ml"}
+	disksWithModifiableAccessMode = []string{common.DiskTypeHdML}
 	disksWithUnsettableAccessMode = map[string]bool{
 		common.DiskTypeHdE: true,
 		common.DiskTypeHdT: true,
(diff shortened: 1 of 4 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -485,12 +486,19 @@ func (gceCS *GCEControllerServer) createMultiZoneDisk(ctx context.Context, req *
 	}
 	defer gceCS.volumeLocks.Release(volumeID)
 
+	// If creating an empty disk (content source nil), always create RWO disks (when supported)
+	// This allows disks to be created as underlying RWO disks, so they can be hydrated.
```
dannawang0221, line 490: This means in volume populator we need to update the disk accessMode?

pwschuurman: No, accessMode is updated when the disk is attached (in controller publish), based on the accessMode of the PV that is using it: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/193aa66666ffba45fee3e3d60ac40e7b9453987f/pkg/gce-pd-csi-driver/controller.go#L1167

So this allows for:
1) An empty RWX PV/PVC to be created with underlying disks in RWO mode
2) A static RWO PV/PVC to be created, with the volumeHandle from (1) to hydrate the disk in RWO mode
3) A consumer to use the RWX PV/PVC from (1) after data has been hydrated

dannawang0221:
lgtm

PR #1983, opened 2025-03-06
Introduce Disk Topology Feature

julianKatz:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR introduces a new, flag-gated feature: Disk Topology. 

GKE will be adding new labels to its nodes that indicate the GCP disk products that each node is compatible with.

With the addition of these new disk type labels to nodes, we can include disk type labels in workload scheduling.  To affect that, those labels will be included in a PersistentVolume's `NodeAffinity` clause.  This PR includes a disk support label consistent with the `type` specified in the StorageClass in each of the Topology segments included in `CreateVolumeResponse`.  External Provisioner will then read these and add them as NodeAffinity requirements on the PV.

The feature is gated on the newly introduced `--disk-topology=[true|false]` flag, which is disabled by default.

**Special notes for your reviewer**: None.

Yes, users can set the flag to `true` and enable the new functionality.

```release-note
Added disk topology feature, enabled using --disk-topology=true (default off).  Disk topology includes support for `topology.gke.io/[disk-type]: true` labels on nodes and adds a label in that format to each CreateVolumeRequest.Volume.AccessibleTopology segment, which equate to NodeAffinity blocks in the eventual Persistent Volume.
```


pkg/gce-pd-csi-driver/gce-pd-driver_test.go
```diff
@@ -54,14 +54,28 @@ func controllerServerForTest(cloudProvider gce.GCECompute) *GCEControllerServer
...
+	return NewControllerServer(gceDriver, cloudProvider, errorBackoffInitialDuration, errorBackoffMaxDuration, fallbackRequisiteZones, enableStoragePools, enableDataCache, multiZoneVolumeHandleConfig, listVolumesConfig, provisionableDisksConfig, true /* enableHdHA */, args)
 }
 
-func initGCEDriverWithCloudProvider(t *testing.T, cloudProvider gce.GCECompute) *GCEDriver {
+func initGCEDriverWithCloudProvider(t *testing.T, cloudProvider gce.GCECompute, args ...*GCEControllerServerArgs) *GCEDriver {
 	vendorVersion := "test-vendor"
 	gceDriver := GetGCEDriver()
-	controllerServer := controllerServerForTest(cloudProvider)
+
+	// Passing args as a variadic argument prevents updating many function
```
mattcary, line 64: This comment won't make any sense once this PR is updated... there's only a dozen call sites, let's just update them all now  as initGCEDriverWithCloudProvider(t, provider, GCEControllerServerArgs{}).

It will simplify the logic below quite a bit which is a win for future generations.

pkg/gce-pd-csi-driver/node.go
```diff
@@ -557,21 +566,61 @@ func (ns *GCENodeServer) NodeGetCapabilities(ctx context.Context, req *csi.NodeG
...
 	}
+
+	klog.V(2).Infof("Returning NodeGetInfoResponse: %+v", resp)
+
 	return resp, err
 }
 
+// gkeTopologyLabels retrieves the node labels with the prefix
+// `topology.gke.io/` for the specified node.
+func (ns *GCENodeServer) gkeTopologyLabels(ctx context.Context, nodeName string) (map[string]string, error) {
```
mattcary, line 604: nit: this should be fetchGkeTopologyLabels or whatever to make it clear that its nontrivial.

mattcary (commented):
basically looks like the right direction.

you may have to do something to make sure the e2e tests work correctly -- I'm not sure what the did there for datacache.

pkg/nodelabels/verifier.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/client-go/kubernetes"
+	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
+)
+
+// resyncInterval is the interval at which the informer resyncs.  Should the
+// informer cache miss an event, the resync will correct the cache.
+const resyncInterval = 10 * time.Minute
```
tonyzhc, line 20: How is this resync interval determined? Does this mean that we can tolerate a maximum of 10 minute delay for pod scheduling in the case of Machine Serenity?

deploy/kubernetes/base/controller/cluster_setup.yaml
```diff
@@ -180,6 +180,9 @@ rules:
     verbs: ["use"]
     resourceNames:
       - csi-gce-pd-controller-psp
+  - apiGroups: [""]
```
tonyzhc, line 183: This file can be reverted right?

tonyzhc (approved):

pkg/nodelabels/verifier.go
mattcary: Resync is weird. It doesn't actually resync the informer from the API server, it just runs all objects in the informer through the handler. This was to deal with some obscure-to-me k8s race conditions that are no longer much of a problem.

PR #1989, opened 2025-03-13
Pass performance parameters when creating regional disks.

tonyzhc:
**What type of PR is this?**
> /kind bug

We found a bug during bugbash that performance parameters are not passed to `insertRegionalDisks`. I decided to consolidate the two variants of the function (`insertZonalDisks` and `insertRegionalDisks`), because they share almost the same logic except some non-applicable parameters (e.g. storage pool) and differences on the GCE interface. I believe that this will prevent future slips similar to this bug.

Fixes b/402880628

```release-note
NONE
```


nathanielford (commented):
LGTM!

I think this ends up being a lot more readable, and I didn't detect any changes in actual behavior. I think breaking up the disk spec construction and the insertion is a really good plan for keeping this clean going forward!

PR #1990, opened 2025-03-13
Add exponential backoff retries for getting Node from API server logic

hungnguyen243:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Add exponential backoff retries for getting Node from API server logic in GetDataCacheCountFromNodeLabel(). After we exceed the retries and still cannot get the Node object, the error will be surfaced for a driver restart (instead of the current logic that allows NodeStageVolume to happen and another error to surface misleadingly).
This is needed since we may hit connectivity issue with API server during driver startup, causing the existing code to fail to get the Node object with just one attempt.

Fixes #

```release-note
Add exponential backoff retries for getting Node from API server logic
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -245,11 +245,15 @@ func handle() {
 		if err != nil {
 			klog.Fatalf("Failed to set up metadata service: %v", err.Error())
 		}
+		isDataCacheEnabledNodePool, err := isDataCacheEnabledNodePool(ctx, *nodeName)
+		if err != nil {
+			klog.Fatalf("Failed to get node info from API server: %v", err.Error())
+		}
 		nsArgs := driver.NodeServerArgs{
 			EnableDeviceInUseCheck:   *enableDeviceInUseCheck,
 			DeviceInUseTimeout:       *deviceInUseTimeout,
 			EnableDataCache:          *enableDataCacheFlag,
-			DataCacheEnabledNodePool: isDataCacheEnabledNodePool(ctx, *nodeName),
+			DataCacheEnabledNodePool: isDataCacheEnabledNodePool,
 		}
 		nodeServer = driver.NewNodeServer(gceDriver, mounter, deviceUtils, meta, statter, nsArgs)
 		if *maxConcurrentFormatAndMount > 0 {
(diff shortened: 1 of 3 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -245,11 +245,15 @@ func handle() {
 		if err != nil {
 			klog.Fatalf("Failed to set up metadata service: %v", err.Error())
 		}
+		isDataCacheEnabledNodePool, err := isDataCacheEnabledNodePool(ctx, *nodeName)
+		if err != nil {
+			klog.Fatalf("Failed to get node info from API server: %v", err.Error())
```
sunnylovestiramisu, line 250: This implementation we will not get signal from alerts for get node info from API server? But I saw the above is doing the same for metadataservice. Is this expected? @mattcary

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -251,7 +253,7 @@ func GetDataCacheCountFromNodeLabel(ctx context.Context, nodeName string) (int,
 	if err != nil {
 		return 0, err
 	}
-	node, err := kubeClient.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	node, err := getNodeWithRetry(ctx, kubeClient, 6, time.Duration(1)*time.Second, nodeName) // maxDelay = 31s
 	if err != nil {
 		// We could retry, but this error will also crashloop the driver which may be as good a way to retry as any.
```
mattcary, line 258: why is this not already crashlooping the driver? Or is the crashlooping a problem?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -264,10 +266,27 @@ func GetDataCacheCountFromNodeLabel(ctx context.Context, nodeName string) (int,
 		klog.V(4).Infof("Number of local SSDs requested for Data Cache: %v", dataCacheCount)
 		return dataCacheCount, nil
 	}
-	// This will be returned for a non-Data-Cache node pool
 	return 0, nil
 }
 
+func getNodeWithRetry(ctx context.Context, kubeClient *kubernetes.Clientset, maxRetries int, initialDelay time.Duration, nodeName string) (*v1.Node, error) {
```
mattcary, line 272: use k8s wait instead, 	"k8s.io/apimachinery/pkg/util/wait", eg

```
err := wait.PollUntilContextTimeout(ctx, 2*time.Second, time.Minute, true, func(ctx context.Context) (bool, error) {
   success, err := doSomething()
   if err != nil {
     return false, err  // this will return an error no matter what the bool is
  } 
  return success, nil // this will retry if success is false, or return with no err if success is true
})
// if the timeout is succeeded err will be non-nil.

mattcary: There are other variations if you want different backoff and such but I don't think that matters in this context.

cmd/gce-pd-csi-driver/main.go
mattcary: I'm not sure what you mean here? err != nil if the api call fails.

pkg/gce-pd-csi-driver/cache.go
hungnguyen243: We want to retry if the API returns an error, but looks like the above code will only retry if a false value is returned, so we'd need to do some extra wrapping for the API call?

cmd/gce-pd-csi-driver/main.go
sunnylovestiramisu: But it only logged as fatal and continued, so what if the isDataCacheEnabledNodePool is always broken?

pkg/gce-pd-csi-driver/cache.go
hungnguyen243: This comment was actually incorrect. Before this PR, the error will be caught in the caller, which returns a bool value, and therefore actually didn't trigger crashlooping.

mattcary: ah! good idea.

cmd/gce-pd-csi-driver/main.go
mattcary: klog.Fatalf kills the program, no? https://pkg.go.dev/k8s.io/klog/v2#Fatalf

PR #1993, opened 2025-03-14
Fix logic bug while checking available LSSDs for RAIDing for Data Cache

hungnguyen243:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Fix logic bug while checking available LSSDs for RAIDing for Data Cache
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -368,24 +368,14 @@ func fetchLssdsForRaiding(lssdCount int) ([]string, error) {
 		return nil, fmt.Errorf("Error listing RAIDed LSSDs %v", err)
 	}
 
-	unRaidedLssds := []string{}
-	for _, l := range allLssds {
-		if !slices.Contains(raidedLssds, l) {
-			unRaidedLssds = append(unRaidedLssds, l)
-		}
-		if len(unRaidedLssds) == lssdCount {
-			break
-		}
-	}
-
 	LSSDsWithEmptyMountPoint, err := driver.FetchLSSDsWihtEmptyMountPoint()
 	if err != nil {
 		return nil, fmt.Errorf("Error listing LSSDs with empty mountpoint: %v", err)
 	}
 
 	// We need to ensure the disks to be used for Data Cache are both unRAIDed & not containing mountpoints for ephemeral storage already
-	availableLssds := slices.Filter(nil, unRaidedLssds, func(e string) bool {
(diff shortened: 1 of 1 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -368,24 +368,14 @@ func fetchLssdsForRaiding(lssdCount int) ([]string, error) {
...
 	LSSDsWithEmptyMountPoint, err := driver.FetchLSSDsWihtEmptyMountPoint()
 	if err != nil {
 		return nil, fmt.Errorf("Error listing LSSDs with empty mountpoint: %v", err)
 	}
 
 	// We need to ensure the disks to be used for Data Cache are both unRAIDed & not containing mountpoints for ephemeral storage already
-	availableLssds := slices.Filter(nil, unRaidedLssds, func(e string) bool {
-		return slices.Contains(LSSDsWithEmptyMountPoint, e)
+	availableLssds := slices.Filter(nil, allLssds, func(e string) bool {
+		return slices.Contains(LSSDsWithEmptyMountPoint, e) && !slices.Contains(raidedLssds, e)
```
Sneha-at, line 378: Can you add a comment on what we are filtering and what cases will that cater to

sunnylovestiramisu: For long term, we need some unit test coverage for these steps. :/

cmd/gce-pd-csi-driver/main.go
```diff
@@ -395,7 +385,8 @@ func fetchLssdsForRaiding(lssdCount int) ([]string, error) {
 	if len(availableLssds) < lssdCount {
 		return nil, fmt.Errorf("Not enough LSSDs available to set up caching. Available LSSDs: %v, wanted LSSDs: %v", len(availableLssds), lssdCount)
 	}
-	return availableLssds, nil
+
+	return availableLssds[:lssdCount], nil
```
sunnylovestiramisu, line 389: So if lssdCount == len(availableLssds), this will return an empty slice? Is this the intention?

hungnguyen243: I think this will return the first 'lssdCount' elements in the slice, which is what we need? So say lssdCount = 2, and we have more than 2 elements in availableLssds, this will just return the first 2 elemnts of availableLssds.

Issue #1782, opened 2024-07-09
gce-pd-driver container OOM killed after upgrade to GKE 1.28


JordanP:
Hi,
You probably have not control over how the `pd-csi` daemonset is deployed on GKE, but I am taking my chance.

After upgrading to GKE 1.28, some `gce-pd-driver` containers started to be OOM killed. Before getting killed, the last log line is `Checking for issues with fsck on disk: /dev/disk/by-id/google-restore-aus-southeast1-fcb9-pg-data-pg-main-0-7279`. That disk is a multi TB disk attached to a pod. My guess is 50MB (resources.memory.limit set by GKE for that  `gce-pd-driver` container) is not enough to run `fsck` on such a large disk. 

Any chance you could reach to someone at GKE to increase that memory limit (although baseline usage, fsck excepted is ~10MB so 50MB seems reasonable)  ? Or how could I skip that `fsck` check ?

If that helps, this is part of my Go code (running elsewhere) that seems to, down the line, trigger the call to `fsck`.

```go
	// make the request to the api /metrics endpoint and handle the response
	req := clientset.
		CoreV1().
		RESTClient().
		Get().
		Resource("nodes").
		Name(nodeName).
		SubResource("proxy").
		Suffix("metrics")
	respBody, err := req.DoRaw(ctx)
	if err != nil {
		return errors.Errorf("failed to get stats from kubelet on node %s: with error %s", nodeName, err)
	}
```

Thanks !

(please, don't recommend I reach to my TAM at GCP, we don't have one haha)

JordanP:
Actually I think there's more to it but I can't figure it out yet
![image](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/assets/221981/5a99ccfa-5dc7-4450-abe1-9d968beecdef)
In the instance, it looks like only the spawned `fsck` process got SIGKILL but not the `gce-pd-driver` process, so it was able to make progress eventually.

JordanP:
I am starting to feel that there's a deeper issue. Let's take a step back and explain the context. I have a script that, every 30min:
* Create a new GCE disk from a disk snapshot
* Attach the disk to a bare K8s pod
```py
    volumes=[
        V1Volume(
            name="pg-data",
            gce_persistent_disk=V1GCEPersistentDiskVolumeSource(fs_type="ext4", pd_name=disk_name),
        )
    ],
```
[...]

PR #2001, opened 2025-03-19
feat: enable btrfs

### LINKED ISSUE #2000: btrfs support
Currently `gcp-compute-persistent-disk-csi-driver` lacks `btrfs` support. A few interesting issues for us:

- transparent compression via `compress=zstd`
- transparent checksums
- snapshots
- `btrfs send` / `btrfs receive`

Adding btrfs support would not impact existing users.

It is [not explicitly supported in GCP docs](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/gce-pd-csi-driver), and it does not have to be, at least in the interim.

motiejus:
This change is enough to provision and run a btrfs filesystems.

Tested workflows:
[x] provisioning (formatting)
[x] disk expansion

/kind feature

Fixes #2000

Enables formatting new PVCs using `btrfs` filesystem.

Mount options like `compress=zstd` work as expected.

**Does this PR introduce a user-facing change?**: YES
```release-note
ubuntu-only: specifying `fsType: btrfs` in the PVC definition now formats & mounts a btrfs filesystem. btrfs support is experimental.
```

Non-btrfs users are not impacted by this change.

It is [not explicitly supported in GCP docs](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/gce-pd-csi-driver), and it does not have to be, at least in the interim. This PR allows users easily experiment with the FS.

The integration tests are not updated to run btrfs, because as of writing [COS does not support btrfs](https://cloud.google.com/container-optimized-os/docs/concepts/supported-filesystems).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ FROM gke.gcr.io/debian-base:bookworm-v1.0.4-gke.3 AS debian
 
 # Install necessary dependencies
 # google_nvme_id script depends on the following packages: nvme-cli, xxd, bash
-RUN clean-install util-linux e2fsprogs mount ca-certificates udev xfsprogs nvme-cli xxd bash kmod lvm2 mdadm
+RUN clean-install util-linux e2fsprogs mount ca-certificates udev xfsprogs nvme-cli xxd bash kmod lvm2 mdadm btrfs-progs
 
 # Since we're leveraging apt to pull in dependencies, we use `gcr.io/distroless/base` because it includes glibc.
 FROM gcr.io/distroless/base-debian12 AS distroless-base
@@ -48,6 +48,7 @@ COPY --from=debian /etc/mke2fs.conf /etc/mke2fs.conf
 COPY --from=debian /lib/udev/scsi_id /lib/udev_containerized/scsi_id
 COPY --from=debian /bin/mount /bin/mount
 COPY --from=debian /bin/umount /bin/umount
+COPY --from=debian /bin/btrfs /bin/btrfs
 COPY --from=debian /sbin/blkid /sbin/blkid
 COPY --from=debian /sbin/blockdev /sbin/blockdev
 COPY --from=debian /sbin/dumpe2fs /sbin/dumpe2fs
@@ -131,6 +132,7 @@ COPY --from=debian /lib/${LIB_DIR_PREFIX}-linux-gnu/libselinux.so.1 \
                    /lib/${LIB_DIR_PREFIX}-linux-gnu/libnvme.so.1 \
                    /lib/${LIB_DIR_PREFIX}-linux-gnu/libsystemd.so.0 \
                    /lib/${LIB_DIR_PREFIX}-linux-gnu/libgpg-error.so.0 \
+                   /lib/${LIB_DIR_PREFIX}-linux-gnu/liblzo2.so.2 \
                    /lib/${LIB_DIR_PREFIX}-linux-gnu/libzstd.so.1 /lib/${LIB_DIR_PREFIX}-linux-gnu/
 
 COPY --from=debian /usr/lib/${LIB_DIR_PREFIX}-linux-gnu/libblkid.so.1 \
--- a/README.md
+++ b/README.md
@@ -92,6 +92,12 @@ As part of the deployment process, the driver is deployed in a newly created nam
 
 Controller-level and node-level deployments will both have priorityClassName set, and the corresponding priority value is close to the maximum possible for user-created PriorityClasses.
 
+## Notes on filesystems
+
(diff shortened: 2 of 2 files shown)


sunnylovestiramisu:
Hmm so supporting btrfs resizing on Hyperdisk volumes requires creating a smaller volume, migrating data, and deleting the original. This seems like a half supported feature for PDCSI driver, I am not sure if we want to have it enable.  

/assign @mattcary

mattcary:
/ok-to-test

Since this is just adding another option to the storage class and not changing any of the control flow, I'm inclined to go ahead and enable it. Although if resizing is not working the tests probably need to be updated to skip them -- let's see what the CI does.

mattcary:
@motiejus can you update the README with details of support? Including any quirks around usage, like resize as Sunny mentioned (I haven't verified any of this myself).

motiejus:
Those failing tests seem like flakes? Windows-2019 and Windows-2022.

To  @mattcary 's comment
>  Including any quirks around usage

Early in testing we realized that the default `50MiB` RAM limit is not enough to `mount` large volumes (1TiB+). So we bumped the limit to 256MiB internally and didn't have any issues since. It is different from #1782, since that was the `mount` call that caused the OOM (the linked ticket is `fsck`).

Given that the memory limits are not part of this repository, do you think this warrants an entry in the README?

mattcary:
I think updating the readme to say that btrfs is available, but it's new and has not been production-tested. Things like increased memory needs would also be helpful.

motiejus:
> I think updating the readme to say that btrfs is available, but it's new and has not been production-tested. Things like increased memory needs would also be helpful.

Added a note in the README. If you prefer other wording, please let me know how, so I'll update it, or feel free to change it yourself.

Thanks!

mattcary:
> > I think updating the readme to say that btrfs is available, but it's new and has not been production-tested. Things like increased memory needs would also be helpful.
> 
> Added a note in the README. If you prefer other wording, please let me know how, so I'll update it, or feel free to change it yourself.
> 
> Thanks!

That looks great, thanks

motiejus:
> Hmm, it didn't look like the integration test ran with your sc-btrfs. I'm not sure why, can you take a look?

Tried to take a look. `tests/run-k8s-integration-ci.sh` wasn't executed, thus xfs tests also didn't run. _Seems_ like we have a bigger problem, but I don't know where to check: I can't find an entrypoint to the integration tests: `.github` directory does not have anything related.

Are we missing a big chunk of integration tests? Anyhow, this seems to be a bigger problem than btrfs alone.

motiejus:
The integration test failed with this:

```

  Mounting arguments: -t btrfs -o defaults /dev/disk/by-id/google-pvc-85a0c313-7a48-4daa-988a-9a4a5aeeabfb /var/lib/kubelet/plugins/kubernetes.io/csi/pd.csi.storage.gke.io/747f7ed65f8237a1726b5c15909179a6908fa3157bc2bfaf1aaf06a4e0362f27/globalmount
  Output: mount: /var/lib/kubelet/plugins/kubernetes.io/csi/pd.csi.storage.gke.io/747f7ed65f8237a1726b5c15909179a6908fa3157bc2bfaf1aaf06a4e0362f27/globalmount: unknown filesystem type 'btrfs'.
         dmesg(1) may have more information after failed mount system call. 
```

`cos_containerd` does not have a btrfs kernel module (we also stumbled upon this internally). I am changing to `ubuntu_containerd` for the sake of the test.

motiejus:
Integration test doesn't like non-COS:

> setting environment vars for bringing up *ubuntu* cluster on GCE is unimplemented

i will add a note in the README that CSI does not currently have btrfs kernel driver, thus only Ubuntu is supported (from the list of GCP-blessed OS images). If the OCI team asks, I would appreciate if you could support the clause.

mattcary:
> 
> i will add a note in the README that CSI does not currently have btrfs kernel driver, thus only Ubuntu is supported (from the list of GCP-blessed OS images). If the OCI team asks, I would appreciate if you could support the clause.

We've started the process internally, we'll see what happens.

> Tried to take a look. tests/run-k8s-integration-ci.sh wasn't executed, thus xfs tests also didn't run. 
>

If you go to the prow job yaml from the build details page, you see the following:

image

So it's the run-integration-test.sh script that's executed for presubmit. The -ci script is run periodically to make the following: https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver. Sorry, I misremembered what was what.

Anyway it looks like for the presubmit we only run on one storage class anyway, to save presubmit time. That's probably a good idea, ie we wouldn't want to add btrfs to the presumbit anyway. Thank you for arranging a one-off run to confirm the COS issue.

Since we've listed this support as experimental, and the presubmit run does confirm that enabling btrfs doesn't break anything else (not that we would have expected to but that's what testing is for....), I think we can submit this.

Thanks for your contribution!

/lgtm
/approve

motiejus:
> @motiejus are you on slack? I have a couple questions to ask. Thanks!

Sent you an email to your @google address.

PR #2004, opened 2025-03-19
Add pageToken response check to ListSnapshots gRPC API

### LINKED ISSUE #2003: ListSnapshots memory leak
Repro (In GKE):
 * Create a VolumeSnapshotContent with a valid snapshotHandle, but which never becomes ready (eg: snapshot creation request fails)
 * csi-snapshotter will periodically call ListSnapshots, which has a memory leak when calling GCE.
 
Problem:
 * The `nextPageToken` should be set at the end of the loop. Currently the loop continues until the client context times out the RPC call. Memory leak here:
   * ListSnapshots: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/daac2c6b8b3936c03aaa57e1f412f6a7b4e52ae7/pkg/gce-cloud-provider/compute/gce-compute.go#L311
  * ListImages: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/daac2c6b8b3936c03aaa57e1f412f6a7b4e52ae7/pkg/gce-cloud-provider/compute/gce-compute.go#L1306

pwschuurman:
**What type of PR is this?**
/kind bug

Fixes memory leak in #2003, by breaking the loop when paginated results are complete.

Fixes #2003

```release-note
Fix infinite loop in ListSnapshots that can cause memory usage to grow unbounded
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -315,6 +315,7 @@ func (cloud *CloudProvider) ListSnapshots(ctx context.Context, filter string) ([
 			return nil, "", err
 		}
 		items = append(items, snapshotList.Items...)
+		nextPageToken = snapshotList.NextPageToken
 	}
 	return items, "", nil
 }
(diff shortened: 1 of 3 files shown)


mattcary:
Failing on the known flaky snapshot tests. Force merging.

PR #2007, opened 2025-03-20
Fix units for cache size while calculating chunk size for LVM

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Currently the cache size unit isn't passed along with the cache size causing conversion issues downstream

Fixes #

```release-note
Fix unit calculation bug for cache size setup
```

### DIFF
--- a/pkg/gce-pd-csi-driver/cache.go
+++ b/pkg/gce-pd-csi-driver/cache.go
@@ -210,7 +210,7 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
 			req.GetPublishContext()[common.ContextDataCacheMode],
 			volumeGroupName + "/" + mainLvName,
 			"--chunksize",
-			chunkSize, // default unit is KiB
+			chunkSize,
 			"--force",
 			"-y",
 		}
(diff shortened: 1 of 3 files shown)


pkg/gce-pd-csi-driver/cache.go
```diff
@@ -563,7 +563,7 @@ func isCachingSetup(mainLvName string) (error, bool) {
 func fetchChunkSizeKiB(cacheSize string) (string, error) {
 	var chunkSize float64
 
-	cacheSizeInt, err := common.ConvertGiStringToInt64(cacheSize)
+	cacheSizeInt, err := strconv.Atoi(cacheSize)
```
sunnylovestiramisu, line 566: Atoi returns an int, do we want to return int64 instead?

Sneha-at: Updated this to `cacheSizeInt, err := strconv.ParseInt(cacheSize, 10, 64)`

sunnylovestiramisu:
publishContext[common.ContextDataCacheSize] = **defaultLocalSsdCacheSize** needs to be changed to a number?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -563,7 +563,7 @@ func isCachingSetup(mainLvName string) (error, bool) {
 func fetchChunkSizeKiB(cacheSize string) (string, error) {
 	var chunkSize float64
 
-	cacheSizeInt, err := common.ConvertGiStringToInt64(cacheSize)
+	cacheSizeInt, err := strconv.ParseInt(cacheSize, 10, 64)
```
sunnylovestiramisu, line 566: The comment below needs to be updated too? "Chunksize should be divisible by 32Kib so we need (chunksize/32*1024)*32*1024"?

Issue #1819, opened 2024-09-02
N4 instances have wrong attach limit


jsafrane:
N4 instances can use only hyperdisk-balanced disks and they have a different attach limit than regular "persistent disks": https://cloud.google.com/compute/docs/general-purpose-machines#disk_and_capacity_limits_2

The CSI driver should report 20 disks per n4-standard-2, now it reports 127.

sunnylovestiramisu:
This is fixed in master

/close

jsafrane:
/reopen

It's much better, just off-by-one :-).
The driver reports 16 attachments on n4-standard-4, but it forgets the root disk. Only 15 Pods with one volume each can run on the machine, 16th shows:

```
AttachVolume.Attach failed for volume "pvc-c1673c2a-7749-4b9b-b270-3544d9b04cef" : rpc error: code = InvalidArgument desc = Failed to Attach: failed when waiting for zonal op: rpc error: code = InvalidArgument desc = operation operation-1744802898292-632e394f4ad33-7301bb9b-ce176f79 failed (UNSUPPORTED_OPERATION): Maximum hyperdisk-balanced disks count should be less than or equal to [16], Requested : [17]
```

The driver reports 127 attachments for pd-disks, so it should report 15 for hyperdisks.

jsafrane:
Proposing a fix in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2057

PR #2019, opened 2025-03-25
Add Attach Limit for Hyperdisk + Gen4 VMs

sunnylovestiramisu:
**What type of PR is this?**
 /kind bug

Fixes # https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/1819

```release-note
Add Attach Limit for Hyperdisk + Gen4 VMs
```

/assign @mattcary

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -23,6 +23,7 @@ import (
 	"regexp"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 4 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -733,5 +741,24 @@ func (ns *GCENodeServer) GetVolumeLimits() (int64, error) {
 			return volumeLimitSmall, nil
 		}
 	}
+	gen4MachineTypesPrefix := []string{"c4a-", "c4-", "n4-"}
+	for _, gen4Prefix := range gen4MachineTypesPrefix {
+		if strings.HasPrefix(machineType, gen4Prefix) {
+			cpuString := machineType[strings.LastIndex(machineType, "-")+1:]
+			cpus, err := strconv.ParseInt(cpuString, 10, 64)
+			if err != nil {
+				return -1, fmt.Errorf("invalid cpuString %s for machine type: %v", cpuString, machineType)
```
sunnylovestiramisu, line 750: Should we return volumeLimitBig instead of -1 here? Because currently the err is not blocking.

mattcary: Since we know it's a gen4 probably should return the limit small to be conservative.

PR #2024, opened 2025-03-28
Bump csi proxy client version to 1.2.1

hajiler:
**What type of PR is this?**
/kind failing-test

Bumps csi proxy client version to 1.2.1. This will address flakiness in our windows testing.

Fixes #

```release-note
NONE
```


mauriciopoppe:
I think this PR won't change the flakiness given that the change that we want https://github.com/kubernetes-csi/csi-proxy/pull/364 is part of the server side component, not the client side which is what this PR is changing.

The server side version is changed in the k/k codebase, let's wait for https://github.com/kubernetes/kubernetes/pull/131146 to be merged and see https://testgrid.k8s.io/provider-gcp-compute-persistent-disk-csi-driver#ci-windows-2019-provider-gcp-compute-persistent-disk-csi-driver for results post-merge.

PR #2035, opened 2025-04-02
Update chunkSize to be calculated on total cache

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

It updates the logic to use total cache size instead of PV cache size to calculate the chunk size.

Fixes #

```release-note
Updates the logic to calculate chunk size for data cache
```

### DIFF
--- a/pkg/gce-pd-csi-driver/cache.go
+++ b/pkg/gce-pd-csi-driver/cache.go
@@ -175,10 +175,18 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
 		klog.V(4).Infof("Assuming valid data cache size and mode, resizing cache is not supported")
 	} else {
 		cacheSize := req.GetPublishContext()[common.ContextDataCacheSize]
-		chunkSize, err := fetchChunkSizeKiB(cacheSize)
+		maxChunkSizeStr := strconv.FormatFloat(maxChunkSize, 'g', -1, 64)
+		var chunkSize string
+		cachePvSize, err := fetchPvSize(raidedLocalSsdPath)
 		if err != nil {
-			klog.Errorf("Errored to fetch cache size, verify the data-cache-size is valid: got %v, error: %q", cacheSize, err)
-			return mainDevicePath, err
+			klog.Errorf("Errored while fetching PV size, got %v, falling back to default chunkSize of %v", err, maxChunkSize)
+			chunkSize = maxChunkSizeStr
+		} else {
+			chunkSize, err = fetchChunkSizeKiB(cachePvSize)
+			if err != nil {
+				klog.Errorf("Errored to fetch cache size, verify the data-cache-size is valid: got %v, error: %q", chunkSize, err)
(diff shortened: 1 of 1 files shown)


pkg/gce-pd-csi-driver/cache.go
```diff
@@ -639,6 +647,21 @@ func watchDiskDetaches(watcher *fsnotify.Watcher, nodeName string, errorCh chan
 			errorCh <- fmt.Errorf("disk update event errored: %v", err)
 		// watch for events
 		case event := <-watcher.Events:
+			args := []string{
+				"--updatemetadata",
+				getVolumeGroupName(nodeName),
```
sunnylovestiramisu, line 652: Why do we need these?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -175,10 +175,18 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+		maxChunkSizeStr := strconv.FormatFloat(maxChunkSize, 'g', -1, 64)
+		var chunkSize string
+		cachePvSize, err := fetchPvSize(raidedLocalSsdPath)
 		if err != nil {
-			klog.Errorf("Errored to fetch cache size, verify the data-cache-size is valid: got %v, error: %q", cacheSize, err)
-			return mainDevicePath, err
+			klog.Errorf("Errored while fetching PV size, got %v, falling back to default chunkSize of %v", err, maxChunkSize)
+			chunkSize = maxChunkSizeStr
+		} else {
+			chunkSize, err = fetchChunkSizeKiB(cachePvSize)
```
sunnylovestiramisu, line 185: The fetchChunkSizeKiB function itself is assuming that the input passing in (previously cacheSize now cachePvSize) is in GiB unit. But your fetchPvSize is returning KiB size already? Maybe the logic in fetchChunkSizeKiB also needs to change?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -674,3 +697,39 @@ func addRaidedLSSDToVg(vgName, lssdPath string) error {
...
+		"-o",
+		"--noheadings",
+		"pv_size",
+		"--units=b",
+	}
+	info, err := common.RunCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "pvs", args...)
+	if err != nil {
+		return "", fmt.Errorf("errored while fetching PV size %v: %s", err, info)
+	}
+	infoString := strings.TrimSpace(string(info))
```
sunnylovestiramisu, line 715: I would extract these string manipulation logic separately and add unit tests.

pkg/gce-pd-csi-driver/cache.go
Sneha-at: I think some stale got committed, realized this just now.

PR #2041, opened 2025-04-03
Use k/k from latest CI instead of build from HEAD.

### LINKED ISSUE #1955: It takes ~50 minutes to build kubernetes from source for Windows presubmits
Example: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_gcp-compute-persistent-disk-csi-driver/1951/pull-gcp-compute-persistent-disk-csi-driver-e2e-windows-2019/1894527667919654912 

The integration script is called with

```
+ /home/prow/go/src/sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/bin/k8s-integration-test ... --kube-version=master ...
```

Which together with the flag `flag.Bool("do-k8s-build", true, "building the driver from source. If false, will fetch precompiled artifacts")`

Forces to build k8s from scratch

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/39a59108472f7c9425b940ac9f1f062afae014d5/test/k8s-integration/main.go#L330-L335

Which takes ~50 minutes:

```
I0225 23:48:36.829509   17242 main.go:331] Building Kubernetes source
...
I0226 00:35:01.275869   17242 utils.go:15] Starting E2E Cluster on GCE
```

Do we need to build k/k from source or just take the latest CI builds? I think we should take the k/k CI builds, I forgot the reason we build from source.

tonyzhc:
**What type of PR is this?**
>
> /kind cleanup

This will reduce our windows CI run time by ~50min.

Fixes #1955

```release-note
NONE
```

### DIFF
--- a/test/run-windows-k8s-integration.sh
+++ b/test/run-windows-k8s-integration.sh
@@ -61,4 +61,5 @@ ${PKGDIR}/bin/k8s-integration-test \
     --snapshotclass-files=pd-volumesnapshotclass.yaml \
     --test-focus='External.Storage' \
     --use-kubetest2="${use_kubetest2}" \
-    --image-type="${windows_distribution}"
+    --image-type="${windows_distribution}" \
+    --do-k8s-build=false


sunnylovestiramisu:
The change seems resulting in windows presubmit tests not running anymore. Maybe that was the reason? @mauriciopoppe

mauriciopoppe:
The error says: `F0404 00:30:25.280270   18490 main.go:207] must set do-k8s-build=true when kubeVersion=master`, I think you could check which flags are set in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/test/run-windows-k8s-integration.sh

And how they're used in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/6aa4dc983569d1638e4a13d8c547a282a649d8d2/test/k8s-integration/main.go#L123

PR #2045, opened 2025-04-07
When using the extended channel, also specify the cluster version

julianKatz:
<**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake

When running CI for the `extended` channel, we intended to specify both the release channel _and_ the version of the cluster.  However, the CI test script was choosing only one of the two depending on the setting of the release channel.

This PR adds a special case of for `extended`, where both the release channel and the cluster version are passed to the test automation.

```release-note
None
```

### DIFF
--- a/test/run-k8s-integration-ci.sh
+++ b/test/run-k8s-integration-ci.sh
@@ -93,11 +93,18 @@ else
 fi
 
 if [ "$deployment_strategy" = "gke" ]; then
-  if [ "$gke_release_channel" ]; then
+  # To leverage old cluster versions, we need set both the release channel and
+  # the cluster version.
+  if [ "$gke_release_channel" = "extended" ]; then
+    base_cmd="${base_cmd} --gke-release-channel=${gke_release_channel}"
+    base_cmd="${base_cmd} --gke-cluster-version=${gke_cluster_version}"
+  elif [ -n "$gke_release_channel" ]; then
     base_cmd="${base_cmd} --gke-release-channel=${gke_release_channel}"
   else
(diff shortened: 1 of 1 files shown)


test/run-k8s-integration-ci.sh
```diff
@@ -93,11 +93,18 @@ else
 fi
 
 if [ "$deployment_strategy" = "gke" ]; then
-  if [ "$gke_release_channel" ]; then
+  # To leverage old cluster versions, we need set both the release channel and
+  # the cluster version.
+  if [ "$gke_release_channel" = "extended" ]; then
```
mattcary, line 98: Does this work if $gke_release_channel is empty? I'm always a little confused by bash parsing, but if the string is empty you may end up with if [ = extended ] which is a parse error

julianKatz: Good catch, I did a little more looking and I believe I've got a better version now.

That first part is right (gke_release_channel is defaulted to empty string in its var declaration).  Stackoverflow says `[ thing = otherThing ]` is [the right POSIX solution](https://superuser.com/a/1076805).

The second one was more ambiguous.  I've changed it to check for a non-empty variable:

```
  elif [ -n "$gke_release_channel" ]; then
  ```

mattcary: ah, the quotes matter. [ $unknown = foo ] fails to parse, but [ "$unknown" = foo ] is okay. Interesting.

+1 to using -n, it's a little more clear (although [ ] seems to be valid anyway)

PR #2048, opened 2025-04-08
Add missing dependencies found using hack/verify-docker-deps.sh

julianKatz:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The Dockerfile was missing image dependencies that prevented it from successfully handling some system-level disk operations.

```release-note
None
```


msau42 (commented):
Should we run verify-docker-deps.sh as a presubmit check?

Or is there any presubmit test that could detect that this is missing?

julianKatz:
> Should we run verify-docker-deps.sh as a presubmit check?
> 
> Or is there any presubmit test that could detect that this is missing?

It's part of the verify check in `master`.  Will see if it is here also.

julianKatz:
Looks like it wasn't part of verify-all.sh on this branch.  Not sure why, as it is on `master`.

I'll have a look through other branches to see if it needs to be added in those other places.

julianKatz:
Seems like I'm getting real test failures here.  Need to investigate.

image

PR #2054, opened 2025-04-11
Fix Hyperdisk Resize That Requires Iops/Throughput Adjustment

sunnylovestiramisu:
**What type of PR is this?**
/kind bug

Currently resizing for Hyperdisk is failing for cases that require IOPS or throughput getting updated in the same resize call.

A [hyperdisk](https://cloud.google.com/compute/docs/disks/hyperdisks)-balanced with size 4Gi needs exactly 2000 IOPS, and 5Gi needs exactly 2500 IOPS. Disks 6Gi and more need a minimum of 3000 IOPS. When resizing a hyperdisk, there is a need to update IOPS as well. 

For hyperdisk-extreme the min IOPS value is 2 IOPS per GiB, if a 64 GiB hyperdisk-extreme disk with 128 IOPS is resized to 65 GiB, the IOPS also need to be updated to 130.

For hyperdisk-ml it is more complicated, min IOPS is 16 IOPS per MBps of provisioned throughput, min throughput is the greater of 0.12 MBps per GiB and 400 MBps per disk. A 3334GiB hyperdisk-ml with 400 MBps throughput and initial 6400 IOPS resize to larger capacity will also need to update both throughput and IOPS.

Fixes #

Tested end to end in GCE clusters.

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-de7f9208-7fff-fe16-b9ef-bc7618c4d1a4"><div dir="ltr" style="margin-left:0pt;" align="left">
Type | From Size to Size | Notes
-- | -- | --
hyperdisk-balanced | 4 Gib to 5Gi | n4-standard-4 IOPS: 2000 -> 2500 Throughput: 140MB/s(no change)
hyperdisk-balanced | 5Gi to 20Gi | n4-standard-4 IOPS: 2500->3000 Throughput: 145MB/s(no change)
hyperdisk-balanced-high-availability | 5Gi to 6Gi | c3-standard-4 IOPS: 2500->3000 Throughput: 145MB/s(no change)
hyperdisk-extreme | 64Gi to 70Gi | IOPS: 128 to 140
hyperdisk-throughput | 2Ti to 3Ti | c3-standard-4 Throughput: 20MB/s to 30MB/s
hyperdisk-ml | 3334Gi to 3500Gi | c3-standard-4 Throughput: 400MB/s to 420MB/s

</div></b>

```release-note
Fix Hyperdisk Resize That Requires Iops/Throughput Adjustment
```

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -30,6 +30,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
 	"github.com/googleapis/gax-go/v2/apierror"
 	"golang.org/x/time/rate"
+	computev1 "google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -80,6 +81,12 @@ const (
 	//   projects/{project}/zones/{zone}
 	zoneURIPattern = "projects/[^/]+/zones/([^/]+)$"
 	alphanums      = "bcdfghjklmnpqrstvwxz2456789"
+
+	HyperdiskBalancedIopsPerGB         = 500
+	HyperdiskBalancedMinIops           = 3000
+	HyperdiskExtremeIopsPerGB          = 2
+	HyperdiskThroughputThroughputPerGB = 10
+	BytesInGB                          = 1024
 )
 
 var (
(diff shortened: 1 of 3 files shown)


mattcary:
This should be easy to add an e2e test for the presubmit?

Let's submit this now though and add e2e later (which I think would be preferable to an internal GKE test).

The e2e test could just be for the simple HdB case, testing thoughput and such gets harder with machine compatibility. 

/lgtm
/approve

PR #2081, opened 2025-05-07
GKE-MT Support for PDCSI

cemakd:
**What type of PR is this?**
/kind feature

Adds capability to the GCE cloud provider for PDCSI to be GKE-MT aware.

N/A
Fixes #

n/a
```release-note
NONE
```


pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -148,6 +156,13 @@ func CreateCloudProvider(ctx context.Context, vendorVersion string, configPath s
 		return nil, err
 	}
 
+	// token, err := tokenSource.Token()
```
mattcary: remove debugging code

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -160,40 +160,67 @@ func (cloud *CloudProvider) listDisksInternal(ctx context.Context, fields []goog
...
+		klog.Infof("Getting regional disks for project: %s", p)
+		rDisks, err := listRegionalDisksForProject(s, p, region, fields, filter)
+		if err != nil {
+			return nil, "", err
+		}
+		disks = append(disks, rDisks...)
+	}
+
+	// listing out zonal disks in all zones of the region for each project
+	for p, s := range cloud.tenantServiceMap {
```
mattcary, line 188: It looks like this map is populated only if there's a tenant informer? What happens for non-multitenant clusters?

pkg/gce-cloud-provider/compute/gce.go
cemakd: Done

pkg/gce-cloud-provider/compute/gce-compute.go
cemakd: Thanks for catching this, this was from an earlier iteration I had where I had put the default project as the first entry into the `tenantServiceMap` with the idea that we can replace all usages of cloud.service with a for loop that goes over each project in tenantServiceMap.

I later scrapped this idea but forgot to revert this change here.

PR #2089, opened 2025-05-12
Only add disk support topologies if StorageClass parameter is true

julianKatz:
**What type of PR is this?**
/kind feature

This PR makes the disk topology feature only add a disk support label to PV NodeAffinity _if_ a new StorageClass parameter is specified: `use-allowed-disk-topology: true`.

The benefit of this is that GKE can enable the `--disk-topology=true` flag without immediately changing PDCSI's behavior.  The feature becomes available, but not enabled.  Customers can then choose to enable the feature when they are ready, plus they gain the granularity of enabling it for specific disk products at a time.  Since most customers need this for Hyperdisk usage, that allows them to target the functionality to their use case.

I've manually tested this feature.  See the steps recorded [here](https://docs.google.com/document/d/13Mh5kMiTeOp4UwOKtRIG7SCUba8V8werVmI9OUuU-zA/edit?resourcekey=0-RVtQpwoT4dL4m-4BZEKQrQ&tab=t.0).

Fixes #

```release-note
Only add disk support topologies if StorageClass parameter is true.  --disk-topology=true flag will no longer be sufficient to invoke the functionality.
```


mattcary:
This should have a release note as it adds a user-visible feature.

tonyzhc:
Can you add an e2e test in test/e2e/tests/single_zone_e2e_test.go as well?

sunnylovestiramisu:
Is this at least manually end to end tested on GCE clusters? Maybe add the test cases you did in the description.

julianKatz:
> This should have a release note as it adds a user-visible feature.

Is this more than what's contained in the PR description?  I have the `release-note` label and I'm looking for PRs that show changes to a separate file but not finding any.

julianKatz:
test run for comparison: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_gcp-compute-persistent-disk-csi-driver/2089/pull-gcp-compute-persistent-disk-csi-driver-e2e/1925356347314933760

julianKatz:
> Is this at least manually end to end tested on GCE clusters? Maybe add the test cases you did in the description.

@sunnylovestiramisu @tonyzhc I have added an e2e test in the zonal file and it is passing

mattcary:
> > This should have a release note as it adds a user-visible feature.
> 
> Is this more than what's contained in the PR description? I have the `release-note` label and I'm looking for PRs that show changes to a separate file but not finding any.

Release notes get pulled out automatically for the change log so they're nice to have.

PR #2096, opened 2025-05-20
Fix Gen4 Custom VM Cases

sunnylovestiramisu:
**What type of PR is this?**
/kind bug

Custom VM naming does not follow the current string handling, it returns an error in NodeGetInfo and breaks customers.

Fixes #

```release-note
Fix Gen4 Custom VM Cases
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -656,7 +656,9 @@ func (ns *GCENodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRe
 
 	volumeLimits, err := ns.GetVolumeLimits(ctx)
 	if err != nil {
-		klog.Errorf("GetVolumeLimits failed: %v", err.Error())
+		klog.Errorf("GetVolumeLimits failed: %v. The error is ignored so that the driver can register", err.Error())
+		// No error should be returned from NodeGetInfo, otherwise the driver will not register
+		err = nil
 	}
 
 	resp := &csi.NodeGetInfoResponse{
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -654,10 +654,8 @@ func (ns *GCENodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRe
 
 	nodeID := common.CreateNodeID(ns.MetadataService.GetProject(), ns.MetadataService.GetZone(), ns.MetadataService.GetName())
 
+	// err is always nil
 	volumeLimits, err := ns.GetVolumeLimits(ctx)
-	if err != nil {
```
mattcary, line 658: Why remove this logging? Even if the error is eventually plumbed through an rpc or whatever, it will be easier to find it in the driver logs.

sunnylovestiramisu: The logging is moved to the inner function klog.Warningf("invalid cpuString %s for machine type: %v", cpuString, machineType) below.

This is the only case that was returning an error, which it should not return the error because it makes the driver not being able to register: https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_plugin.go#L149-L154

mattcary: ah I see, thx.

So we will still return the error & prevent registration if GetVolumeLimits returns an error. Do we still want to do that, or return default volume limits if something goes awry?

The latter might be a good idea, because the volume limits only apply to a corner case of workloads. So if something else comes up in the future we may not want to fail registration.

sunnylovestiramisu: The GetVolumeLimits(which is the only place in func NodeGetInfo will return an error) never returned an error until we added the support for different attach limit for Gen4 VM.

And within the GetVolumeLimits, only strconv.ParseInt below and GetAttachLimitsOverrideFromNodeLabel returns an error. But error from GetAttachLimitsOverrideFromNodeLabel is swallowed, and will continue to use the default.

Then the question is for ParseInt, do we want to treat the error as "absolutely we should not register the driver or not". If the naming format of GCE VMs do not change, we covered the following formats right now:

* "x4-megamem-960-metal"
* "a4-highgpu-8g"
* "c3d-highmem-8-lssd"
* "n4-custom-8-12345-ext"
* "n4-custom-16-12345"
* "n4-standard-8"

Let say later GCE provides a new VM does not follow the naming convention we assumed, we will need to backport the support again. But the difference is this time they can use the nodelabels to override until we support it.

 
Return default volume limits if something goes awry -> this is what we do until now, do we want to change this behavior?

pkg/gce-pd-csi-driver/node.go
```diff
@@ -843,10 +841,12 @@ func (ns *GCENodeServer) GetVolumeLimits(ctx context.Context) (int64, error) {
 	gen4MachineTypesPrefix := []string{"c4a-", "c4-", "n4-"}
 	for _, gen4Prefix := range gen4MachineTypesPrefix {
 		if strings.HasPrefix(machineType, gen4Prefix) {
-			cpuString := machineType[strings.LastIndex(machineType, "-")+1:]
+			machineTypeSlice := strings.Split(machineType, "-")
+			cpuString := machineTypeSlice[2]
```
sunnylovestiramisu, line 845: This line will panic with runtime error if one day they decided to have a VM n4-micro.

pkg/gce-pd-csi-driver/node.go
sunnylovestiramisu: Thinking a bit more, if the strconv.ParseInt throws an error, it is probably a case that the machine type name is not what we think it should be. But this VM is still created by GCE, do we want to block PDCSI driver to be registered on this VM or it is better to register and then customers can still create volume etc. even the attach limit is not accurate?

mattcary: As discussued offline, if we block registration then we make the VM unusable until we push out a bug fix. That's not so good.

So I think better to have inaccurate limits, especially because most customers aren't affected by the limits.

PR #2097, opened 2025-05-21
Cache and print devices

julianKatz:
changes and printing the full list.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


pkg/linkcache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+// it is ignored.
+func (d *linkCache) BrokenSymlink(symlink string) {
+	if entry, ok := d.devices[symlink]; ok {
+		entry.brokenSymlink = true
+		d.devices[symlink] = entry
+	}
+}
+
+func (d *linkCache) RemoveDevice(symlink string) {
+	delete(d.devices, symlink)
```
tonyzhc, line 152: Should we log here as well? Will that information be valuable?

pkg/linkcache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+
+	var errs []error
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+
+		// TODO(juliankatz): To have certainty this works for all edge cases, we
+		// need to test this with a manually partitioned disk.
+		if partitionNameRegex.MatchString(entry.Name()) {
```
tonyzhc, line 83: For my understanding, could you provide more context on why we skip partitions?

pkg/linkcache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (l *ListingCache) listAndUpdate() error {
+	visited := make(map[string]struct{})
+
+	entries, err := os.ReadDir(l.dir)
+	if err != nil {
+		return fmt.Errorf("failed to read directory %s: %w", l.dir, err)
+	}
+
+	var errs []error
```
tonyzhc, line 75: Maybe split out the rest of the function and make unit tests, so that you avoid dependency on os.ReadDir.

pkg/linkcache/cache.go
```diff
@@ -0,0 +1,173 @@
...
+		d.devices[symlink] = entry
+	}
+}
+
+func (d *linkCache) RemoveDevice(symlink string) {
+	delete(d.devices, symlink)
+}
+
+func (d *linkCache) DeviceIDs() []string {
+	ids := make([]string, 0, len(d.devices))
```
tonyzhc, line 156: Can just be [maps.Keys](https://pkg.go.dev/golang.org/x/exp/maps#Keys)

pkg/linkcache/cache.go
julianKatz: For the boot disks there will be something like 12 partitions.  Those end up being noise when we are answering the question of "which PV maps to which path on the VM".

pkg/linkcache/cache.go
julianKatz: Done

pkg/linkcache/cache.go
julianKatz: Turns out you can just make a mock filesystem.  Added unit test with that.

pkg/linkcache/cache.go
julianKatz: Good call.  DOne.

pkg/linkcache/cache.go
msau42: Add it as a comment in the code to make it easier for future readers

pkg/linkcache/cache.go
```diff
@@ -0,0 +1,193 @@
...
+				klog.Warningf("Error listing and updating symlinks: %v", err)
+				continue
+			}
+
+			klog.Infof("periodic symlink cache read: %s", l.links.String())
+		}
+	}
+}
+
+func (l *ListingCache) listAndUpdate() error {
```
msau42, line 88: How do normal attach/detach flows show up with this? When we log, do we have a way to distinguish between normal attach/detach sequence vs when the device mapping unexpectantly changes post-mount?

I'm concerned about normal flows creating too much noise in the logs and drowning out the real issues.

pkg/linkcache/cache.go
julianKatz: Moved to different model that doesn't have this problem

test/remote/instance.go
```diff
@@ -235,7 +235,7 @@ func (i *InstanceInfo) CreateOrGetInstance(localSSDCount int) error {
 		}
 
 		if i.cfg.CloudtopHost {
```
julianKatz, line 237: This seems to no longer be necessary on cloudtop, and was actively causing my local test run to fail.

sunnylovestiramisu:
Not under active development anymore.

PR #2112, opened 2025-06-12
default data cache enabled to false

Sneha-at:
**What type of PR is this?**
>
 /kind bug

We recently had an issue where data cache related watcher was running on non-data cache nodes as we were defaulting to `true` so updating the default values here to `false`

Fixes #

```release-note
Disabling data cache watcher by default if node details are not available.
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -270,7 +270,7 @@ func handle() {
 		if err != nil {
 			klog.Fatalf("Failed to set up metadata service: %v", err.Error())
 		}
-		isDataCacheEnabledNodePool, err := isDataCacheEnabledNodePool(ctx, *nodeName)
+		isDataCacheEnabledNodePool, err := driver.IsDataCacheEnabledNodePool(ctx, *nodeName, *enableDataCacheFlag)
 		if err != nil {
 			klog.Fatalf("Failed to get node info from API server: %v", err.Error())
 		}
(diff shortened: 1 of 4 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -382,7 +382,7 @@ func isDataCacheEnabledNodePool(ctx context.Context, nodeName string) (bool, err
 		dataCacheLSSDCount, err := driver.GetDataCacheCountFromNodeLabel(ctx, nodeName)
 		return dataCacheLSSDCount != 0, err
 	}
-	return true, nil
+	return false, nil
```
sunnylovestiramisu: Should we also add *enableDataCacheFlag and len(nodeName) > 0 && nodeName != common.TestNode above?

This logic does not seem correct for gating component versions with the enableDataCacheFlag.

Should we have some unit tests for this coverage?

sunnylovestiramisu: Actually expanded the code above the flag check was ahead.

sunnylovestiramisu: Any other possibility that the nodeName maybe empty other than the bug that we already knew?

pwschuurman: Don't we want data cache to be enabled for E2E testing? I think there's 3 scenarios here:
 * nodeName empty (false)
 * nodeName is testnode (true)
 * nodeName is set (check LSSD from label)

Also, would it be good to consolidate the enable/disable logic somewhere (eg: in a feature flags class), so that all codepaths are routed through the same enablement behavior? I'm thinking about the NodeUnstage case, where we aren't checking if this is a data cache enabled node.

Sneha-at: Moved the function to utils and added some unit test case. I have also added logic to perform this check during UnStageVolume

pkg/gce-pd-csi-driver/utils.go
```diff
@@ -339,3 +339,17 @@ func containsZone(zones []string, zone string) bool {
...
 }
+
+func IsDataCacheEnabledNodePool(ctx context.Context, nodeName string, enableDataCacheFlag bool) (bool, error) {
+	if !enableDataCacheFlag {
+		return false, nil
+	}
+	if nodeName == common.TestNode {
+		return true, nil
+	}
+	if len(nodeName) > 0 { // disregard logic below when E2E testing.
```
pwschuurman: Is this comment still applicable here? It seems like it should be moved up to line 347.

Sneha-at: Thanks! Done.

PR #2113, opened 2025-06-13
Release 1.20 cut details for v1.20.0

cemakd:
/kind feature

Put 1.20.0 release into stable-master.

```release-note
NONE
```


mattcary:
/lgtm
/approve

You'll need to rebase though it looks like.

Also, it seems that we're missing the 1.19 and 1.20 changelogs? IIRC these should be in the master now and not release branches.

PR #2114, opened 2025-06-17
Added support for NVME Hyperdisks on Windows Nodes

ivanLeont:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR fixes the ability to use hyperdisks/nvme drives on window's nodes. This works by having the CSI node query Google's API to extract the correct serial number of the hyperdisk and match it to the correct index/serial of the mounted disk on windows.

Fixes # N/A

```release-note
NONE
```

### DIFF
--- a/pkg/resizefs/resizefs_windows.go
+++ b/pkg/resizefs/resizefs_windows.go
@@ -21,7 +21,6 @@ package resizefs
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	volumeapiv1 "github.com/kubernetes-csi/csi-proxy/client/api/volume/v1"
 	volumeapiv1beta1 "github.com/kubernetes-csi/csi-proxy/client/api/volume/v1beta1"
@@ -73,11 +72,6 @@ func (resizefs *resizeFs) resizeV1(devicePath string, deviceMountPath string) (b
 	}
 	_, err = proxy.VolumeClient.ResizeVolume(context.Background(), request)
 	if err != nil {
-		// Check if this is the Windows error indicating partition is already at correct size
-		if strings.Contains(err.Error(), "The size of the extent is less than the minimum of 1MB") {
-			klog.V(3).Infof("Partition is already at target size (extent difference < 1MB), treating as success: %v", err)
-			return false, nil // Return false to indicate no resize was needed, but no error
-		}
 		return false, err
 	}
 	return true, nil
@@ -103,11 +97,6 @@ func (resizefs *resizeFs) resizeV1Beta(devicePath string, deviceMountPath string
 	}
 	_, err = proxy.VolumeClient.ResizeVolume(context.Background(), request)
 	if err != nil {
-		// Check if this is the Windows error indicating partition is already at correct size
-		if strings.Contains(err.Error(), "The size of the extent is less than the minimum of 1MB") {
-			klog.V(3).Infof("Partition is already at target size (extent difference < 1MB), treating as success: %v", err)
-			return false, nil // Return false to indicate no resize was needed, but no error
-		}
 		return false, err
 	}
 	return true, nil


ivanLeont:
Adding a note here: This still works even if the GKE cluster has Workload Identity, as the Daemonset kube-system/gke-metadata-server that manages metadata access, seems to allow the SA pdcsi-node-sa (which is used by the pdcsi-node-windows Daemonset) to bypass all metadata restrictions, which compared to a user created SA has more restricted access to the metadata. I have confirmed this by creating a pod that uses the SA pdcsi-node-sa and was able to access the necessary disk data.

mattcary:
> Adding a note here: This still works even if the GKE cluster has Workload Identity, as the Daemonset kube-system/gke-metadata-server that manages metadata access, seems to allow the SA pdcsi-node-sa (which is used by the pdcsi-node-windows Daemonset) to bypass all metadata restrictions, which compared to a user created SA has more restricted access to the metadata. I have confirmed this by creating a pod that uses the SA pdcsi-node-sa and was able to access the necessary disk data.

Hmm, super interesting, this isn't expected. Can you give details of your setup? gcloud compute instances describe as well as kubectl get -o yaml for both the node and the pd csi daemonset deployment you use with this PR?

Thx.

ivanLeont:
> > Adding a note here: This still works even if the GKE cluster has Workload Identity, as the Daemonset kube-system/gke-metadata-server that manages metadata access, seems to allow the SA pdcsi-node-sa (which is used by the pdcsi-node-windows Daemonset) to bypass all metadata restrictions, which compared to a user created SA has more restricted access to the metadata. I have confirmed this by creating a pod that uses the SA pdcsi-node-sa and was able to access the necessary disk data.
> 
> Hmm, super interesting, this isn't expected. Can you give details of your setup? gcloud compute instances describe as well as kubectl get -o yaml for both the node and the pd csi daemonset deployment you use with this PR?
> 
> Thx.

Here are the details of the setup:


[gke-metadata-server.txt](https://github.com/user-attachments/files/20843398/gke-metadata-server.txt)
[cd4-node.txt](https://github.com/user-attachments/files/20843399/cd4-node.txt)
[cluster-info.txt](https://github.com/user-attachments/files/20843400/cluster-info.txt)
[csi-gce-pd-node-win.yaml.txt](https://github.com/user-attachments/files/20843401/csi-gce-pd-node-win.yaml.txt)
[pdcsi-node.yaml.txt](https://github.com/user-attachments/files/20843402/pdcsi-node.yaml.txt)



When deploying a cluster with Workload Identity enabled, i noticed that after running ./deploy-drivers.sh, the windows node images deployed by the DS `csi-gce-pd-node-win`  were crashing with the following error message

```
│ I0620 20:06:02.254719    5792 main.go:122] Operating compute environment set to: production and computeEndpoint is set to: <nil>                                                                                                                                        │
│ I0620 20:06:02.255761    5792 main.go:131] Sys info: NumCPU: 4 MAXPROC: 1                                                                                                                                                                                               │
│ I0620 20:06:02.255761    5792 main.go:136] Driver vendor version a9a2d90593f36050b197c7576b2970e8420b99e9                                                                                                                                                               │
│ I0620 20:06:02.257324    5792 safe-mounter_windows.go:66] using CSIProxyMounterV1, API Versions Disk: v1, Filesystem: v1, Volume: v1                                                                                                                                    │
│ F0620 20:06:02.267481    5792 main.go:271] Failed to set up metadata service: failed to get machine-type: metadata: GCE metadata "instance/machine-type" not defined  
```
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/cmd/gce-pd-csi-driver/main.go#L271

I then enabled the default CSI and noticed that the windows image there was not crashing at that point, so I created two pods in both namespaces with different SA's

This pod, `metadata-test`, deployed in ns `gce-pd-csi-driver` uses the sa `csi-gce-pd-node-sa-win` which is is created using the ./deploy-driver.sh (I added the required bindings and annotations to work with Workload Identity)
```
PS C:\> Invoke-RestMethod -Uri 'http://metadata.google.internal/computeMetadata/v1/instance' -Headers @{'Metadata-Flavor'='Google'}
attributes/
hostname
id
name
service-accounts/
zone


PS C:\> Invoke-RestMethod -Uri 'http://metadata.google.internal/computeMetadata/v1/instance/service-accounts' -Headers @{'Metadata-Flavor'='Google'}
default/
sa_number@project_id.iam.gserviceaccount.com/


PS C:\> 

```

In this pod, `metadata-test`, deployed in ns `kube-system` uses the sa `pdcsi-node-sa`  which comes in by default when enabling the default CSI

```
PS C:\> Invoke-RestMethod -Uri 'http://metadata.google.internal/computeMetadata/v1/instance/' -Headers @{'Metadata-Flavor'='Google'}
attributes/
cpu-platform
credentials/
description
disks/
gce-workload-certificates/
guest-attributes/
hostname
id
image
licenses/
machine-type
maintenance-event
name
network-interfaces/
partner-attributes/
preempted
remaining-cpu-time
scheduling/
service-accounts/
shutdown-details/
tags
virtual-clock/
zone

PS C:\> Invoke-RestMethod -Uri 'http://metadata.google.internal/computeMetadata/v1/instance/service-accounts' -Headers @{'Metadata-Flavor'='Google'}
240471850237-compute@developer.gserviceaccount.com/
default/

PS C:\> 
``` 

So when I used a webhook to swap the original image in the DS kube-system/pdcsi-node-windows my custom image was able to work fine. It was also observed that in the gke-metadata-server DS, there is a field `--passthrough-ksa-list` that has `kube-system:pdcsi-node-sa` as one of the SA's allowd passthrough

Let me know if you need any more details! 😃

mattcary:
>  It was also observed that in the gke-metadata-server DS, there is a field --passthrough-ksa-list that has kube-system:pdcsi-node-sa as one of the SA's allowd passthrough

ahhh interesting we didn't know the security team did that :-P

Let us figure out if this is the right way to go, then

julianKatz:
TLDR: Let's merge this, excluding the changes in resizefs_windows.go

1. GKE has pursued a fix for the resize issue separate of this PR.  That has seen a multi-step investigation.

1. Let's remove the "error skipping" logic in resizefs_windows.go from this PR.  It's more work to re-validate this PR than rely on existing work as far as the resize issue is concerned.

1. I've made a version of this PR _without_ the changes to `resizefs_windows.go` and tested them on a C3 family Windows machine.  That connects a disk via NVMe.  The pod got to `Running`, signifying a successful mounting of the disk and the solving of the issue.

1. If the windows tests pass for this PR, that serves as validation that the change does not break Windows SCSI support.  Currently, the windows tests (inherited from k/k) are run solely on a GCE machine type that uses SCSI to connect to disks.

1. Thus, our next step should be to remove those "error skip" clauses in resizefs_windows.go from this PR and merge it.

msau42:
/approve

Thank you for your contribution and thorough investigation!

arsiesys:
> TLDR: Let's merge this, excluding the changes in resizefs_windows.go
> 
> 1. GKE has pursued a fix for the resize issue separate of this PR.  That has seen a multi-step investigation.
> 2. Let's remove the "error skipping" logic in resizefs_windows.go from this PR.  It's more work to re-validate this PR than rely on existing work as far as the resize issue is concerned.
> 3. I've made a version of this PR _without_ the changes to `resizefs_windows.go` and tested them on a C3 family Windows machine.  That connects a disk via NVMe.  The pod got to `Running`, signifying a successful mounting of the disk and the solving of the issue.
> 4. If the windows tests pass for this PR, that serves as validation that the change does not break Windows SCSI support.  Currently, the windows tests (inherited from k/k) are run solely on a GCE machine type that uses SCSI to connect to disks.
> 5. Thus, our next step should be to remove those "error skip" clauses in resizefs_windows.go from this PR and merge it.

Hi @julianKatz !
Do you have a link to the change that would handle the resize differently ? This PR is not usable alone without the fix for the resize.

```
  Normal   SuccessfulAttachVolume  5m27s  attachdetach-controller  AttachVolume.Attach succeeded for volume "xxxx"
  Warning  FailedMount             5m18s  kubelet                  MountVolume.MountDevice failed for volume "xxxx" : rpc error: code = Internal desc = error when resizing volume projects/xxxx/zones/us-east4-a/disks/xxxx from device '1' at path '\var\lib\kubelet\plugins\kubernetes.io\csi\pd.csi.storage.gke.io\236fde50928689809686339669c30afad62b9776b7e9c28a415ec66a8c05ee6a\globalmount': rpc error: code = Unknown desc = error resizing volume. cmd: Get-Volume -UniqueId "$Env:volumeID" | Get-Partition | Resize-Partition -Size 107357388288, output: Resize-Partition : Size Not Supported\r
 \r
Extended information:\r
The size of the extent is less than the minimum of 1MB.\r
 \r
Activity ID: {0083c501-d947-47dc-bf00-8c910e706fb4}\r
At line:1 char:56\r
+ ... "$Env:volumeID" | Get-Partition | Resize-Partition -Size 107357388288\r
+                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\r
    + CategoryInfo          : NotSpecified: (StorageWMI:ROOT/Microsoft/.../MSFT_Partition) [Resize-Partition], CimExce \r
   ption\r
    + FullyQualifiedErrorId : StorageWMI 4097,Resize-Partition\r
 \r
 size:0, finalSize 107357388288, error: exit status 1
 ```

tonyzhc:
Hi @arsiesys, could you tell me the node machine type and example pv/pvc/pod manifest of when you see this error? In our internal testing we see that the error is temporary and the mount should succeed. The event may stick around for longer but that should be a red herring.

arsiesys:
Hi @tonyzhc , I just retested and you are correct, it will end up being mounted. I wasn't patient enough to let it pull the images and assumed the "FailedMount" in the kubelet to be a fatal and didn't let it go through the end.
Sorry for the noise !

arsiesys:
The 'FailedMount" Event from kubelet, though it may work eventually, could benefit from more user-friendly error handling. The original PR's approach provided clearer feedback, which would help users understand what's happening and reduce the back-and-forth needed to resolve issues.

tonyzhc:
@arsiesys the concern was that hardcoding the error may cover other cases we haven't considered. But I agree with you, we will investigate how to make the error not show up in the first place. Thanks for your patience!

PR #2116, opened 2025-06-25
Remove check on snapshot create support for multi-writer HyperDisk-HA

sumanthsuresh1117:
Let such validations occur at the PD control plane layer as the support will evolve/change going forward and it's better for the CSI-driver layer to not test such supportability.

**What type of PR is this?**
bug

Removes an unnecessary and potentially incorrect check that multi-writer Hyperdisk doesn't support snapshot. Hyperdisk is in the process of supporting this functionality and the CSI-driver must defer such checks to the PD control plane.

```release-note
NONE
```


msau42:
/approve

As a followup, should we add an e2e test that takes a snapshot of a hyperdisk-ha disk?

tonyzhc:
> /approve
> 
> As a followup, should we add an e2e test that takes a snapshot of a hyperdisk-ha disk?

Filed b/427994669

PR #2118, opened 2025-06-27
Added filtering for incorrect slo boosting logs

chrisThePattyEater:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Recently we were out of SLO for the PDCSI CreateSnapshot endpoint. After analysis of the errors contributing to our slo we have identified the issue to be an excessive amount of INTERNAL errors. Further analysis of these internal errors has brought to light two major types of errors that are being incorrectly tagged as internal. These should not be contributing to our slos, they are the following:
Already exists errors should be converted from INTERNAL to ALREADY_EXISTS
Org policy violations should be converted from INTERNAL to FAILED_PRECONDITION

These errors are both returned from the create snapshot gcp endpoint

This change introduces methods that use regex to identify these two types of errors in the response payload and correctly marks them as ALREADY_EXISTS and FAILED_PRECONDITION errors.

Fixes #
https://b.corp.google.com/issues/398008774

```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"runtime"
 	"sync"
 	"time"
@@ -75,6 +76,11 @@ const (
 	// gcpTagsRequestTokenBucketSize is the burst/token bucket size used
 	// for limiting API requests.
 	gcpTagsRequestTokenBucketSize = 8
+)
+
+var (
+	ssAlreadyExistsRegex = regexp.MustCompile("The resource [^']+ already exists")
+	gcpViolationRegex    = regexp.MustCompile("violates constraint constraints/gcp.")
 )
 
 // ResourceType indicates the type of a compute resource.
@@ -440,4 +446,14 @@ func IsGCENotFoundError(err error) bool {
 // invalid reason
 func IsGCEInvalidError(err error) bool {
 	return IsGCEError(err, "invalid")
+}
+
+// IsSnapshotAlreadyExistsError checks if the error is a snapshot already exists error.
+func IsSnapshotAlreadyExistsError(err error) bool {
+	return ssAlreadyExistsRegex.MatchString(err.Error())
+}
+
+// IsGCPOrgViolationError checks if the error is a GCP organization policy violation error.
+func IsGCPOrgViolationError(err error) bool {
+	return gcpViolationRegex.MatchString(err.Error())
 }
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -79,8 +79,9 @@ const (
 )
 
 var (
-	ssAlreadyExistsRegex = regexp.MustCompile("The resource [^']+ already exists")
-	gcpViolationRegex    = regexp.MustCompile("violates constraint constraints/gcp.")
+	ssAlreadyExistsRegex          = regexp.MustCompile("The resource [^']+ already exists")
+	gcpErrorCodeRegex             = regexp.MustCompile(`Error (\d+)`)
+	orgPolicyConstraintErrorCodes = []string{"400", "412"}
 )
 
 // ResourceType indicates the type of a compute resource.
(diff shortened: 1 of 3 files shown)


pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -441,3 +442,33 @@ func IsGCENotFoundError(err error) bool {
 func IsGCEInvalidError(err error) bool {
 	return IsGCEError(err, "invalid")
 }
+
+// IsGCENotFoundError returns true if the error is a googleapi.Error with
+// notFound reason
+func ErrorContainsRegX(err error) bool {
```
cemakd: This method is unused right?

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1668,6 +1668,16 @@ func (gceCS *GCEControllerServer) createPDSnapshot(ctx context.Context, project
 			if gce.IsGCEError(err, "notFound") {
 				return nil, status.Errorf(codes.NotFound, "Could not find volume with ID %v: %v", volKey.String(), err.Error())
 			}
+
+			// Identified as incorrect error handling in https://b.corp.google.com/issues/398008774
+			if gce.IsSnapshotAlreadyExistsError(err) {
+				return nil, status.Errorf(codes.AlreadyExists, "Snapshot already exists: %v", err.Error())
+			}
+
+			// Identified as incorrect error handling in https://b.corp.google.com/issues/398008774
```
cemakd: No need to include Buganizer links in OSS, probably not useful for non-Googlers.

pkg/gce-cloud-provider/compute/gce.go
chrisThePattyEater: you are right :-)

pkg/gce-pd-csi-driver/controller.go
chrisThePattyEater: ack

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -441,3 +442,27 @@ func IsGCENotFoundError(err error) bool {
 func IsGCEInvalidError(err error) bool {
 	return IsGCEError(err, "invalid")
 }
+
+// ErrorContainsRegex checks if the error message contains a specific regex pattern.
+func ErrorContainsRegex(err error, regexPattern string) bool {
+	if err == nil {
+		return false
+	}
+	re, compileErr := regexp.Compile(regexPattern)
```
pwschuurman: Can you set re as a static var, with `regexp.MustCompile`?

pkg/gce-cloud-provider/compute/gce.go
```diff
@@ -441,3 +442,27 @@ func IsGCENotFoundError(err error) bool {
...
+}
+
+// IsSnapshotAlreadyExistsError checks if the error is a snapshot already exists error.
+func IsSnapshotAlreadyExistsError(err error) bool {
+	return ErrorContainsRegex(err, "The resource '[^']+' already exists")
+}
+
+// IsGCPOrgViolationError checks if the error is a GCP organization policy violation error.
+func IsGCPOrgViolationError(err error) bool {
+	return ErrorContainsRegex(err, "violates constraint constraints/gcp.")
```
pwschuurman: Is there a more complex error policy violation field in the response that you can check instead of string error message extraction?

The risk is if the error message from the API is reformatted, this regex matching may fail. If there is a more dependable API type response we could use, I would prefer to use that.

pkg/gce-cloud-provider/compute/gce_test.go
```diff
@@ -101,6 +103,114 @@ func TestIsGCEError(t *testing.T) {
...
+}
+
+func TestErrorIsSnapshotExistsError(t *testing.T) {
+	testCases := []struct {
+		name           string
+		inputErr       error
+		expectedResult bool
+	}{
+		{
+			name:           "is gcp org violation error",
```
pwschuurman: Test name does not match Test function.

pwschuurman (approved):

PR #2120, opened 2025-06-30
Add c4d to gen4 attach limit list

hajiler:
**What type of PR is this?**
/kind bug

For GCP Gen4 machine types, there are CPU dependent volume attach limits. This PR adds missing machine family c4d to this list of machine types. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node_test.go
+++ b/pkg/gce-pd-csi-driver/node_test.go
@@ -340,6 +340,11 @@ func TestNodeGetVolumeLimits(t *testing.T) {
 			machineType:    "c4a-standard-32-lssd",
 			expVolumeLimit: 49,
 		},
+		{
+			name:           "c4d-standard-32",
+			machineType:    "c4d-standard-32",
+			expVolumeLimit: 49,
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -340,6 +340,11 @@ func TestNodeGetVolumeLimits(t *testing.T) {
 			machineType:    "c4a-standard-32-lssd",
 			expVolumeLimit: 49,
 		},
+		{
+			name:           "c4d-standard-48",
```
tonyzhc: Change the name to be consistent with the machine type.

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -340,6 +340,11 @@ func TestNodeGetVolumeLimits(t *testing.T) {
...
+			machineType:    "c4d-standard-2",
```
tonyzhc: Can you use a machine type that will return a value other than `volumeLimitSmall` or `volumeLimitBig`? This will make the test case more robust.

hajiler: Done.

pkg/gce-pd-csi-driver/node_test.go
hajiler: Done.

PR #2126, opened 2025-07-08
Change Attach Limit to Hyperdisk Balanced Only

sunnylovestiramisu:
**What type of PR is this?**
/kind bug

Fixes #

```release-note
Change Attach Limit to Hyperdisk Balanced Only
```

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -70,10 +70,6 @@ var Gen4MachineHyperdiskAttachLimitMap = []struct {
 	max   int64
 	value int64
 }{
-	{max: 4, value: 15},
-	{max: 8, value: 23},
-	{max: 16, value: 31},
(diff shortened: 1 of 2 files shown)


arsiesys:
Hi,

I believe this fix may introduce side effects since each machine family has different limits. The current approach would only work properly for the N4 family. For better clarity and correctness, we should use different limit maps based on the machine type, with each map containing the specific limits for that family. I'll submit a separate PR that addresses this concern.

arsiesys:
As mentioned, I created this alternative PR:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2127

sunnylovestiramisu:
/close 

In favor of https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2127

PR #2127, opened 2025-07-09
fix: update Hyperdisk attach limits to match GCP documentation for Ge…

arsiesys:
**What type of PR is this?**

> /kind bug

This PR updates the Hyperdisk balanced attach limits to align with the latest GCP documentation for Gen4 machine families (`C4`, `C4D`, `N4`, and `C4A`).
Previously, a single generic map was used, which did not reflect the correct per-family attach limits for newer machine types.

Key changes:

* Adds separate Hyperdisk attach limit maps for `C4`, `C4D`, `N4`, and `C4A` machines.
* Updates the attach limit calculation logic to use the appropriate map based on the machine type prefix.
* Updates unit tests to cover new machine type mappings and verify correct limits.

This fixes inaccurate attach limits for Gen4 machines, ensuring compliance with the [[official GCP guidance](https://cloud.google.com/compute/docs/general-purpose-machines)](https://cloud.google.com/compute/docs/general-purpose-machines).

Fixes #

* Please verify the machine type prefixes and limit values match the official GCP tables.

```release-note
fix: Updated Hyperdisk balanced attach limits for C4, C4D, N4, and C4A machine families to match current GCP documentation.
```

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -65,15 +65,42 @@ const (
 	AttachLimitOverrideLabel   = "gke-volume-attach-limit-override"
 )
 
-// doc https://cloud.google.com/compute/docs/disks/hyperdisks#max-total-disks-per-vm
-var Gen4MachineHyperdiskAttachLimitMap = []struct {
+// doc https://cloud.google.com/compute/docs/general-purpose-machines
+// MachineHyperdiskLimit represents the mapping between max vCPUs and hyperdisk (balanced) attach limit
+type MachineHyperdiskLimit struct {
 	max   int64
 	value int64
-}{
+}
+
+// C4 Machine Types - Hyperdisk Balanced Limits
+var C4MachineHyperdiskAttachLimitMap = []MachineHyperdiskLimit{
+	{max: 2, value: 7},
 	{max: 4, value: 15},
-	{max: 8, value: 23},
-	{max: 16, value: 31},
-	{max: 32, value: 49},
-	{max: 64, value: 63},
-	{max: 1024, value: 127},
+	{max: 24, value: 31},
+	{max: 48, value: 63},
+	{max: 96, value: 127},
+}
+
+// C4D Machine Types - Hyperdisk Balanced Limits
+var C4DMachineHyperdiskAttachLimitMap = []MachineHyperdiskLimit{
+	{max: 2, value: 3},
+	{max: 4, value: 7},
+	{max: 8, value: 15},
+	{max: 96, value: 31},
+	{max: 192, value: 63},
+	{max: 384, value: 127},
+}
+
+// N4 Machine Types - Hyperdisk Balanced Limits
+var N4MachineHyperdiskAttachLimitMap = []MachineHyperdiskLimit{
+	{max: 8, value: 15},
+	{max: 80, value: 31},
+}
+
+// C4A Machine Types - Hyperdisk Balanced Limits
+var C4AMachineHyperdiskAttachLimitMap = []MachineHyperdiskLimit{
+	{max: 2, value: 7},
+	{max: 8, value: 15},
+	{max: 48, value: 31},
+	{max: 72, value: 63},
 }
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -762,14 +762,39 @@ func ShortString(s string) string {
 	return string(short)
 }
 
-// MapNumber is a function to map input cpu number to the Hyperdisk attach limit
-func MapNumber(num int64) int64 {
-	for _, r := range Gen4MachineHyperdiskAttachLimitMap {
-		if num <= r.max {
-			return r.value
+// GetHyperdiskAttachLimit returns the hyperdisk attach limit based on machine type prefix and vCPUs
+func GetHyperdiskAttachLimit(machineTypePrefix string, vCPUs int64) int64 {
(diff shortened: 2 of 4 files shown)


pkg/common/constants.go
```diff
@@ -65,15 +65,41 @@ const (
...
+var C4MachineHyperdiskAttachLimitMap = []MachineHyperdiskLimit{
+	{max: 2, value: 7},
 	{max: 4, value: 15},
-	{max: 8, value: 23},
 	{max: 16, value: 31},
-	{max: 32, value: 49},
-	{max: 64, value: 63},
-	{max: 1024, value: 127},
+	{max: 48, value: 63},
+	{max: 192, value: 127},
```
sunnylovestiramisu: Set this one higher, so that in the future if they have a higher max, we still have some value set. Otherwise we need to backport to earlier versions.

pkg/common/utils.go
```diff
@@ -762,13 +762,38 @@ func ShortString(s string) string {
...
+	for _, limit := range limitMap {
+		if vCPUs <= limit.max {
+			return limit.value
 		}
 	}
+	// Return the last value if vCPUs exceeds all max values
+	if len(limitMap) > 0 {
+		return limitMap[len(limitMap)-1].value
+	}
 	return 0
```
sunnylovestiramisu: Do not think we should return 0 here, maybe use the default volumeLimitSmall.

pkg/common/constants.go
arsiesys: Thanks ! The MapNumber function will take the last value of the map if no max match, is that correct or you still prefer to also have a high max in the list ?

pkg/common/utils.go
arsiesys: I can return a static 15 as the constant is declared in node.go.
Or, I could move the constant from node.go to the common package, let me know which path we prefer.

pkg/common/constants.go
sunnylovestiramisu: The MapNumber function has:
func MapNumber(num int64) int64 {
	for _, r := range Gen4MachineHyperdiskAttachLimitMap {
		if num <= r.max {
			return r.value
		}
	}
	return 0
}

Hmm maybe still we should just return the VolumeLimitSmall from the MapNumber itself

arsiesys: The MapNumber has been updated in this PR:

```
func MapNumber(vCPUs int64, limitMap []MachineHyperdiskLimit) int64 {
	for _, limit := range limitMap {
		if vCPUs <= limit.max {
			return limit.value
		}
	}
	return 0
	// Return the last value if vCPUs exceeds all max values
	if len(limitMap) > 0 {
		return limitMap[len(limitMap)-1].value
	}
	return 15
}
```

pkg/gce-pd-csi-driver/node.go
```diff
@@ -858,17 +858,19 @@ func (ns *GCENodeServer) GetVolumeLimits(ctx context.Context) (int64, error) {
...
 			}
 		}
-		if strings.HasPrefix(machineType, "x4-") {
-			return x4HyperdiskLimit, nil
-		}
-		if strings.HasPrefix(machineType, "a4-") {
-			return a4HyperdiskLimit, nil
-		}
+	}
+	if strings.HasPrefix(machineType, "x4-") {
```
sunnylovestiramisu, line 869: Why moved this outside? Because if it is inside the machineType is x4 or a4 it does not need to go through the loop. But on the other hand it does not matter so much in the real world

pkg/common/constants.go
```diff
@@ -65,15 +65,41 @@ const (
...
 	value int64
-}{
+}
+
+// C4 Machine Types - Hyperdisk Balanced Limits
+var C4MachineHyperdiskAttachLimitMap = []MachineHyperdiskLimit{
+	{max: 2, value: 7},
 	{max: 4, value: 15},
-	{max: 8, value: 23},
 	{max: 16, value: 31},
```
sunnylovestiramisu: This one should be max: 24 value:31

pkg/gce-pd-csi-driver/node.go
arsiesys: You're right! Moving these checks outside was what I though could be a performance optimisation.

Since x4/a4 machine types don't match any of the gen4 prefixes (c4a-, c4-, n4-, c4d-), keeping them inside the loop means we'd unnecessarily check them 4 times (once per iteration) before ultimately returning the correct limit.

By moving them outside:
- gen4 machines: Exit early when matched, never reaching these checks
- Other machines: Only check x4/a4 once at the end

This reduces string comparisons from up to 12 (4 iterations × 3 checks) to at most 6 (4 gen4 + x4 + a4).

I also considered that x4 and a4 may be less frequently used than the ones in the loop. I  will put it back if you think it would be better or not the subject of the current PR (would be understandable).

pkg/common/constants.go
arsiesys: 24 vcpu machine being > 4 and <= 16, it would already get 31 in value with the logic in the MapNumber unless I  misunderstood something.

arsiesys: Oh no, you right, I am tired.. Fixing

arsiesys: Pushed, thanks.
Also noticed a missing value for C4D:
	{max: 384, value: 127},

PR #2141, opened 2025-08-04
Cache and print devices for debugging future outages

cemakd:
This is a copy of PR: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2097

**What type of PR is this?**
/kind cleanup

This PR adds a cache that periodically (configured to every minute currently) looks at the /dev/disk/by-id/ directory and evaluates the symlinks there. It maintains a cache of the symlink and the real path it points to.

This will help with debugging future filesystem issues. In a past OMG, we found that our insight into changes in symlinks for specific disks hampered our ability to debug. Logging marked the real path of the disk at mount and unmount, but the change in between couldn't be detected.

This PR will print those links every minute, also logging when elements of the cache change.

An example:
```
I0521 00:23:51.418261      12 cache.go:62] periodic symlink cache read: /dev/disk/by-id/google-persistent-disk-0 -> /dev/sda; /dev/disk/by-id/google-pvc-f5418f78-dc07-4d69-9487-6c4a7232dd67 -> /dev/sdb; /dev/disk/by-id/scsi-0Google_PersistentDisk_persistent-disk-0 -> /dev/sda; /dev/disk/by-id/scsi-0Google_PersistentDisk_pvc-f5418f78-dc07-4d69-9487-6c4a7232dd67 -> /dev/sdb
```

The cache will also note if a symlink is broken.

NOTE: Currently this filters out any thing in by-id/ that ends with -part[0-9]*$. This removes partitions, which are noise. Mounting partitions directly isn't well supported in GKE, but we may want to test that in the future.

Fixes #

```release-note
NONE
```


pkg/linkcache/devices_linux.go
```diff
@@ -0,0 +1,153 @@
...
+func NewDeviceCacheForNode(ctx context.Context, period time.Duration, nodeName string) (*DeviceCache, error) {
+	node, err := k8sclient.GetNodeWithRetry(ctx, nodeName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+	}
+
+	return newDeviceCacheForNode(period, node), nil
+}
+
+func TestDeviceCache(period time.Duration, node *v1.Node) *DeviceCache {
```
msau42: Can we move the test functions to the _test.go file?

pkg/linkcache/devices_linux.go
```diff
@@ -0,0 +1,153 @@
...
+		volumes: make(map[string]deviceMapping),
+		period:  period,
+		dir:     byIdDir,
+	}
+
+	// Look at the status.volumesInUse field.  For each, take the last section
+	// of the string (after the last "/") and call AddVolume for that
+	for _, volume := range node.Status.VolumesInUse {
+		klog.Infof("Adding volume %s to cache", string(volume))
+		vID, err := pvNameFromVolumeID(string(volume))
```
msau42: Does this ignore volume types that are not PD/hyperdisk?

pkg/linkcache/devices_linux.go
```diff
@@ -0,0 +1,153 @@
...
+	klog.Infof("Removing volume %s from cache", volumeID)
+	delete(d.volumes, volumeID)
+}
+
+func (d *DeviceCache) listAndUpdate() {
+	for volumeID, device := range d.volumes {
+		// Evaluate the symlink
+		realPath, err := filepath.EvalSymlinks(device.symlink)
+		if err != nil {
+			klog.Warningf("Error evaluating symlink for volume %s: %v", volumeID, err)
```
msau42: Be careful of logging this since this is called repeatedly.

cmd/gce-pd-csi-driver/main.go
```diff
@@ -275,6 +276,13 @@ func handle() {
 			klog.Fatalf("Failed to get node info from API server: %v", err.Error())
 		}
 
+		deviceCache, err := linkcache.NewDeviceCacheForNode(ctx, 1*time.Minute, *nodeName)
```
msau42: Can we make the time period configurable so that we can adjust it if it ends up spamming too many logs? In fact I would probably start the logging at 10 mins initially. 

In the future, if we want to turn this into a metric, I would be more comfortable reducing the polling period but removing logs to avoid log spam.

pkg/linkcache/devices_linux.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("error converting volume ID to key: %w", err)
+	}
+	deviceName, err := common.GetDeviceName(volumeKey)
+	if err != nil {
+		return fmt.Errorf("error getting device name: %w", err)
+	}
+
+	// Look at the dir for a symlink that matches the pvName
+	symlink := filepath.Join(d.dir, "google-"+deviceName)
```
msau42: Has this been tested with nvme disks as well? It would be better to reuse functions in https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/deviceutils/device-utils.go

msau42 (commented):
Can you describe the testing that has been done for this change?

cemakd:
> Can you describe the testing that has been done for this change?

I haven't done any additional testing on this PR since I took it over from Julian. As far as I know he was struggling to get the e2e tests passing, which I fixed with a few nil pointers checks.

I'll do some more testing on this and get back to you

pkg/linkcache/devices_linux.go
cemakd: I haven't done specific testing for nvme disks but I've pushed a new commit that utilizes device-utils.go for evaluating the symlinks for a volumeID

cmd/gce-pd-csi-driver/main.go
cemakd: Done, added a flag `disk-cache-sync-period` with a default value of 10 minutes when unspecified.

pkg/linkcache/devices_linux.go
cemakd: Now that the polling period is 10 mins, would it be okay to log this? Otherwise it might be hard to figure out any problems with realpath evaluation?
I'm not sure if that's a realistic problem to log, I can remove this warning log if you think it's not necessary

pkg/linkcache/devices_linux.go
cemakd: It was not ignoring them, I've pushed a commit that checks if the volume is for the driver `pd.csi.storage.gke.io`. IIUC that should be sufficient to do the filtering.

pkg/linkcache/devices_linux.go
cemakd: These were not real test function but rather creating a DeviceCache for testing purposes used by other packages. I've renamed them to `NewTestDeviceCache` to distinguish them from actual test functions.

pkg/linkcache/devices_linux.go
cemakd: I've completed the testing for both NVMe and SCSI disk interface types.

I used the following machine types for the tests:

- SCSI: e2-standard-2
- NVMe: c3-standard-4

Here are the log outputs:

SCSI: The log shows both the google- and scsi- prefixed symlinks successfully resolving to the same device path (/dev/sdb).

Cache contents: 

```
map[
  /dev/disk/by-id/google-pvc-b08e725b-7bdf-499a-af8f-9bfba4c9ef69:{volumeID:projects/enginakdemir-joonix/zones/us-central1-c/disks/pvc-b08e725b-7bdf-499a-af8f-9bfba4c9ef69 realPath:/dev/sdb} 
  /dev/disk/by-id/scsi-0Google_PersistentDisk_pvc-b08e725b-7bdf-499a-af8f-9bfba4c9ef69:{volumeID:projects/enginakdemir-joonix/zones/us-central1-c/disks/pvc-b08e725b-7bdf-499a-af8f-9bfba4c9ef69 realPath:/dev/sdb}
]
```

NVMe: The log shows the google- prefixed symlink resolving to the NVMe device path (/dev/nvme0n2), while the scsi- symlink has an empty realPath.

Cache contents: 

```
map[
  /dev/disk/by-id/google-pvc-289e075f-e731-4e75-9519-4f0e5e4832ae:{volumeID:projects/enginakdemir-joonix/zones/us-central1-c/disks/pvc-289e075f-e731-4e75-9519-4f0e5e4832ae realPath:/dev/nvme0n2} 
  /dev/disk/by-id/scsi-0Google_PersistentDisk_pvc-289e075f-e731-4e75-9519-4f0e5e4832ae:{volumeID:projects/enginakdemir-joonix/zones/us-central1-c/disks/pvc-289e075f-e731-4e75-9519-4f0e5e4832ae realPath:}
]
```

pkg/linkcache/devices_linux.go
```diff
@@ -41,36 +42,37 @@ func TestNodeWithVolumes(volumes []string) *v1.Node {
...
-		period:  period,
-		dir:     byIdDir,
+		symlinks:    make(map[string]deviceMapping),
+		period:      period,
+		deviceUtils: deviceUtils,
+		dir:         byIdDir,
 	}
 
 	// Look at the status.volumesInUse field.  For each, take the last section
 	// of the string (after the last "/") and call AddVolume for that
```
msau42: Add a comment with what the expected syntax is

pkg/linkcache/devices_linux.go
```diff
@@ -41,36 +42,37 @@ func TestNodeWithVolumes(volumes []string) *v1.Node {
...
+		tokens := strings.Split(volumeName, "^")
+		if len(tokens) != 2 {
+			klog.V(5).Infof("Skipping volume %q because splitting volumeName on `^` returns %d tokens, expected 2", volumeName, len(tokens))
+			continue
+		}
+
+		// The first token is of the form "kubernetes.io/csi/<driver-name>" or just "<driver-name>".
+		// We should check if it contains the driver name we are interested in.
+		if !strings.Contains(tokens[0], driverName) {
+			klog.V(5).Infof("Skipping volume %q because it is not a %s volume.", volumeName, driverName)
```
msau42: This will be common if a customer has multiple csi drivers. We probably don't need this log.

pkg/linkcache/devices_linux.go
cemakd: Done

pkg/linkcache/devices_linux.go
cemakd: Removed this log line, done

PR #2144, opened 2025-08-06
Replace Get operations with polling

prachirp:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Replace Get operations with polling in gce-compute.go

Fixes #

**Does this PR introduce a user-facing change?**: NONE

```release-note
NONE
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -76,12 +76,12 @@ var AttachDiskBackoff = wait.Backoff{
 	Cap:      0}
 
 // WaitForOpBackoff is backoff used to wait for Global, Regional or Zonal operation to complete.
-// Default values are similar to Poll every 3 seconds with 5 minute timeout.
+// Default values are similar to Poll every 2 minutes with 6 minute timeout.
 var WaitForOpBackoff = wait.Backoff{
-	Duration: 3 * time.Second,
+	Duration: 2 * time.Minute,
 	Factor:   0.0,
 	Jitter:   0.0,
-	Steps:    100,
+	Steps:    3,
 	Cap:      0}
 
 // Custom error type to propagate error messages up to clients.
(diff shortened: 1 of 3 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -1066,7 +1066,7 @@ func (cloud *CloudProvider) getRegionalDiskTypeURI(project string, region, diskT
 func (cloud *CloudProvider) waitForZonalOp(ctx context.Context, project, opName string, zone string) error {
 	// The v1 API can query for v1, alpha, or beta operations.
 	return wait.ExponentialBackoff(WaitForOpBackoff, func() (bool, error) {
```
tonyzhc, line 1067: Looking at the timeout, I'm wondering if we tune the backoff here? According to GCE documentation, "it waits for no more than the default deadline (2 minutes)", so we may need to set the retry policy to be the current expectation of a 5min timeout.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -76,12 +76,12 @@ var AttachDiskBackoff = wait.Backoff{
 	Cap:      0}
 
 // WaitForOpBackoff is backoff used to wait for Global, Regional or Zonal operation to complete.
-// Default values are similar to Poll every 3 seconds with 5 minute timeout.
+// Default values are similar to Poll every 2 minutes with 6 minute timeout.
 var WaitForOpBackoff = wait.Backoff{
-	Duration: 3 * time.Second,
```
sunnylovestiramisu, line 81: Isn't this shared with some other api calls as well? 3 seconds to 2 mins is a big change.

pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -1066,7 +1066,7 @@ func (cloud *CloudProvider) getRegionalDiskTypeURI(project string, region, diskT
 func (cloud *CloudProvider) waitForZonalOp(ctx context.Context, project, opName string, zone string) error {
 	// The v1 API can query for v1, alpha, or beta operations.
 	return wait.ExponentialBackoff(WaitForOpBackoff, func() (bool, error) {
-		pollOp, err := cloud.service.ZoneOperations.Get(project, zone, opName).Context(ctx).Do()
+		pollOp, err := cloud.service.ZoneOperations.Wait(project, zone, opName).Context(ctx).Do()
```
sunnylovestiramisu: Change naming of pollOp to waitOp too?

pkg/gce-cloud-provider/compute/gce-compute.go
sunnylovestiramisu: Actually I saw you change it to all the places already.

prachirp: There's similarly named waitForOpBackoff* variables in [main.go](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/cmd/gce-pd-csi-driver/main.go#L60), I will update there as well. Otherwise we're updating all the instances where this var is used in gce-compute.go.

pkg/gce-cloud-provider/compute/gce-compute.go
prachirp: Done

pkg/gce-cloud-provider/compute/gce-compute.go
prachirp: Updated to 3 tries at 2 min interval

pkg/gce-cloud-provider/compute/gce-compute.go
tonyzhc: Actually, one risk I see is that "In uncommon cases, when the server is overloaded, the request might return before the default deadline is reached, or might return after zero seconds." Is there anyway to change this backoff to become constant wait until it's been 5min? Sorry for not catching this earlier.

prachirp: Good catch, do we know how this server overload error exactly looks like? If the exponential backoff conditional function can match on and swallow this error then we will get constant 2 min wait like we want. Some illustrative code: https://go.dev/play/p/iNPzsuYTMgD

prachirp: The wait operation returns HTTP 503 in case of server overload. I updated the logic to swallow this error. We will have 2 min wait then retry for this error, and return immediately without retrying for all other "legit" errors.

PR #2176, opened 2025-09-23
Fix partial cache tail latency by correcting the cache chunk size calc

cemakd:
**What type of PR is this?**
/kind bug

The GKE Data Cache was exhibiting extreme tail latency on partially
cached volumes, performing significantly worse than non-cached volumes.
Initial investigation into LVM cache policies and `atime`-induced
write-back thrashing did not reveal the root cause.

A series of `fio` benchmarks comparing cached and uncached volumes
uncovered a bug where the LVM cache was being configured with a
massive 1 GB chunk size. This was traced to a failure in the
`fetchPvSizeGiB` function in `cache.go`. The `pvs` command syntax was
malformed, and a subsequent string parsing error on its empty output
caused the driver to fall back to the default `maxChunkSize` of 1 GB.

This 1 GB chunk size led to extreme "read and write amplification," where a
single 4KB cache miss would trigger an inefficient 1 GB read from the
backing disk, explaining the catastrophic per-miss penalty.

This change corrects the `pvs` command syntax and the string parsing
logic within the driver. With this fix, the cache is now correctly
configured with a reasonable chunk size (e.g., 384 KiB), which resolves
the extreme read and write amplification and restores performance to expected
levels.

Fixes #

```release-note
Fix partial data cache tail latency by correcting the cache chunk size calc
```

### DIFF
--- a/pkg/gce-pd-csi-driver/cache.go
+++ b/pkg/gce-pd-csi-driver/cache.go
@@ -573,6 +573,7 @@ func isCachingSetup(mainLvName string) (error, bool) {
 	return nil, false
 }
 
+// cacheSize is always in GiB
 func fetchChunkSizeKiB(cacheSize string) (string, error) {
 	var chunkSize float64
 
--- a/pkg/gce-pd-csi-driver/cache.go
+++ b/pkg/gce-pd-csi-driver/cache.go
@@ -576,6 +576,7 @@ func isCachingSetup(mainLvName string) (error, bool) {
 func fetchChunkSizeKiB(cacheSize string) (string, error) {
 	var chunkSize float64
 
+	cacheSize = strings.TrimSuffix(cacheSize, "GiB")
 	cacheSizeInt, err := strconv.ParseInt(cacheSize, 10, 64)
 	if err != nil {
 		return "0", err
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/cache.go
```diff
@@ -576,6 +576,7 @@ func isCachingSetup(mainLvName string) (error, bool) {
 func fetchChunkSizeKiB(cacheSize string) (string, error) {
 	var chunkSize float64
 
+	cacheSize = strings.TrimSuffix(cacheSize, "GiB")
```
sunnylovestiramisu, line 580: Anything we want to cover in the TestFetchChunkSizeKiB unit tests?

cemakd: Added a test to make sure strings like "375GiB" parses correctly and results in the expected "384KiB"

sunnylovestiramisu, line 580: Nit: can we add a comment saying cacheSize passed in is always in GiB unit?

pkg/gce-pd-csi-driver/cache_test.go
```diff
@@ -31,6 +31,11 @@ func TestFetchChunkSizeKiB(t *testing.T) {
 			cacheSize:    "1",
 			expChunkSize: "160KiB", //range defined in fetchChunkSizeKiB - min 160 KiB
 		},
+		{
+			name:         "chunk size with GiB string parses correctly",
```
sunnylovestiramisu, line 35: I think the existing test cases are not correct then, because it should always be something GiB

cemakd: Updated the existing ones to have the suffix GiB

pkg/gce-pd-csi-driver/cache.go
cemakd: Done!

sunnylovestiramisu:
Waiting for e2e testing edge cases confirmation

/hold

PR #2178, opened 2025-09-24
Add NodeStageVolume disk size validation before mounting

hajiler:
**What type of PR is this?**
> /kind bug

This PR introduces a safeguard to compare a disk's reported size against its expected size before running mkfs. If the disk's actual size does not match the expected size, the NodeStageVolume fails, and the operation will be retried. 

Fixes #
b/442408876

```release-note
NONE
```


pkg/gce-pd-csi-driver/node.go
```diff
@@ -446,6 +446,15 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
 		klog.V(4).Infof("CSI volume is read-only, mounting with extra option ro")
 	}
 
+	expectedSize := req.GetPublishContext()[common.ContextDiskSizeGB]
```
mattcary: If the expected size is missing, what should we do? 

I'm thinking of an upgrade scenario where the publish context was created on an old version of the driver. If we just fail and wait for it to updated, it might never happen. But on the other hand it might not be safe to assume the size matches if it doesn't exist. I'm not sure.

msau42: IMO I think we should assume upgrade scenario if it's not set and skip enforcement.

mattcary: sgtm

mattcary:
The govet error must be a new check, as that setup_e2e_test code has been in since february.

I'll force-submit, and push a PR to fix the e2e test.

PR #2184, opened 2025-09-29
Move wg.Add outside of gofunc in setup_e2e_test

mattcary:
/kind cleanup

Fix waitgroup usage in setup_e2e_test, for govet error discovered in #2178.

Also trying to fix flakes on ssh (I suspect C3s may have more delay from the instance going ready to ssh being available than n2s?)

```release-note
None
```


mattcary:
/retest

e2e test is flaking on machine bringup

PR #2186, opened 2025-10-03
Move constants.go into new pkg/common/constants

julianKatz:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

I need to make some significant changes to how parameters are handled for Generic Volumes.  To make that easier, I'm doing some cleanup of the pkg/common package, which is overgrown and interwoven.  

This is the easiest place to start, since the constants are just strings and don't have dependencies elsewhere.  If I don't start here, it's very easy to create circular dependencies between a new planned package pkg/common/parameters and pkg/common.

Fixes #

```release-note
NONE
```


mattcary:
The integration tests are failing apparently due to some change in the gcloud boot image specification flags.

Since this change is non-functional, if the unit tests pass & it builds, the risk of breaking the integration tests are low. So I'll force-merge this.

PR #2189, opened 2025-10-07
Move parameters and conversion functions to their own packages

julianKatz:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR is a follow-up to #2186.

This PR moves the parsing of parameters into their own package: `pkg/parameters`.  This is a lot of logic and worthy of its own package.

Along the way, I found that `pkg/common` had a number of conversion helper functions (`Convert.*`) that were public and occasionally used by both parameters.go and other files, and sometimes only parameters.go.  Several of them also had their own unit tests already.  Thus, I thought this was also worthy of its own package.  This PR thus adds `pkg/convert`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/parameters/parameters.go
+++ b/pkg/parameters/parameters.go
@@ -449,7 +449,7 @@ func ExtractModifyVolumeParameters(parameters map[string]string) (ModifyVolumePa
 	return modifyVolumeParams, nil
 }
 
-// ConvertStringToAvailabilityClass converts a string to an availability class string.
+// convertStringToAvailabilityClass converts a string to an availability class string.
 func convertStringToAvailabilityClass(str string) (string, error) {
 	switch strings.ToLower(str) {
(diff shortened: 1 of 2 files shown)


pkg/parameters/parameters.go
```diff
@@ -444,3 +448,39 @@ func ExtractModifyVolumeParameters(parameters map[string]string) (ModifyVolumePa
 	}
 	return modifyVolumeParams, nil
 }
+
+// ConvertStringToAvailabilityClass converts a string to an availability class string.
```
mattcary: nit: make the uppercase match the actual ftn name (now that it's not exported)

julianKatz: Done!

PR #2199, opened 2025-10-14
Integrate disk support labels into NodeGetInfo

hajiler:
**What type of PR is this?**
>

> /kind feature

This PR is a pre-requisite for the generic volumes feature. It adds disk types supported by a node to the topology information using NodeGetInfo. The PR will be followed up by using this topology information to select a compatible disk type when provisioning a volume. 

Fixes #

Node attach limit override and the disk type labels rely on reading the Node k8s object. Instead of repeating this request, I migrated it out to NodeGetInfo and piped down the Node object. This also allowed me to add more comprehensive unit testing. 

```release-note
NONE
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -32,11 +32,11 @@ import (
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	volumehelpers "k8s.io/cloud-provider/volume/helpers"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
-
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/constants"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/deviceutils"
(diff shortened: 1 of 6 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -717,9 +721,26 @@ func (ns *GCENodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoRe
 		Segments: map[string]string{constants.TopologyKeyZone: ns.MetadataService.GetZone()},
 	}
 
+	node, err := k8sclient.GetNodeWithRetry(ctx, ns.MetadataService.GetName())
```
cemakd, line 734: Nit: You can rename `err` to `nodeErr` and avoid having to set it to nil in the if statement.

pkg/gce-pd-csi-driver/node_test.go
```diff
@@ -431,25 +435,59 @@ func TestNodeGetVolumeLimits(t *testing.T) {
...
+		{
+			name:        "attach limit override out of bounds",
+			machineType: "n1-standard-1",
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						fmt.Sprintf(constants.NodeRestrictionLabelPrefix, constants.AttachLimitOverrideLabel): "9999",
+					},
+				},
+			},
```
cemakd, line 473: Let's add a test case where node is nil as well, just to add coverage for the node == nil case

pkg/gce-pd-csi-driver/node.go
hajiler: I removed the err = nil statement, since err should be replaced by logic below. I also changed the failure to get disk labels to just log the error.

pkg/gce-pd-csi-driver/node_test.go
hajiler: The previous test cases lines 251-437 are all node == nil

cemakd: Ah makes sense, I missed those since I hadn't expanded the unchanged lines :)

PR #2203, opened 2025-10-15
Adds a 4-hour maintenance exclusion window to the GKE cluster creation process in the integration tests

tonyzhc:
**What type of PR is this?**
>
> /kind flake

This prevents unexpected upgrade disruptions during test runs by ensuring that GKE maintenance does not occur while tests are in progress. Our tests have been flaking more frequently because of this.

This change does not impact https://testgrid.canary.kubernetes.io/provider-gcp-compute-persistent-disk-csi-driver

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/k8s-integration/cluster.go
+++ b/test/k8s-integration/cluster.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	apimachineryversion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


test/k8s-integration/cluster.go
```diff
@@ -206,6 +207,13 @@ func clusterUpGKE(gceZone, gceRegion string, numNodes int, numWindowsNodes int,
 		locationArg, locationVal, "--num-nodes", strconv.Itoa(numNodes),
 		"--quiet", "--machine-type", "n1-standard-2", "--no-enable-autoupgrade"}
 
+	// To avoid unexpected GKE maintenance, create a 4-hour maintenance exclusion window.
+	t := time.Now().UTC()
+	cmdParams = append(cmdParams, "--maintenance-exclusion",
```
sunnylovestiramisu, line 212: Can we distinguish a GKE cluster from a GCE cluster in this file? Because the OSS GCE tests do not need a maintenance window.

sunnylovestiramisu: I meant this testgrid https://testgrid.canary.kubernetes.io/provider-gcp-compute-persistent-disk-csi-driver does not need maintenance window.

tonyzhc: This argument is appended to the `gcloud container clusters create` command, so it should only be applied to GKE clusters.

sunnylovestiramisu: Can you add this detail to the PR's description? Like, this does not impact https://testgrid.canary.kubernetes.io/provider-gcp-compute-persistent-disk-csi-driver because the maintainence window only applies to GKE clusters.

PR #2206, opened 2025-10-22
scope roles/iam.serviceAccountUser to node service accounts

### LINKED ISSUE #1700: Reduce iam.serviceAccountUser role requirement
Currently, GCP PD CSI driver requires a Service Account that has the `iam.serviceAccountUser` role:

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/eead51bd9029e92e6b6789813ff345f12a04c1a7/docs/kubernetes/user-guides/driver-install.md?plain=1#L21

However, that goes agains Google's Security Health Analytics recommendation:

https://cloud.google.com/security-command-center/docs/how-to-remediate-security-health-analytics-findings#over_privileged_service_account_user

Is it possible to remove this role from the driver's requirements? What would it take to do that?

I looked around this repository but couldn't find the reason this role is required, only this comment that references it: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/134#issuecomment-431612007

dobsonj:
**What type of PR is this?**

/kind documentation

(driver install documentation + deploy scripts)

- **scope roles/iam.serviceAccountUser to node service accounts**

gce-pd-csi-driver requires roles/iam.serviceAccountUser to impersonate node service accounts to attach and detach disks. However, the current driver install instructions and deployment scripts recommend granting this role at the project level, which allows it to impersonate any service account in the project.

[Google Security Health Analytics](https://cloud.google.com/security-command-center/docs/how-to-remediate-security-health-analytics-findings#over_privileged_service_account_user) recommends NOT to scope this role at the project level, and some users are sensitive to this and must follow the principle of least privilege when deploying the CSI driver.

This PR introduces an explanation in the driver install instructions, along with an optional `NODE_SERVICE_ACCOUNTS` environment variable to specify node service accounts that the CSI driver may impersonate. If `NODE_SERVICE_ACCOUNTS` is set, then it scopes `roles/iam.serviceAccountUser` just to those provided service accounts instead of granting it at the project level.

There are two additional commits to fix issues I encountered while testing:

- **add-iam-policy-binding should set condition=None to avoid prompt**
- **remove invalid --enable-multitenancy option from deployments**

Testing: manually tested driver installation using `./deploy/setup-project.sh` and `./deploy/kubernetes/deploy-driver.sh`, with and without `NODE_SERVICE_ACCOUNTS` defined, and confirmed the driver can provision and attach volumes when `roles/iam.serviceAccountUser` is restricted just to the provided node service accounts.

Fixes #1700

/cc @msau42 @mattcary

```release-note
Reduce scope of roles/iam.serviceAccountUser to specified node service accounts during driver installation
```

### DIFF
--- a/deploy/kubernetes/base/controller/controller.yaml
+++ b/deploy/kubernetes/base/controller/controller.yaml
@@ -146,7 +146,6 @@ spec:
             - "--supports-dynamic-throughput-provisioning=hyperdisk-balanced,hyperdisk-throughput,hyperdisk-ml"
             - --enable-data-cache
             - --run-node-service=false
-            - --enable-multitenancy
           command:
             - /gce-pd-csi-driver
           env:
--- a/deploy/kubernetes/base/node_linux/node.yaml
+++ b/deploy/kubernetes/base/node_linux/node.yaml
@@ -47,7 +47,6 @@ spec:
             - "--endpoint=unix:/csi/csi.sock"
             - "--run-controller-service=false"
             - "--enable-data-cache"
-            - "--enable-multitenancy"
             - "--node-name=$(KUBE_NODE_NAME)"
           securityContext:
             privileged: true


sunnylovestiramisu:
What type of clusters have you tested? Is it a GCE cluster or a GKE cluster that disabled the managed PDCSI driver?

dobsonj:
> What type of clusters have you tested? Is it a GCE cluster or a GKE cluster that disabled the managed PDCSI driver?

I tested on GCE - specifically OpenShift 4.20 on GCE with WIF enabled and the included PD CSI driver disabled so that these deployment scripts can deploy successfully to the `gce-pd-csi-driver` namespace. If you'd like to see results on a different configuration I'm happy to give it a try.

mattcary:
I think k8s on GCE is the right platform to test this. The SA and controller structure on GKE is entirely different.

/review @sunnylovestiramisu

deploy/setup-project.sh
```diff
@@ -102,22 +105,36 @@ gcloud iam roles $action gcp_compute_persistent_disk_csi_driver_custom_role --qu
...
+# Grant scoped serviceAccountUser role for node service accounts
+if use_scoped_sa_role;
+then
+	IFS=',' read -ra NODE_SA_ARRAY <<< "${NODE_SERVICE_ACCOUNTS}"
+	for node_sa in "${NODE_SA_ARRAY[@]}";
+	do
+		node_sa=$(echo "${node_sa}" | xargs) # trim whitespace
+		echo "Granting ${SA_USER_ROLE} for ${node_sa} to serviceAccount:${IAM_NAME}"
+		gcloud iam service-accounts add-iam-policy-binding "${node_sa}" \
+			--member="serviceAccount:${IAM_NAME}" --condition=None \
```
sunnylovestiramisu, line 120: Do we need condition=None? What is the default of condition?

deploy/kubernetes/deploy-driver.sh
```diff
@@ -65,6 +68,23 @@ function check_service_account()
 			MISSING_ROLES=true
 		fi
 	done
+
+	# Check each node service account for serviceAccountUser role
+	if use_scoped_sa_role;
+	then
+		IFS=',' read -ra NODE_SA_ARRAY <<< "${NODE_SERVICE_ACCOUNTS}"
```
sunnylovestiramisu, line 75: Can we do some input validation for NODE_SERVICE_ACCOUNTS? Like not empty etc.

dobsonj: The `if use_scoped_sa_role` check already returns false at line 73 if `NODE_SERVICE_ACCOUNTS` is an empty string:
```
function use_scoped_sa_role()
{
	[ -n "${NODE_SERVICE_ACCOUNTS:-}" ]
}
```
for example:
```
$ unset NODE_SERVICE_ACCOUNTS
$ [ -n "$NODE_SERVICE_ACCOUNTS" ]
$ echo $?
1
$ export NODE_SERVICE_ACCOUNTS=""
$ [ -n "$NODE_SERVICE_ACCOUNTS" ]
$ echo $?
1
```
I considered checking for a comma in `NODE_SERVICE_ACCOUNTS`, but it is valid to pass in a single service account with no comma delimiter, and this line still handles that case:
```
$ export NODE_SERVICE_ACCOUNTS="sa1"
$ IFS=',' read -ra NODE_SA_ARRAY <<< "${NODE_SERVICE_ACCOUNTS}" && echo ${NODE_SA_ARRAY[@]}
sa1
$ export NODE_SERVICE_ACCOUNTS="sa1,sa2"
$ IFS=',' read -ra NODE_SA_ARRAY <<< "${NODE_SERVICE_ACCOUNTS}" && echo ${NODE_SA_ARRAY[@]}
sa1 sa2
```
The only other thing I can think to validate is whether it's a valid service account, which we won't know until calling the `gcloud` command, which gives a useful error message:
```
$ gcloud iam service-accounts get-iam-policy "invalid-sa1" --project="${PROJECT}" --flatten="bindings[].members" --format='table(bindings.role)' --filter="bindings.members:${IAM_NAME}"
ERROR: (gcloud.iam.service-accounts.get-iam-policy) argument SERVICE_ACCOUNT: Bad value [invalid-sa1]: Not a valid service account identifier. It should be either a numeric string representing the unique_id or an email of the form: my-iam-account@somedomain.com or my-iam-account@PROJECT_ID.iam.gserviceaccount.com

$ gcloud iam service-accounts get-iam-policy "invalid-sa1@somedomain.com" --project="${PROJECT}" --flatten="bindings[].members" --format='table(bindings.role)' --filter="bindings.members:${IAM_NAME}"
ERROR: (gcloud.iam.service-accounts.get-iam-policy) NOT_FOUND: Unknown service account. This command is authenticated as jdobson@redhat.com which is the active account specified by the [core/account] property
```

deploy/setup-project.sh
dobsonj: I added `--condition=None` because without it I was getting prompted several times when running the script, for example:
```
$ gcloud projects add-iam-policy-binding "${PROJECT}" --member serviceAccount:"${IAM_NAME}" --role "${SA_USER_ROLE}"
...
 [5] TITLE=test-platform__ filter
 [6] None
 [7] Specify a new condition
The policy contains bindings with conditions, so specifying a condition is required when
adding a binding. Please specify a condition.:
```
The list of conditions might be specific to our project(?) but IMO we always want None for these bindings.

I did _not_ see that prompt when running `gcloud iam service-accounts add-iam-policy-binding` to create a binding for a specific service account. The condition defaulted to None in that case, but I included it anyway to leave no ambiguity.

deploy/kubernetes/deploy-driver.sh
sunnylovestiramisu: It is good enough with use_scoped_sa_role :)

PR #2207, opened 2025-10-22
Unpublish legacy volume even if it appears in multiple zones

mattcary:
/kind bug

If there are disks with the same name across multiple zones, and a legacy (pre-migration) volume is used, unpublish will fail when trying to reconstruct the full volume key. However, since we have the node when we are unpublishing, we do know the zone and can detach the correct disk.

Failing to do this results in stuck workloads for customers, as if they manually create a node in a second zone before detaching a similarly named one, they can never unpublish.

This was found in internal testing in GKE.

```release-note
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -322,16 +326,29 @@ func (cloud *CloudProvider) RepairUnderspecifiedVolumeKey(ctx context.Context, p
...
 			}
 
-			if len(foundZone) == 0 {
+			if len(diskZones) == 0 {
 				return "", nil, notFoundError()
+			} else if len(diskZones) > 1 && fallbackZone == "" {
+				return "", nil, fmt.Errorf("found disk %s in more than one zone and no fallback: %v", volumeKey.Name, diskZones)
+			} else if len(diskZones) > 1 && fallbackZone != "" {
+				match := false
+				for _, z := range diskZones {
```
tonyzhc: Use slices.Contains()

mattcary: ooh, thanks!

PR #2214, opened 2025-10-27
Unpublish legacy volume even if it appears in multiple zones

mattcary:
Manual cherry-pick of #2207

(conflicts due to refactoring of pkg/common).

/kind bug

```release-note
```


mattcary:
Skipping unrelated arm error

PR #2216, opened 2025-10-27
Manually create udev links if udev trigger doesn't work

mattcary:
/kind bug

We have seen some cases in the field where `udevadm --trigger` may not function correctly, even when disk serial numbers are found. This PR makes sure a proper symlink is created even if udevadm --trigger doesn't work.

```release-note
Manually create device paths when udevadm --trigger fails.
```

/assign @sunnylovestiramisu


test/e2e/tests/setup_e2e_test.go
```diff
@@ -38,6 +38,9 @@ import (
 	remote "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/test/remote"
 )
 
+const (
+	noMachineType = "none"
```
sunnylovestiramisu, line 42: Is this change related? Seems like the change for ARM test fix

sunnylovestiramisu:
Is there anyway we can test this end to end?

test/e2e/tests/setup_e2e_test.go
mattcary: Weird, that was in a different PR. Let me check.

mattcary:
> Is there anyway we can test this end to end?

Yeah, I'm thinking how to do that. I was going to at least run the regular tests and see what happened.

test/e2e/tests/setup_e2e_test.go
mattcary: Ah, I must have forked off of my arm fix rather than master. Fixed now at any rate.

mattcary:
Holding while I see if I can get an e2e test on a node with udev disabled.

test/e2e/tests/single_zone_e2e_test.go
```diff
@@ -1668,6 +1668,94 @@ var _ = Describe("GCE PD CSI Driver", func() {
 		Expect(err).To(BeNil(), "Failed to rm file path %s: %v", fp, err)
 	})
 
+	It("Should mount if udev disabled, and remount if it's enabled again", func() {
```
msau42, line 1671: Can we verify a few additional scenarios?

1) Detaching a disk with a manually created symlink cleans up the links
2) After re-enabling udev, could it override the manually created link?

mattcary: I think those should be covered already in the scenario below?

PR #2225, opened 2025-10-31
Return early when node name is empty

### LINKED ISSUE #2200: Device Cache initialization slowing down csi-node when `--node-name` isn't specified
This PR https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2141 introduced a new deviceCache that relies on the `--node-name` flag to work. However, nothing prevent the user from not configuring this flag and when this happens, the node pods initialization get delayed significantly because it has to wait for the full backoff:

```
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:24.784762       1 main.go:125] Operating compute environment set to: production and computeEndpoint is set to: <nil>
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:24.785339       1 main.go:134] Sys info: NumCPU: 15 MAXPROC: 1
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:24.785357       1 main.go:139] Driver vendor version v1.21.4-dd.202541
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:24.786456       1 mount_linux.go:316] Cannot create temp dir to detect safe 'not mounted' behavior: mkdir /tmp/kubelet-detect-safe-umount3379456297: read-only file system
...
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:24.790026       1 request.go:1178] Error in request: resource name may not be empty
gcp-pd-csi-driver-node-vbl4c csi-plugin W1014 07:32:24.790071       1 node.go:37] Error getting node : resource name may not be empty, retrying...
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:25.790198       1 request.go:1178] Error in request: resource name may not be empty
gcp-pd-csi-driver-node-vbl4c csi-plugin W1014 07:32:25.790245       1 node.go:37] Error getting node : resource name may not be empty, retrying...
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:27.791374       1 request.go:1178] Error in request: resource name may not be empty
gcp-pd-csi-driver-node-vbl4c csi-plugin W1014 07:32:27.791414       1 node.go:37] Error getting node : resource name may not be empty, retrying...
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:31.793285       1 request.go:1178] Error in request: resource name may not be empty
gcp-pd-csi-driver-node-vbl4c csi-plugin W1014 07:32:31.793325       1 node.go:37] Error getting node : resource name may not be empty, retrying...
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:39.793657       1 request.go:1178] Error in request: resource name may not be empty
gcp-pd-csi-driver-node-vbl4c csi-plugin W1014 07:32:39.793724       1 node.go:37] Error getting node : resource name may not be empty, retrying...
gcp-pd-csi-driver-node-vbl4c csi-plugin E1014 07:32:39.793731       1 node.go:46] Failed to get node  after retries: timed out waiting for the condition
gcp-pd-csi-driver-node-vbl4c csi-plugin W1014 07:32:39.793767       1 main.go:283] Failed to create device cache: failed to get node : timed out waiting for the condition
...
gcp-pd-csi-driver-node-vbl4c csi-plugin I1014 07:32:39.794316       1 gce-pd-driver.go:187] Driver: pd.csi.storage.gke.io
```

That's 15s lost which could be avoided and that's slowing down significantly the rollout of the driver on large clusters.
I know this can be avoided by configuring the flag but I think the situation could also be improved with a saner default.

I have 2 ideas:
1. check the nodeName isn't an empty string before running [NewDeviceCacheForNode](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2141/files#diff-a81f078f88d79f44c944252605b241dc262580acf27c26d395a981c1b9f6241eR284)
2. or if the deviceCache is needed to be running in all cases, then the flag should become mandatory

What do you think?

cemakd:
This allows node driver to avoid an unnecessary 15 second startup delay when node name flag isn't set.

**What type of PR is this?**
> /kind bug

Fixes #2200

```release-note
NONE
```

### DIFF
--- a/pkg/k8sclient/node.go
+++ b/pkg/k8sclient/node.go
@@ -2,6 +2,7 @@ package k8sclient
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -13,6 +14,9 @@ import (
 )
 
 func GetNodeWithRetry(ctx context.Context, nodeName string) (*v1.Node, error) {
+	if nodeName == "" {
+		return nil, fmt.Errorf("node name is empty")
+	}
 	cfg, err := rest.InClusterConfig()
 	if err != nil {
 		return nil, err


pkg/k8sclient/node.go
```diff
@@ -13,6 +14,9 @@ import (
 )
 
 func GetNodeWithRetry(ctx context.Context, nodeName string) (*v1.Node, error) {
+	if nodeName == "" {
```
sunnylovestiramisu, line 17: Will this cause log spamming in the issue's case? Because they will not be specifying nodeName, and we will log the error. Can you confirm that the error will only get log once, and will not spam the logging?

cemakd: Sorry forgot to test this before merging the change. In theory this shouldn't cause a log spam because there would have been an error thrown from this method before this change as well. We just end up failing faster in this case.

cemakd: Confirmed on a cluster that this doesn't cause a log spam

PR #2226, opened 2025-11-03
Increase e2e tests timeout for ARM tests

sunnylovestiramisu:
**What type of PR is this?**
 /kind failing-test

ARM tests failed with [timeout mostly](https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-gcp-compute-persistent-disk-csi-driver-e2e-arm), increasing it as a temp fix.

The original PR for controller modify volume was set to 15mins longer than the base e2e tests, so removing that check as well: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/1826

Fixes #

```release-note
Increase e2e tests timeout for ARM tests
```

### DIFF
--- a/test/run-e2e.sh
+++ b/test/run-e2e.sh
@@ -5,9 +5,6 @@ set -x
 
 readonly PKGDIR=sigs.k8s.io/gcp-compute-persistent-disk-csi-driver
 
-TIMEOUT=50m
-if [ "$RUN_CONTROLLER_MODIFY_VOLUME_TESTS" = true ]; then
-    TIMEOUT=45m
-fi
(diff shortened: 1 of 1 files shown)


sunnylovestiramisu:
I want to make sure the arm tests passed. It is not a required tests right now.
/hold

mattcary:
Hmm, we're not timing out doing things, it looks like we're not able to create an t2a instances. Let me look into the capacity story.

mattcary:
/close

The problem is fixed in #2228. The test runs in 10m, so timeout is not the issue.

PR #2227, opened 2025-11-04
Correctly use the wait group when there is no hyperdisk

mattcary:
An extra wg.Add(1) was mistakenly called when no hyperdisk was used, leading to deadlock in the e2e test on platforms like arm where there is no hyperdisk.

/kind failing-test

```release-note
None
```


mattcary:
Hmm, the arm job still hangs. But it ran okay locally.

mattcary:
Ah, I missed that we wait on the hd test context channel even if there are no such disks created.

mattcary:
gah, lost the race with the merge, I'll make a new PR.

PR #2228, opened 2025-11-05
Refactor test context creation to be less error-prone

mattcary:
Follow-on to #2227, where conditionally creating a hyperdisk test context caused problems in waiting on a channel for those test contexts. Now we avoid that error-prone channel extraction by just writing in to the test context slices directly.

/kind failing-test

```release-note
None
```


mattcary:
Converting to draft this time so that if the arm test fails it won't accidentally merge.

mattcary:
Arm test passed! And in 10m too, so timeout is not the issue.

PR #2229, opened 2025-11-06
Fix disk size validation branch

hajiler:
**What type of PR is this?**
>

> /kind bug

This PR fixes a bug in  disk size validation for NodeStageVolume. During NodeStageVolume, the actual device size is read using blockdev which the size in bytes. However, the expected value of the disk is specified in Gigabytes. This PR converts the blockdev disk size to gigabytes, and changes the validation to only fail if the actual size is less than the expected size to avoid any failures due to precision loss. 

Fixes #

```release-note
Fix a safeguard in NodeStageVolume to compare a disk's reported size against its expected size before running mkfs. If the sizes do not match, the operation will fail and be retried.

Note: To enable this validation, the --enable-disk-size-validation flag must be set to true on the node server.
```

### DIFF
--- a/pkg/gce-pd-csi-driver/node.go
+++ b/pkg/gce-pd-csi-driver/node.go
@@ -32,6 +32,8 @@ import (
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+	volumehelpers "k8s.io/cloud-provider/volume/helpers"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/node.go
```diff
@@ -451,14 +451,18 @@ func (ns *GCENodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStage
...
-		actualSize, err := getBlockSizeBytes(devicePath, ns.Mounter)
+	expectedDiskSizeStr := req.GetPublishContext()[constants.ContextDiskSizeGB]
+	if expectedSizeGib, err := strconv.ParseInt(expectedDiskSizeStr, 10, 64); err == nil {
+		actualSizeBytes, err := getBlockSizeBytes(devicePath, ns.Mounter)
 		if err != nil {
 			return nil, status.Error(codes.Internal, fmt.Sprintf("failed to get block size for '%s': %v", devicePath, err.Error()))
 		}
-		if expectedSize != strconv.FormatInt(actualSize, 10) {
-			return nil, status.Error(codes.Internal, fmt.Sprintf("expected block size %q, got %q", expectedSize, strconv.FormatInt(actualSize, 10)))
+		actualSizeGib := actualSizeBytes / (1 << 30) // convert bytes to GiB
```
mattcary: Use the RoundUpToGiB etc ftns in k8s.io/cloud-provider/volume/helpers instead of doing the math yourself. See pkg/convert for examples.

This will help being consistent about exactly how things round.

PR #2234, opened 2025-11-11
Add parameter sanitization logic

hajiler:
**What type of PR is this?**
>
> /kind feature

This PR adds functionality for sanitizing parameters based on disk type. This will be used by the generic volumes feature to remove any parameters that are invalid for the selected disk. 

It also cleans up the ParameterProcessor to reduce the amount of arguments in ExtractAndProcessParameters. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/parameters/parameters.go
+++ b/pkg/parameters/parameters.go
@@ -29,9 +29,7 @@ import (
 
 // ExtractAndDefaultParameters will take the relevant parameters from a map and
 // put them into a well defined struct making sure to default unspecified fields.
-// extraVolumeLabels are added as labels; if there are also labels specified in
-// parameters, any matching extraVolumeLabels will be overridden.
-func (pp *ParameterProcessor) ExtractAndDefaultParameters(parameters map[string]string, extraVolumeLabels map[string]string, enableDataCache bool, extraTags map[string]string) (DiskParameters, DataCacheParameters, error) {
+func (pp *ParameterProcessor) ExtractAndDefaultParameters(parameters map[string]string) (DiskParameters, DataCacheParameters, error) {
 	p := DiskParameters{
 		DiskType:             "pd-standard",           // Default
 		ReplicationType:      replicationTypeNone,     // Default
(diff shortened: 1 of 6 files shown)


pkg/parameters/sanitize.go
```diff
@@ -0,0 +1,51 @@
...
+			dp.ProvisionedIOPSOnCreate = 0
+		}
+	},
+	"ProvisionedThroughputOnCreate": func(dp *DiskParameters) {
+		if isPD(dp.DiskType) {
+			dp.ProvisionedThroughputOnCreate = 0
+		}
+	},
+	// Does ForceAttach need to be sanitized?
+	"StoragePools": func(dp *DiskParameters) {
```
cemakd, line 24: Do we also need to sanitize `AvailabilityClass` and `EnableMultiZoneProvisioning`? Based on the design doc looks like availability class doesn't cause an error, but enable-multi-zone-provisioning does. But the latter is also only for hdml, what was the concensus on sanitizing this parameter or not?

pkg/parameters/sanitize.go
```diff
@@ -0,0 +1,51 @@
+package parameters
+
+import (
+	"strings"
+)
+
+func sanitizeDiskParameters(dp *DiskParameters) {
```
cemakd: Nit: Just so it's clear from the naming of the method that the sanitization is intended to be used with generic volumes/dynamic storage classes, what do you think about renaming the top method to something along the lines of: `sanitizeDynamicDiskParameters` or `sanitizeDiskParametesForDynamicStorageClass`?

hajiler: Added a comment to be a bit more clear here.

pkg/parameters/sanitize.go
hajiler: We are not supporting HD-ML for dynamic volumesEnableMultiZoneProvisioning, so it will not be a part of sanitization. 

Sanitization will only be performed on DiskParameters whose fields are extracted from the CreateVolumeRequest. AvailabilityClass is specified in the request, but then used to [determine](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/19978a530463fcfed9a4f6a50db15aa2fa0c8cb1/pkg/parameters/parameters.go#L108-L115) the ForceAttach parameter. So sanitization will only apply to ForceAttach.

pkg/parameters/parameters.go
```diff
@@ -31,7 +31,7 @@ import (
 // put them into a well defined struct making sure to default unspecified fields.
 // extraVolumeLabels are added as labels; if there are also labels specified in
```
mattcary: The comment needs to be updated?

pkg/parameters/sanitize.go
```diff
@@ -0,0 +1,56 @@
...
+		return
+	}
+	for _, sanitizer := range sanitizers {
+		sanitizer(dp)
+	}
+}
+
+type sanitizer func(dp *DiskParameters)
+
+// Map of sanitizers for parameters that apply to only a subset of disk types. Each sanitizeer
```
mattcary, line 5: You should put the sanitizers var at the top of the file, and in a var ( ) block. Since that's where people expect to find global vars.

PR #2235, opened 2025-11-12
Enable dynamic disk type selection

hajiler:
**What type of PR is this?**
/kind feature

This PR enables the dynamic volumes feature, allowing the selection of disk type during CreateVolume depending on what disks are actually supported by the node. 

Fixes #

pdcsi-releaser-new-branch=true

```release-note
Introduce dynamic disk provisioning, automatically selecting a volume's disk type for type based on node compatibility (denoted by`disk-type.gke.io/[disk-type]` node labels).
Action: Required: Add the flag "--dynamic-volumes=true" to the PDCSI controller and node components.
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -115,6 +115,9 @@ type GCEControllerServer struct {
 	// If set to true, the CSI Driver will allow volumes to be provisioned with data cache configuration
 	enableDataCache bool
 
+	// If set to true, the CSI Driver will allow volumes to be provisioned with dynamic disk type selection.
+	enableDynamicVolumes bool
+
 	multiZoneVolumeHandleConfig MultiZoneVolumeHandleConfig
 
 	listVolumesConfig ListVolumesConfig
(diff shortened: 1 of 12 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -322,6 +325,13 @@ func (gceCS *GCEControllerServer) createVolumeInternal(ctx context.Context, req
 	// Apply Parameters (case-insensitive). We leave validation of
 	// the values to the cloud provider.
 	params, dataCacheParams, err := gceCS.parameterProcessor().ExtractAndDefaultParameters(req.GetParameters())
+	if gceCS.enableDynamicVolumes && params.DiskType == "dynamic" {
```
cemakd: You can use params.DynamicVolumeType instead of "dynamic" here right?

pkg/gce-pd-csi-driver/controller.go
cemakd, line 1: We can fix the confusing error message mentioned in your doc here as well. On line 748 we can change the error messages into:

```
// Verify the disk type and encryption key of the clone are the same as that of the source disk.
if diskFromSourceVolume.GetPDType() != params.DiskType {
	return nil, status.Errorf(codes.InvalidArgument, "the disk type of the source volume (%q) and the disk type of the new volume (%q) must match", diskFromSourceVolume.GetPDType(), params.DiskType)
}
if !gce.KmsKeyEqual(diskFromSourceVolume.GetKMSKeyName(), params.DiskEncryptionKMSKey) {
	return nil, status.Errorf(codes.InvalidArgument, "the KMS key of the source volume and the new volume must match")
}
```

cemakd (changes requested):

pkg/gce-pd-csi-driver/controller.go
hajiler: Added a IsDiskDynamic method to params to replace this check for readability.

pkg/gce-pd-csi-driver/disk_selection.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Extract disk types from the CSI CreateVolumeRequest.
+func getDynamicDiskTypes(reqParams map[string]string) (*dynamicDiskTypes, error) {
+	if reqParams == nil {
+		return nil, fmt.Errorf("request parameters are nil")
+	}
+
+	pdType := strings.ToLower(reqParams[parameters.ParameterPDType])
+	if pdType == "" {
+		return nil, fmt.Errorf("missing required parameter %q", parameters.ParameterPDType)
```
mattcary: I thought we were going to default pd and hd types to pd-balanced and HdB, respectively?

mattcary:
e2e failure: `W1204 22:11:43.529337    9527 instance.go:234] instance gce-pd-csi-e2e-us-east4-a-n2d-standard-4-0 not in state RUNNING, was REPAIRING`

:-/

/retest

PR #2238, opened 2025-11-20
Check for nil waitOp before examining error code

mattcary:
/kind bug

If the nil check for the waitOp is not done, the driver can panic.

This PR also cleans up some error code handling and does a better job at retrying errors that are transient (although the flakes we're hitting on the e2e test seem to last longer than the two minute retry).

```release-note
Fix segfault when GCE compute backend fails to wait for an operation.
```

/assign @msau42

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -363,7 +363,7 @@ func isGoogleAPIError(err error) (codes.Code, error) {
 }
 
 func loggedErrorForCode(msg string, code codes.Code, err error) error {
-	klog.Errorf(msg+"%v", err.Error())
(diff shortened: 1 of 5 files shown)


mattcary:
/assign @msau42 

I think the test failures are flakes --- we're not seeming to be able to create instances. I'll keep retrying.

/retry

mattcary:
Seems to be flaking on creating an HaHA. It's probably a flake, let's see.

/retest

mattcary:
Hmm, same HdHA provisioning flake. I'll see if I can reproduce, maybe the test needs to be disabled.

mattcary:
Seems to be a GCP backend problem. I'll just skip the HdHA test while we work that out.

PR #2257, opened 2025-12-18
feat: add a controller to label Nodes with PD compatibility labels

hdp617:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

/kind feature

This PR introduces a new component `gce-pd-node-labeler`. It's a controller implemented with `controller-runtime`, which automatically labels Nodes with PD labels. The source of truth is a ConfigMap that can be configured with `--configmap-name` and `--configmap-namespace` flags.

Fixes #

Re. testing, I've verified with a k8s-on-gce cluster with 2 gen2 nodes and 2 gen4 nodes:
```
kubectl get nodes -L node.kubernetes.io/instance-type
NAME                               STATUS   ROLES           AGE     VERSION   INSTANCE-TYPE
control-plane-us-central1-a-xp2s   Ready    control-plane   7h18m   v1.33.6   e2-medium
nodes-gen2-6vfp                    Ready    node            7h10m   v1.33.6   n2-standard-4
nodes-gen2-dshk                    Ready    node            7h6m    v1.33.6   n2-standard-4
nodes-gen4-g4z2                    Ready    node            7h1m    v1.33.6   n4-standard-4
nodes-gen4-l17x                    Ready    node            6h57m   v1.33.6   n4-standard-4
```

Add a configmap (not actually SoT):
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: machine-pd-compatibility
  namespace: gce-pd-csi-driver
data:
  machine-pd-compatibility.json: |-
    {
      "n2": {
        "pd-standard": true
      },
      "n4": {
        "hyperdisk-balanced": true,
      },
    }
```

Build the container image and add this component as a sidecar in the `csi-gce-pd-controller`:
```
      - args:
        - --v=5
        - --configmap-name=machine-pd-compatibility
        - --configmap-namespace=gce-pd-csi-driver
        image: us-central1-docker.pkg.dev/<PROJECT_ID>/gce-pd-node-labeler/gce-pd-node-labeler:test_linux_amd64
        imagePullPolicy: Always
        name: gce-pd-node-labeler
        resources:
          requests:
            cpu: 10m
            memory: 20Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
```

And verify that the labels are added.

```release-note
NONE
```


Dockerfile.node-labeler
```diff
@@ -0,0 +1,29 @@
+# Copyright 2025 The Kubernetes Authors.
```
sunnylovestiramisu, line 1: Will this also support Windows nodes?

deploy/kubernetes/overlays/node-labeler/rbac.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: rbac.authorization.k.io/v1
```
sunnylovestiramisu: Should this be rbac.authorization.**k8s**.io/v1

deploy/kubernetes/overlays/node-labeler/rbac.yaml
```diff
@@ -0,0 +1,45 @@
...
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: gce-pd-node-labeler-binding
+  namespace: gce-pd-csi-driver
+subjects:
+- kind: ServiceAccount
+  name: csi-gce-pd-controller-sa
```
sunnylovestiramisu, line 18: Does this SA already have all the permissions that the labeler needs?

deploy/kubernetes/overlays/node-labeler/configmap.yaml
```diff
@@ -0,0 +1,238 @@
+
+apiVersion: v1
```
sunnylovestiramisu, line 2: Will this file eventually exceed the configmap limit? What is the plan then?

deploy/kubernetes/overlays/node-labeler/patch.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: csi-gce-pd-controller
+spec:
+  template:
+    spec:
+      containers:
+      - name: gce-pd-node-labeler
+        image: cloud-provider-gcp/gce-pd-node-labeler
+        args:
+          - "--v=5"
```
sunnylovestiramisu, line 12: Any request and limit we should restrict for the labeler?

pkg/nodelabeler/reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+
+// machineFamily extracts the machine family from the instance type
+// e.g. "n2-standard-4" -> "n2", "e2-medium" -> "e2".
+func machineFamily(instanceType string) string {
+	mf, _, _ := strings.Cut(instanceType, "-")
+	return mf
+}
+
+// UpdateMachinePDCompatibility updates the controller's machine PD compatibility map from a ConfigMap.
+func (r *Reconciler) UpdateMachinePDCompatibility(ctx context.Context) error {
```
sunnylovestiramisu, line 133: This does not seem thread-safe, will there be some race of updating the map?

cmd/gce-pd-node-labeler/main.go
```diff
@@ -0,0 +1,126 @@
...
+			klog.Errorf("Expected a ConfigMap object but got %T", obj)
+			return nil
+		}
+		if err := reconciler.UpdateMachinePDCompatibility(ctx); err != nil {
+			klog.Errorf("Failed to update disk compatibility map from ConfigMap %s/%s: %v", cm.Namespace, cm.Name, err)
+			return nil
+		}
+
+		var nodes corev1.NodeList
+		if err := mgr.GetClient().List(ctx, &nodes); err != nil {
```
sunnylovestiramisu: This may cause scalability issue if we have a very large cluster - listing all nodes on every ConfigMap change could be heavy.

cmd/gce-pd-node-labeler/main.go
```diff
@@ -0,0 +1,126 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
sunnylovestiramisu, line 4: Any health check for the labeler? How do we know if it is functioning?

pkg/nodelabeler/reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+	cm := &corev1.ConfigMap{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: r.ConfigMapName, Namespace: r.ConfigMapNamespace}, cm)
+	if err != nil {
+		return err
+	}
+	return r.updateMap(cm)
```
sunnylovestiramisu, line 139: Also can we add unit tests around mapFn and UpdateMachinePDCompatibility (mock the client)?

sunnylovestiramisu:
Last but not least, how about documentation update in the repo(feel free to do it in another PR as well)?

Dockerfile.node-labeler
hdp617: I don't expect the node labeler to support Windows nodes because it should only run on the control plane nodes.

cmd/gce-pd-node-labeler/main.go
hdp617: IIUC, it should only reconcile when the `machine-pd-compatibility` ConfigMap changes because of the `WithPredicates` block below.

```
			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
				return obj.GetName() == *configmapName && obj.GetNamespace() == *configmapNamespace
			}))
```

deploy/kubernetes/overlays/node-labeler/configmap.yaml
hdp617: Currently, it's about 1% of the limit. It's very **unlikely** that the number of machine families will grow past the limit. Also, the plan is to eventually migrate to the GCE API.

pkg/nodelabeler/reconciler.go
hdp617: I'm using a mutex in `updateMap()` below when updating the map. Any race conditions that you're seeing?

hdp617:
@sunnylovestiramisu Thanks for looking into this PR. I've replied to a few comments and will address the rest soon.

deploy/kubernetes/overlays/node-labeler/rbac.yaml
hdp617: Review the base [rbac](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/base/controller/cluster_setup.yaml), the SA doesn't have any permissions for `configmaps` and is missing `patch` verb for `nodes`. I also think it's cleaner to have a separate role binding for the node labeler.

deploy/kubernetes/overlays/node-labeler/rbac.yaml
hdp617: Good catch! Fixed.

deploy/kubernetes/overlays/node-labeler/patch.yaml
hdp617: Added request and limit that I think are sensible. We can adjust if needed.

cmd/gce-pd-node-labeler/main.go
hdp617: Added health checks.

pkg/nodelabeler/reconciler.go
hdp617: Done.

hdp617:
@sunnylovestiramisu @mattcary @hajiler Would you mind taking another look? Thanks!

sunnylovestiramisu:
Please provide end to end testing results before the PR review meeting.

pkg/nodelabeler/reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+		klog.V(4).Infof("No compatible PDs found for instance type %q on node %q", instanceType, node.Name)
+		return reconcile.Result{}, nil
+	}
+
+	newLabels := make(map[string]string)
+	maps.Copy(newLabels, node.Labels)
+
+	labelsChanged := false
+	for _, diskType := range pds {
+		label := constants.DiskTypeKeyPrefix + "/" + diskType
```
hajiler: nit: Use dedicated helper function https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/166b533533f0680f122753e339bcbc963f559fec/pkg/common/utils.go#L504-L506

hajiler: IIUC this seems to only add labels for new compatible disks. What about removing disks that are no longer compatible?

hdp617: Sounds good. Updated.

hdp617: Thanks. Added the logic to remove labels and test cases.

hdp617:
Quick summary of the offline discussion with @hajiler below. Feel free to add if I missed something. Would you mind re-LGTM?
* Added the logic to remove labels to align with GKE's behavior. This is more for consistency and completeness, in practice this case is unlikely as removing disk support is a breaking change.
* One consideration is with confidential nodes which can have different PD support. Customers can override the ConfigMap. We can also consider adding an overlay with a ConfigMap for confidential nodes.

I've squashed the commits. @sunnylovestiramisu Could you give the final approval? Thanks!

PR #2259, opened 2026-01-07
Replace kubelet volume mount  with subdirectories

hajiler:
**What type of PR is this?**
>

/kind bug

The driver was previously mounting the entire /var/lib/kubelet host directory. This broad access violates the principle of least privilege by exposing sensitive subdirectories—such as /var/lib/kubelet/pki/ (node certificates) and /var/lib/kubelet/pods/ (all pod volumes and tokens)—to the CSI container. 

This change replaces the single broad root mount with restricted hostPath mounts.

Fixes #

```release-note
NONE
```


sunnylovestiramisu:
Failed tests related to VolumeSnapshot.
/lgtm

PR #2260, opened 2026-01-09
Cherry-pick #2207 to release-1.20

mattcary:
/kind bug

Cherry-pick #2207 to release-1.20

```release-note
```


mattcary:
The integration tests failed on VAC tests which is not a supported feature back at this version. Force-merging.

PR #2268, opened 2026-02-02
Feature: smart attach limits for nodes

davis-haba:
Implement behavior to prevent kube-scheduler from scheduling workloads on a Node until a CSINode resource for that Node is created.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This feature implements logic to solve the problem of the kube-scheduler not respecting a Node's attach limit, due to the CSINode resource not being present before workloads are scheduled.

It works by:
- Registering a MWH which will inject the startup taint on a new nodes, blocking workloads from being scheduled on the node until it is removed
- Adding a controller that waits for the CSINode to exist, and then removes the taint once the CSINode resource is detected

The feature is protected by a flag gate. As an additional failsafe, if a `pd.csi.storage.gke.io/disable-startup-taint=true` label is found on the Node, taint injection will be suppressed. 

Fixes #

```release-note
NONE
```


sunnylovestiramisu:
hmm the e2e tests are missing configuration.

pkg/webhook/taint/node_taint.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// Define the Taint details
+const (
+	StartupTaintKey    = "pd.csi.storage.gke.io/startup"
+	StartupTaintValue  = "true"
+	StartupTaintEffect = corev1.TaintEffectNoSchedule
+
+	// If this label is present on the Node with value "true", the taint will not be applied
+	DisableTaintLabelKey   = "pd.csi.storage.gke.io/disable-startup-taint"
```
sunnylovestiramisu, line 17: This taint should work both on GCE and GKE? We probably need some documentation change.

pkg/webhook/taint/node_taint.go
```diff
@@ -0,0 +1,73 @@
...
+// Handle serves as the admission logic
+func (a *NodeTainter) Handle(ctx context.Context, req admission.Request) admission.Response {
+	node := &corev1.Node{}
+	err := a.decoder.Decode(req, node)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	// Failsafe to skip tainting if the DisableTaintLabelKey=true
+	if val, ok := node.Labels[DisableTaintLabelKey]; ok && val == DisableTaintLabelValue {
```
sunnylovestiramisu, line 40: Can node.Labels be nil?

pkg/webhook/taint/node_taint.go
```diff
@@ -0,0 +1,73 @@
...
+	StartupTaintEffect = corev1.TaintEffectNoSchedule
+
+	// If this label is present on the Node with value "true", the taint will not be applied
+	DisableTaintLabelKey   = "pd.csi.storage.gke.io/disable-startup-taint"
+	DisableTaintLabelValue = "true"
+)
+
+// NodeTainter appends a specific taint to Nodes on creation
+type NodeTainter struct {
+	Client  client.Client
```
sunnylovestiramisu: Is this Client used anywhere?

pkg/webhook/taint/node_taint.go
```diff
@@ -0,0 +1,73 @@
...
+
+// InjectDecoder injects the decoder (required by controller-runtime)
+func (a *NodeTainter) InjectDecoder(d *admission.Decoder) error {
+	a.decoder = *d
+	return nil
+}
+
+func hasStartupTaint(node *corev1.Node) bool {
+	for _, t := range node.Spec.Taints {
+		if t.Key == StartupTaintKey && t.Effect == StartupTaintEffect {
```
sunnylovestiramisu: Do we need to check the value of the taints?

pkg/webhook/taint/node_taint.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// NodeTainter appends a specific taint to Nodes on creation
+type NodeTainter struct {
+	Client  client.Client
+	decoder admission.Decoder
+}
+
+// Handle serves as the admission logic
+func (a *NodeTainter) Handle(ctx context.Context, req admission.Request) admission.Response {
```
sunnylovestiramisu, line 27: The ctx is passed in but never used, did you mean to use it for context aware logging?
```
klog.V(5).Infof("NodeTainter webhook invoked for node creation")
```

pkg/webhook/taint/node_taint_test.go
```diff
@@ -0,0 +1,185 @@
...
+						{Key: StartupTaintKey, Value: StartupTaintValue, Effect: StartupTaintEffect},
+					},
+				},
+			},
+			wantAllowed:   true,
+			wantPatches:   []jsonpatch.JsonPatchOperation{},
+			wantResultMsg: "Node already has startup taint",
+		},
+		{
+			name: "Should ignore node if Fail-Safe label is present",
```
sunnylovestiramisu, line 69: "Should append to node with multiple existing taints"

```
{
    name: "Should append to node with multiple existing taints",
    node: &corev1.Node{
        ObjectMeta: metav1.ObjectMeta{Name: "multi-taint-node"},
        Spec: corev1.NodeSpec{
            Taints: []corev1.Taint{
                {Key: "taint-1", Effect: corev1.TaintEffectNoSchedule},
                {Key: "taint-2", Effect: corev1.TaintEffectNoExecute},
                {Key: "taint-3", Effect: corev1.TaintEffectPreferNoSchedule},
            },
        },
    },
    wantAllowed: true,
    wantPatches: []jsonpatch.JsonPatchOperation{
        {
            Operation: "add",
            Path:      "/spec/taints/3",
            Value: map[string]interface{}{
                "effect": string(StartupTaintEffect),
                "key":    StartupTaintKey,
                "value":  StartupTaintValue,
            },
        },
    },
},

File: pkg/controller/taint/taint_remover.go [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// DriverName must match the name returned by the CSI GetPluginInfo call
+	DriverName = "pd.csi.storage.gke.io"
+
+	// StartupTaintKey is the key that blocks scheduling until the driver is ready
+	StartupTaintKey = "pd.csi.storage.gke.io/startup"
```
sunnylovestiramisu: Nit: this exists in both taint_remove.go and node_taint.go, if not using the common package, maybe make a comment that it needs to be the same as each other's file.

pkg/controller/taint/taint_remover.go
```diff
@@ -0,0 +1,133 @@
...
+			// CSINode deleted, nothing to do
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	// Check if OUR driver is registered in this CSINode
+	// If the driver is not yet in the list, it means the node-driver-registrar hasn't finished.
+	// We return nil and wait for the next update to the CSINode object.
+	if !isDriverRegistered(csiNode, DriverName) {
```
sunnylovestiramisu: There can be potential race here?

1. CSINode created, but node-driver-registrar crashes before registering the driver
2. No updates to CSINode → controller never retries → taint stays forever → workloads can't schedule

Maybe add exponential backoff requeue or timeout.

pkg/controller/taint/taint_remover.go
```diff
@@ -0,0 +1,133 @@
...
+		if driver.Name == driverName {
+			return true
+		}
+	}
+	return false
+}
+
+func hasStartupTaint(node *corev1.Node) bool {
+	for _, t := range node.Spec.Taints {
+		if t.Key == StartupTaintKey {
```
sunnylovestiramisu: Curious why don't we just return newTaints here?

pkg/controller/taint/taint_remover.go
```diff
@@ -0,0 +1,133 @@
...
+				continue // Filter out our taint
+			}
+			newTaints = append(newTaints, t)
+		}
+		node.Spec.Taints = newTaints
+
+		// Issue Patch
+		err = r.client.Patch(ctx, node, client.MergeFrom(originalNode))
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to remove taint from node %s: %v", node.Name, err)
```
sunnylovestiramisu: Maybe make the error message a bit more informative:

return reconcile.Result{}, fmt.Errorf("failed to remove taint from node %s (may indicate permission issues or node deletion in progress): %w", node.Name, err)

sunnylovestiramisu:
Is there any other cases you can think of that the CSINode is deleted without removing the taint, then the taint would be orphaned?

davis-haba:
> Is there any other cases you can think of that the CSINode is deleted without removing the taint, then the taint would be orphaned?

I can think of a few race/edge cases, but they should be covered:
1. Driver is uninstalled
If the DaemonSet running the Driver is deleted, Kubelet cleans up the CSINode. However, the recent iteration added a grace period of 12 minutes, where if the CSINode is still not detected in that time we untaint the Node. 

2. Missed events / taint controller pod crashes
If the taint controller pod crashes, is evicted, or is stuck in leader election exactly when K8s deletes the CSINode object, then we might not get the delete event and trigger the taint removal. The recent iteration to the taint remover watches Nodes too, so when the controller reboots/finishes leader election we will re-sync and perform the appropriate Reconcile. 

3. Kubelet crash or network problems
If a network blip of Kublet crashes causes the cluster to temporarily drop CSI driver registration, the node could be stranded. Our requeue logic in taint remover should handle this.

4. User deletes `CSINode` manually.
Our grace period discussed in (1) should handle this.

pkg/webhook/taint/node_taint.go
davis-haba: Yes, it should work with both. Do you mean documentation on this public repo?

pkg/webhook/taint/node_taint.go
davis-haba: I don't think so, but I added a safe guard just in case

pkg/webhook/taint/node_taint.go
davis-haba: No, good catch (old code). I removed it.

pkg/webhook/taint/node_taint.go
davis-haba: Done

pkg/webhook/taint/node_taint.go
davis-haba: We are implementing the admission.Handle interface, so we need to match the function signature. 
sigs.k8s.io/controller-runtime/pkg/webhook/admission#Handler

pkg/webhook/taint/node_taint_test.go
davis-haba: thanks, good test. I added it.

pkg/controller/taint/taint_remover.go
davis-haba: Good catch. I moved it to constants.go

pkg/controller/taint/taint_remover.go
davis-haba: Added a requeue

pkg/controller/taint/taint_remover.go
davis-haba: I'm not sure what you mean by returning newTaints, could you elaborate?

pkg/controller/taint/taint_remover.go
davis-haba: done

PR #2274, opened 2026-02-18
Apply GCE disk provisioning/attach status labels 

hajiler:
**What type of PR is this?**
>

/kind feature

Fixes #

```release-note
Add provisioning and attached labels to disks to help customers identify leaked volumes. To enable set gce-disk-status flag to true. 
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -111,6 +111,8 @@ var (
 
 	dynamicVolumes = flag.Bool("dynamic-volumes", false, "If set to true, the CSI driver will automatically select a compatible disk type based on the presence of the dynamic-volume parameter and disk types defined in the StorageClass. Disabled by default.")
 
+	gceDiskStatus = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on GCE disks after successful attachment to indicate the attachment status. Disabled by default.")
+
 	diskCacheSyncPeriod = flag.Duration("disk-cache-sync-period", 10*time.Minute, "Period for the disk cache to check the /dev/disk/by-id/ directory and evaluate the symlinks")
 
 	enableDiskSizeValidation = flag.Bool("enable-disk-size-validation", false, "If set to true, the driver will validate that the requested disk size is matches the physical disk size. This flag is disabled by default.")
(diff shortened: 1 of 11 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1242,6 +1260,12 @@ func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Con
 		return nil, common.LoggedError("Errored during WaitForAttach: ", err), disk
 	}
 
+	if gceCS.enableGCEDiskStatus {
```
msau42: We may want to do this first. I'm thinking about a scenario where this fails and then we retry the ControllerPublish call. Will we skip this because the disk is already attached?

hajiler: I moved it earlier. But do we want retry ControllerPublishVolume if the labeling fails? I has assumed it was better to fail open

cmd/gce-pd-csi-driver/main.go
```diff
@@ -101,6 +101,8 @@ var (
 
 	dynamicVolumes = flag.Bool("dynamic-volumes", false, "If set to true, the CSI driver will automatically select a compatible disk type based on the presence of the dynamic-volume parameter and disk types defined in the StorageClass. Disabled by default.")
 
+	gceDiskStatus = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on GCE disks after successful attachment to indicate the attachment status. Disabled by default.")
```
sunnylovestiramisu, line 114: Does this mean the customer needs to set this value in GKE? Or in GKE it will always be true?

hajiler: In GKE this will be always be set to true via the component manifest.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -1209,6 +1227,12 @@ func (gceCS *GCEControllerServer) executeControllerPublishVolume(ctx context.Con
 		return nil, status.Errorf(codes.Internal, "error getting device name: %v", err.Error()), disk
 	}
 
+	if gceCS.enableGCEDiskStatus {
+		if err := gceCS.updateVolumePublishStatusLabel(ctx, project, volKey, disk, constants.AttachedStatus); err != nil {
+			klog.Warningf("Failed to update VolumePublishStatus label for disk %v: %v", volKey, err)
```
sunnylovestiramisu, line 1232: We might need some metrics for this behavior so we can get alert if something is wrong.

But it can come later, not a P0.

PR #2296, opened 2026-04-02
PD Restiping Feature

cemakd:
**What type of PR is this?**
/kind feature

This PR implements automated Persistent Disk conversion logic to support pd-balanced and pd-ssd disks on Gen2 & Gen4 machine types. It addresses the requirement that disks must be "striped for Gen4" to attach to these machines.

The implementation introduces a workflow in the CSI driver to handle attachment failures by attempting a fast metadata conversion or, if opted-in, a slow snapshot-based conversion. The opt-in prevents unintended application downtime for large disks that require significant time to convert.

Fixes #

- The feature is guarded by the --enable-pd-conversion binary flag.
- Slow conversion requires an explicit opt-in via the pd-conversion.gke.io/slow-conversion-opt-in: "true" annotation.
- The driver checks for this annotation across four surfaces: StorageClass, PersistentVolume, Node, and Pod.
- Any explicit "false" (opt-out) on any of these surfaces will override any opt-ins.

```release-note
Implemented automated PD restriping for Gen4 support. Users can now opt-in to slow, snapshot-based disk conversions using the 'pd-conversion.gke.io/slow-conversion-opt-in' annotation on K8s resources to enable seamless migration between machine generations.
```


cmd/gce-pd-csi-driver/main.go
```diff
@@ -90,6 +90,7 @@ var (
 	enableHdHAFlag              = flag.Bool("allow-hdha-provisioning", false, "If set to true, will allow the driver to provision Hyperdisk-balanced High Availability disks")
 	enableDataCacheFlag         = flag.Bool("enable-data-cache", false, "If set to true, the CSI Driver will allow volumes to be provisioned with Data Cache configuration")
 	enableMultitenancyFlag      = flag.Bool("enable-multitenancy", false, "If set to true, the CSI Driver will support running on multitenant GKE clusters")
+	enablePdConversion          = flag.Bool("enable-pd-conversion", false, "If set to true, the CSI Driver will attempt to convert PD disks if the disk is eligible for conversion for the given machien type it is being attached to")
```
hajiler: nit: to be more concise maybe try something similar to

". . .CSI Driver will attempt to convert eligible disks to be compatible with attaching machine types"

also typo on machine.

cemakd: Done, thanks!

deploy/kubernetes/base/node_linux/node.yaml
```diff
@@ -46,7 +46,8 @@ spec:
             - "--v=5"
             - "--endpoint=unix:/csi/csi.sock"
             - "--run-controller-service=false"
-            - "--enable-data-cache"
+            - "--enable-data-cache=true"
+            - "--enable-multitenancy=false"
```
sunnylovestiramisu: Why do we need to set enable-multitenancy? It does not seem relate to this PR?

sunnylovestiramisu:
Will we be able to distinguish the convert disk operation specifically for our SLO?

deploy/kubernetes/base/node_linux/node.yaml
cemakd: This change was temporary, I've just reverted it. Thanks for catching it!

cemakd:
> Will we be able to distinguish the convert disk operation specifically for our SLO?

Yes, each convert API call adds a descriptive label to the metric we export, 'none' is conversion wasn't attempted, this is the default value. 'fast' if fast conversion is attempted and 'slow' if slow conversion was attempted.

In our SLOs we can include or exclude metrics based on the implications of the conversion.

PR #2299, opened 2026-04-09
Wait for getting node longer during startup

rhrmo:
**What type of PR is this?**

/kind bug

On some nodes, the CSI plugin starts while cluster networking to the apiserver is not ready yet (for example right after a reboot). The driver’s startup path calls the API to read the Node object. The previous short backoff could give up too early, the process exited before the CSI gRPC server was listening, and node-driver-registrar then failed to dial the CSI socket, exiting fatally.

Startup paths (data-cache node label and device cache) now use a longer, bounded wait for the Node API call, while runtime paths keep the original shorter retries so CSI RPCs are not blocked for long.

```release-note
NONE
```

### DIFF
--- a/pkg/k8sclient/node.go
+++ b/pkg/k8sclient/node.go
@@ -13,7 +13,7 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func GetNodeWithRetry(ctx context.Context, nodeName string) (*v1.Node, error) {
+func GetNodeWithRetry(ctx context.Context, nodeName string, steps int) (*v1.Node, error) {
 	if nodeName == "" {
 		return nil, fmt.Errorf("node name is empty")
 	}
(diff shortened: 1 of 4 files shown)


pkg/k8sclient/node.go
```diff
@@ -13,15 +13,43 @@ import (
...
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+	return kubernetes.NewForConfig(cfg)
+}
+
+// GetNodeWithRetry uses a short backoff for runtime CSI RPC paths (e.g. NodeGetInfo)
+// so callers are not blocked for long on transient errors.
 func GetNodeWithRetry(ctx context.Context, nodeName string) (*v1.Node, error) {
```
jsafrane: A new parameter to `GetNodeWithRetry` with nr. of steps to perform + some const with nr. of steps to roughly cover ~10 minutes would do the same job as this whole PR.

rhrmo: Ah, I misunderstood how the ExponentialBackoffWithContext is set up, I missed that we never return the error so it will step enough of times, you're right I will update it.

jsafrane:
I think this PR won't help much the issue we're facing.

During node startup, the CSI driver Pod is started so early that the networking is not yet fully established and the CSI driver can't talk to the API server. Increasing the timeout in the driver won't help, because it's the node-driver-registrar that times out waiting for the driver socket and crashes. We need to increase the timeout in the registar, but it's hardcoded.

We'll look for a better way how to start the driver pod, e.g. by an init container that waits for the API server to be reachable before starting the driver + registrar.

PR #2306, opened 2026-04-23
Add cluster identification label

hajiler:
**What type of PR is this?**
/kind feature

This PR adds a cluster id label to GCE disks on creation. These labels can be used to identify which GCE disks are associated with what cluster. 

This is an important prerequisite for the garbage collector to prevent deleting disks associated with tombstone clusters. In GKE this flag will be set via the control plane to a base 32 encoding of the cluster hash. 

Fixes #

```release-note
Enable cluster-id labels add to GCE disks by setting the `--cluster-identifier` flag.
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -113,7 +113,8 @@ var (
 
 	dynamicVolumes = flag.Bool("dynamic-volumes", false, "If set to true, the CSI driver will automatically select a compatible disk type based on the presence of the dynamic-volume parameter and disk types defined in the StorageClass. Disabled by default.")
 
-	gceDiskStatus = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on GCE disks after successful attachment to indicate the attachment status. Disabled by default.")
+	gceDiskStatus      = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on the GCE disk resource after successful attachment to indicate the attachment status. Disabled by default.")
+	clusterOwnershipID = flag.String("cluster-ownership-id", "", "The identifier for the cluster to which the CSI driver belongs. When specified, the id is applied to the GCE disk resource as a label")
 
 	diskCacheSyncPeriod = flag.Duration("disk-cache-sync-period", 10*time.Minute, "Period for the disk cache to check the /dev/disk/by-id/ directory and evaluate the symlinks")
 
@@ -231,6 +232,10 @@ func handle() {
 		klog.Fatalf("Bad extra tags: %v", err.Error())
 	}
 
+	if convert.CheckLabelValue(*clusterOwnershipID) != nil {
+		klog.Fatalf("Bad cluster ownership ID: %v", *clusterOwnershipID)
+	}
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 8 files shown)


pkg/parameters/parameters_test.go
```diff
@@ -550,12 +551,29 @@ func TestExtractAndDefaultParameters(t *testing.T) {
 				ResourceTags: map[string]string{},
 			},
 		},
+		{
+			name:       "Cluster identified specified",
```
cemakd: indentified vs identifier typo?

cmd/gce-pd-csi-driver/main.go
```diff
@@ -112,7 +112,8 @@ var (
 
 	dynamicVolumes = flag.Bool("dynamic-volumes", false, "If set to true, the CSI driver will automatically select a compatible disk type based on the presence of the dynamic-volume parameter and disk types defined in the StorageClass. Disabled by default.")
 
-	gceDiskStatus = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on GCE disks after successful attachment to indicate the attachment status. Disabled by default.")
+	gceDiskStatus     = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on GCE disks after successful attachment to indicate the attachment status. Disabled by default.")
```
cemakd: By "GCE disks" do you mean the PV object here? Clarifying that could be helpful

cmd/gce-pd-csi-driver/main.go
```diff
@@ -112,7 +112,8 @@ var (
...
+	clusterIdentifier = flag.String("cluster-identifier", "", "The identifier for the cluster to which the CSI driver belongs.")
```
cemakd: "cluster-identifier" seems a bit vague here. Is it typically just the cluster name/full cluster resource path? Is it a custom parameter the customer is expected to set or is it optional? Is there any way we can automatically populate it and if customer needs to override it perhaps the flag could be "cluster-identifier-override"?

pkg/gce-pd-csi-driver/gce-pd-driver.go
```diff
@@ -183,6 +183,7 @@ func NewControllerServer(gceDriver *GCEDriver, cloudProvider gce.GCECompute, err
 		EnableDiskTopology:          args.EnableDiskTopology,
 		EnableDiskSizeValidation:    args.EnableDiskSizeValidation,
 		enableGCEDiskStatus:         args.EnableGCEDiskStatus,
+		clusterID:                   args.ClusterID,
```
cemakd: If we go down the route of automatically populating the clusterID field, there should be a way to populate it using the values in gce.conf

Let me know if that is a path we want to go down and I can point you in the right direction. We can discuss offline as well.

cmd/gce-pd-csi-driver/main.go
cemakd: After reading the design doc the usage of this flag is a bit more clear to me. But for the naming of it I would still suggest a more specific name as clusterIdentifier might be too generic for someone not familiar with the feature.

How about `--cluster-ownership-id`?

pkg/parameters/parameters.go
```diff
@@ -52,6 +52,9 @@ func (pp *ParameterProcessor) ExtractAndDefaultParameters(parameters map[string]
 	if pp.EnableGCEDiskStatus {
 		p.Labels[constants.VolumePublishStatus] = constants.ProvisioningStatus
 	}
+	if pp.ClusterID != "" {
+		p.Labels[constants.ClusterIDLabel] = pp.ClusterID
```
cemakd: Since the value fo this label is user provided, do we do any sanitization/validation to make sure that the value doesn't contain any characters that wouldn't be allowed as a label in K8s?

cmd/gce-pd-csi-driver/main.go
hajiler: No, this applies only the GCE Disk resource. Updated accordingly.

cmd/gce-pd-csi-driver/main.go
hajiler: Done.

pkg/gce-pd-csi-driver/gce-pd-driver.go
hajiler: GKE will populate this automatically using cluster hash.

pkg/parameters/parameters.go
hajiler: Done, added validation for the flag in main.go.

PR #2309, opened 2026-04-28
Garbage collector for leaked disks

hajiler:
**What type of PR is this?**

/kind feature

This PR adds a cleanup routine the controller start up, that verifies GCE disks associated with the cluster and deletes
any disks that never successfully became associated with a PersistentVolume. 

Fixes #

```release-note
Introduced `--enable-disk-cleanup` flag to enable deletion of leaked GCE disk resources. This requires both `--gce-disk-status` and `--cluster-ownership-id` to be set. 
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -114,6 +114,7 @@ var (
 	dynamicVolumes = flag.Bool("dynamic-volumes", false, "If set to true, the CSI driver will automatically select a compatible disk type based on the presence of the dynamic-volume parameter and disk types defined in the StorageClass. Disabled by default.")
 
 	gceDiskStatus      = flag.Bool("gce-disk-status", false, "If set to true, the CSI driver will update the volume-publish-status-gke-io label on the GCE disk resource after successful attachment to indicate the attachment status. Disabled by default.")
+	enableDiskCleanup  = flag.Bool("enable-disk-cleanup", false, "If set to true, the CSI driver will run a routine on startup to cleanup leaked GCE disks resources. Requires the cluster-ownership-id flag to be set. Disabled by default.")
 	clusterOwnershipID = flag.String("cluster-ownership-id", "", "The identifier for the cluster to which the CSI driver belongs. When specified, the id is applied to the GCE disk resource as a label")
 
 	diskCacheSyncPeriod = flag.Duration("disk-cache-sync-period", 10*time.Minute, "Period for the disk cache to check the /dev/disk/by-id/ directory and evaluate the symlinks")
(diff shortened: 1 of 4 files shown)


cmd/gce-pd-csi-driver/main.go
```diff
@@ -305,6 +302,15 @@ func handle() {
 		}
 
 		controllerServer = driver.NewControllerServer(gceDriver, cloudProvider, initialBackoffDuration, maxBackoffDuration, fallbackRequisiteZones, *enableStoragePoolsFlag, *enableDataCacheFlag, multiZoneVolumeHandleConfig, listVolumesConfig, provisionableDisksConfig, *enableHdHAFlag, args)
+		if *gceDiskStatus {
+			// Block startup if configured incorrectly, but fail-open on the cleanup routine.
+			if *clusterOwnershipID == "" {
+				klog.Fatalf("Cannot enable cleanup routine without cluster ownership ID specified")
+			}
+			if err := controllerServer.VerifyClusterDisks(ctx, *enableDiskCleanup); err != nil {
```
sunnylovestiramisu, line 310: Do we need some timeout configuration of this operation at all? On clusters with many disks, this could delay controller startup.

pkg/gce-pd-csi-driver/cleanup.go
```diff
@@ -0,0 +1,162 @@
...
+
+	hasPV, hasPVC, err := validateAssociatedObjects(ctx, kc, tags)
+	if err != nil {
+		return fmt.Errorf("failed to validate associated objects: %v", err)
+	}
+
+	// Apply "provisioned" status to disks with existing PV.
+	if hasPV {
+		disk.Labels[constants.VolumePublishStatus] = constants.ProvisionedStatus
+		err = gceCS.CloudProvider.SetDiskLabels(ctx, project, volKey, gcecloudprovider.CloudDiskFromV1(disk), disk.Labels)
```
sunnylovestiramisu: Calling out a small risk that the disk object might not have all the current labels if it was fetched with a limited fields selector.
But the best we can do is to call GetDisk right before this line again.

hajiler: Done.

cmd/gce-pd-csi-driver/main.go
hajiler: I don't think the clean up routine should delay controller start up for very long at all. Once, the existing disks have been queried, I think it's fine to perform the clean up routine asynchronously. 

I added a goroutine to do this without a timeout, but if you think a timeout would be better I can add one. 

https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2309/changes#diff-00b821eac26407cb89766bbd140a41db839089234a329e0f679ae9b9eb8f1b38R51-R60

PR #2315, opened 2026-05-07
[btrfs] add an integration test

motiejus:
Now that the integration tests seem to be running on COS 125, we can add an integration test for Btrfs (for which supported is enabled from COS 125 onwards).

Adding various mountOptions: some of them are native to btrfs, others (`btrfs-`) are this pdcsi-specific and are handled specially.

**What type of PR is this?**

/kind feature

Follow-up from #2001

**Special notes for your reviewer**:  The change in `test/run-k8s-integration.sh is up for debate: do we want to use Btrfs in the pre-merge integration tests, or shall we satisfy running them on TestGrid alone?

**Does this PR introduce a user-facing change?**: NONE

### DIFF
--- a/test/k8s-integration/config/sc-btrfs.yaml
+++ b/test/k8s-integration/config/sc-btrfs.yaml
@@ -1,0 +1,15 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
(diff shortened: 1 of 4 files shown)


motiejus:
Sorry have to get some paperwork done first. Will resubmit.

PR #2319, opened 2026-05-08
Improve observability in the pdcsi process

motiejus:
1. Add pprof endpoints, so one can use `go tool pprof <...>`:

    go tool pprof -http=127.0.0.1:8081 http://localhost:22015/debug/pprof/heap

    .. opens up a browser with the detailed heap profile.

2. Add `/metrics` endpoint, which will emit Go runtime metrics for scraping:

```
$ curl -s localhost:22015/metrics | grep -v ^# | head
go_gc_duration_seconds{quantile="0"} 4.4184e-05
go_gc_duration_seconds{quantile="0.25"} 5.0415e-05
go_gc_duration_seconds{quantile="0.5"} 5.1984e-05
go_gc_duration_seconds{quantile="0.75"} 5.4292e-05
go_gc_duration_seconds{quantile="1"} 6.5013e-05
go_gc_duration_seconds_sum 0.483941779
go_gc_duration_seconds_count 8954
go_gc_gogc_percent 100
go_gc_gomemlimit_bytes 9.223372036854776e+18
go_goroutines 15
```

This is only enabled when pdcsi is invoked with `--http-endpoint=:PORT`, which is off by default.

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: understand the Go runtime of pdcsi better.

**Does this PR introduce a user-facing change?**: YES

```release-note
If --http-endpoint=:PORT is passed to the pdcsi-node process, `/debug/pprof` and `/metrics` endpoints become available.
```

### DIFF
--- a/cmd/gce-pd-csi-driver/main.go
+++ b/cmd/gce-pd-csi-driver/main.go
@@ -198,6 +198,7 @@ func handle() {
 	runServiceWithMetrics := *runControllerService || *runNodeService
 	if runServiceWithMetrics && *httpEndpoint != "" {
 		mm := metrics.NewMetricsManager()
+		mm.RegisterRuntimeMetrics()
 		mm.InitializeHttpHandler(*httpEndpoint, *metricsPath)
 
 		switch {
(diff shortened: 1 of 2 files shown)


sunnylovestiramisu:
If the HTTP endpoint is exposed publicly or on a non-localhost network:
- /debug/pprof could leak internal goroutine state, memory layout, and call stacks
- /metrics exposes runtime behavior (GC patterns, goroutine count, memory limits)
- This information could help attackers understand system internals

However since this feature is also off by default, that mitigate the issue partially.

Maybe document clearly that --http-endpoint should only be bound to localhost (127.0.0.1:PORT) in production

motiejus:
> Maybe document clearly that --http-endpoint should only be bound to localhost (127.0.0.1:PORT) in production

Good idea! I've updated the commit message and added a caveat to the cmdline description.

mattcary:
Sorry for the delay on this, Sunny has been out.

/lgtm
/approve

PR #2325, opened 2026-05-15
Add init-container for storage class default annotation restoration

savirg:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature
> /kind flake

This PR adds a new init-container component that restores StorageClass default
annotations after GKE cluster upgrades. This is required when default RWX
is disabled in Kubernetes 1.36+.

The init-container:
- Checks cluster version and default RWX configuration
- Waits for kubeconfig and API server readiness
- Restores the default annotation on the target StorageClass
- Either sets as default (if no other defaults exist) or removes the
  marker annotation (if other defaults already exist)
  
Changes:
- Add pkg/scdefaultrestorer package with core restoration logic
- Add cmd/init-container/main.go as the entry point
- Add Dockerfile.init-container for multi-arch builds
- Update Makefile with init-container build targets
- Update cloudbuild.yaml for CI/CD pipeline integration

Fixes #

```release-note
Add init-container for storage class default annotation restoration
```

### DIFF
--- a/cmd/default-sc-rollback-handler/main.go
+++ b/cmd/default-sc-rollback-handler/main.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package main is the GCE PD CSI default StorageClass rollback handler entrypoint.
+package main
+
+import (
+	"context"
+	"flag"
+	"os"
+	"time"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/scdefaultrestorer"
+)
+
+func main() {
+	var config scdefaultrestorer.Config
+
+	flag.StringVar(&config.StorageClassName, "storageclass-name", "standard-rwo", "Target StorageClass to restore default annotation for")
+	flag.DurationVar(&config.APITimeout, "api-timeout", 30*time.Second, "Maximum time to wait for API server readiness")
+
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	klog.InfoS("Starting MAP default storageclass rollback handler")
+
+	restorer := scdefaultrestorer.New(config)
+	if err := restorer.Run(context.Background()); err != nil {
+		klog.ErrorS(err, "Failed to restore storage class defaults")
+		os.Exit(1)
+	}
+}
--- a/pkg/scdefaultrestorer/restorer.go
+++ b/pkg/scdefaultrestorer/restorer.go
@@ -1,0 +1,148 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scdefaultrestorer
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+const (
+	defaultClassAnnotation = "storageclass.kubernetes.io/is-default-class"
+	// markerAnnotation is applied to StorageClasses when their default annotation is
+	// disabled and defaulting is handled by the Mutating Admission Policy (MAP). It tracks that the default annotation
+	// was disabled by the system, allowing us to rollback and restore the default
+	// annotation if required on the clusters.
+	markerAnnotation = "pdcsi.storage.gke.io/install-default-using-map"
+)
+
+type Config struct {
+	StorageClassName string
+	APITimeout       time.Duration
+}
+
+type Restorer struct {
+	config Config
+}
+
+func New(cfg Config) *Restorer {
+	return &Restorer{config: cfg}
+}
+
+func (r *Restorer) Run(ctx context.Context) error {
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get kubeconfig: %w", err)
+	}
+	clientset, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return fmt.Errorf("failed to build Kubernetes client: %w", err)
+	}
+
+	// Wait for API server to be ready
+	if err := r.waitForAPIServer(ctx, clientset); err != nil {
+		return err
+	}
+
+	// Restore the default annotation on StorageClass
+	return r.restoreDefaultAnnotation(ctx, clientset)
+}
+
+func (r *Restorer) waitForAPIServer(ctx context.Context, client kubernetes.Interface) error {
+	klog.InfoS("Waiting for API server to be ready")
+
+	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, r.config.APITimeout, true, func(ctx context.Context) (bool, error) {
+		result := client.Discovery().RESTClient().Get().AbsPath("/healthz").Do(ctx)
+		if result.Error() != nil {
+			return false, nil
+		}
+		var statusCode int
+		result.StatusCode(&statusCode)
+		return statusCode == 200, nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("timeout waiting for API server: %w", err)
+	}
+	return nil
+}
+
+func (r *Restorer) restoreDefaultAnnotation(ctx context.Context, client kubernetes.Interface) error {
+	scClient := client.StorageV1().StorageClasses()
+	targetSCName := r.config.StorageClassName
+
+	targetSC, err := scClient.Get(ctx, targetSCName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warningf("StorageClass not found, skipping restoration: %s", targetSCName)
+			return nil
+		}
+		return fmt.Errorf("failed to get StorageClass %s: %w", targetSCName, err)
+	}
+
+	annotations := targetSC.GetAnnotations()
+	if annotations == nil || annotations[markerAnnotation] != "true" {
+		klog.InfoS("Marker annotation not found, no restoration needed", "storageclass", targetSCName)
+		return nil
+	}
+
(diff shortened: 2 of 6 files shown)


cmd/default-sc-rollback-handler/main.go
```diff
@@ -0,0 +1,52 @@
+/*
```
msau42, line 1: I would call this "default-sc-rollback-handler" or something more descriptive of its functionality.

cmd/init-container/main.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/scdefaultrestorer"
+)
+
+func main() {
+	var config scdefaultrestorer.Config
+
+	flag.StringVar(&config.StorageClassName, "storageclass-name", "standard-rwo", "Target StorageClass to restore default annotation for")
+	flag.StringVar(&config.KubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file. If empty, in-cluster configuration is used.")
+	flag.StringVar(&config.DefaultRWXEnabled, "default-rwx-enabled", "", "State of default RWX enablement")
```
msau42: It's more about the MAP enablement, less about RWX.

cmd/init-container/main.go
```diff
@@ -0,0 +1,52 @@
...
+	flag.StringVar(&config.EmulatedVersion, "emulated-version", "", "GKE cluster emulated version")
```
msau42: s/GKE/k8s

cmd/init-container/main.go
```diff
@@ -0,0 +1,52 @@
...
+	flag.DurationVar(&config.KubeconfigTimeout, "kubeconfig-timeout", 60*time.Second, "Maximum time to wait for kubeconfig generation if path is specified")
+	flag.DurationVar(&config.APITimeout, "api-timeout", 150*time.Second, "Maximum time to wait for API server readiness")
+
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	klog.InfoS("Starting GCE PD CSI init-container default storage class verification")
```
msau42: "MAP default storageclass rollback handler"

pkg/scdefaultrestorer/restorer.go
```diff
@@ -0,0 +1,223 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+const (
+	defaultClassAnnotation = "storageclass.kubernetes.io/is-default-class"
+	markerAnnotation       = "pdcsi.storage.gke.io/legacy-default-unpatched"
```
msau42: Suggest to rename to "pdcsi.storage.gke.io/install-default-using-map" to be more descriptive. And add a comment explaining what it means.

pkg/scdefaultrestorer/restorer.go
```diff
@@ -0,0 +1,223 @@
...
+		klog.ErrorS(err, "Failed to parse emulated version", "version", r.config.EmulatedVersion)
+		return fmt.Errorf("invalid emulated version %q: %w", r.config.EmulatedVersion, err)
+	}
+	if !is136Plus {
+		klog.InfoS("Skipping: cluster version is below 1.36", "version", r.config.EmulatedVersion)
+		return nil
+	}
+
+	// Wait for kubeconfig if specified
+	if r.config.KubeconfigPath != "" {
```
msau42: Can you leverage https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/cmd/gce-pd-csi-driver/main.go#L522

This automatically handles multiple ways to configure the kubeconfig including incluster config, and file.

pkg/scdefaultrestorer/restorer.go
```diff
@@ -0,0 +1,223 @@
...
+	// Wait for API server to be ready
+	if err := r.waitForAPIServer(ctx, clientset); err != nil {
+		return err
+	}
+
+	// Restore the default annotation on StorageClass
+	return r.restoreDefaultAnnotation(ctx, clientset)
+}
+
+func (r *Restorer) isVersion136Plus(ver string) (bool, error) {
```
msau42: You can probably leverage https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version

pkg/scdefaultrestorer/restorer.go
```diff
@@ -0,0 +1,223 @@
...
+	minor, err := strconv.Atoi(minorDigits.String())
+	if err != nil {
+		return false, fmt.Errorf("invalid minor version: %w", err)
+	}
+
+	// Check if version is 1.36 or higher
+	return major > 1 || (major == 1 && minor >= 36), nil
+}
+
+func (r *Restorer) waitForKubeconfig(ctx context.Context) error {
```
msau42: Is this required? All of our other controllers assume kubeconfig already exists.

pkg/scdefaultrestorer/restorer_test.go
```diff
@@ -0,0 +1,531 @@
...
+				t.Errorf("expected error, got nil")
+			}
+			if !tt.expectErr && err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+		})
+	}
+}
+
+func TestRestoreDefaultAnnotation(t *testing.T) {
```
msau42, line 30: add a test case with other unrelated annotations to make sure that the patch doesn't impact those.

cmd/default-sc-rollback-handler/main.go
savirg: Done

cmd/init-container/main.go
savirg: Done

cmd/init-container/main.go
savirg: Done

cmd/init-container/main.go
savirg: Done

pkg/scdefaultrestorer/restorer.go
savirg: Done

pkg/scdefaultrestorer/restorer.go
savirg: Done

pkg/scdefaultrestorer/restorer.go
savirg: Thanks, done

pkg/scdefaultrestorer/restorer.go
savirg: When I was initially testing this with a bash script, I ran into errors where the kubeconfig wasn't ready yet, which is why I added this wait logic. I will test it again without the wait to see if it's still necessary. If it works fine without it, I will remove the wait.

pkg/scdefaultrestorer/restorer_test.go
savirg: Done

pkg/scdefaultrestorer/restorer.go
```diff
@@ -0,0 +1,208 @@
...
+	config Config
+}
+
+func New(cfg Config) *Restorer {
+	return &Restorer{config: cfg}
+}
+
+func (r *Restorer) Run(ctx context.Context) error {
+	// Check if SC defaulting by MAP is disabled - only proceed if it is
+	if strings.ToLower(r.config.SCDefaultingByMAPEnabled) != "false" {
```
msau42: I'm actually wondering if it would be simpler if we just don't include the init container at all in the deployment spec if the feature is enabled. This would save us a flag to pass in. Same thing with emulated version.

pkg/scdefaultrestorer/restorer.go
```diff
@@ -0,0 +1,208 @@
...
+			if err == nil {
+				// Config built successfully, attempt to create the clientset
+				clientset, clientErr := kubernetes.NewForConfig(kubeConfig)
+				if clientErr == nil {
+					return clientset, nil
+				}
+				err = clientErr
+			}
+
+			klog.V(4).Infof("Waiting for kubeconfig to be ready: %v", err)
```
msau42: we're not waiting for kubeconfig at this point are we?

cmd/default-sc-rollback-handler/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+func main() {
+	var config scdefaultrestorer.Config
+
+	flag.StringVar(&config.StorageClassName, "storageclass-name", "standard-rwo", "Target StorageClass to restore default annotation for")
+	flag.StringVar(&config.KubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file. If empty, in-cluster configuration is used.")
+	flag.StringVar(&config.SCDefaultingByMAPEnabled, "sc-defaulting-by-map-enabled", "", "Indicates if MAP handles StorageClass defaulting")
+	flag.StringVar(&config.EmulatedVersion, "emulated-version", "", "k8s cluster emulated version")
+	flag.DurationVar(&config.KubeconfigTimeout, "kubeconfig-timeout", 60*time.Second, "Maximum time to wait for kubeconfig generation if path is specified")
+	flag.DurationVar(&config.APITimeout, "api-timeout", 150*time.Second, "Maximum time to wait for API server readiness")
```
msau42: this seems long, does it really take this much time?

pkg/scdefaultrestorer/restorer_test.go
```diff
@@ -0,0 +1,541 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "standard-rwo",
+					Annotations: map[string]string{
+						markerAnnotation: "true",
+					},
+				},
+			},
+			otherDefaults: []*storagev1.StorageClass{},
+			expectErr:     false,
+			expectDefault: true,
```
msau42: it may be simpler to have `expectedStorageClass` instead of each field.

cmd/default-sc-rollback-handler/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/scdefaultrestorer"
+)
+
+func main() {
+	var config scdefaultrestorer.Config
+
+	flag.StringVar(&config.StorageClassName, "storageclass-name", "standard-rwo", "Target StorageClass to restore default annotation for")
+	flag.StringVar(&config.KubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file. If empty, in-cluster configuration is used.")
```
msau42: does this get used?

pkg/scdefaultrestorer/restorer.go
savirg: Done

pkg/scdefaultrestorer/restorer_test.go
savirg: Done

cmd/default-sc-rollback-handler/main.go
savirg: Removed. It's actually automatically registered by GetConfig method. So, we don't need to define it here.

cmd/default-sc-rollback-handler/main.go
savirg: I will check the actual time it takes during my testing and update the default value to be more reasonable accordingly.

pkg/scdefaultrestorer/restorer.go
savirg: Removed kubeconfig waiting

cmd/default-sc-rollback-handler/main.go
savirg: reduced it to 30s

pkg/scdefaultrestorer/restorer.go
savirg: Not required, I've removed the wait method

cmd/default-sc-rollback-handler/main.go
msau42: i still see it here

cmd/default-sc-rollback-handler/main.go
msau42: not seeing updates, did you push the new changes?

savirg: sorry, forgot to push. Pushed now

PR #2327, opened 2026-05-27
e2e: add Basic Provisioning test suite for type=dynamic disk selection

su-sudhir:
**What type of PR is this?**

/kind test

---

Adds a new E2E test file `test/e2e/tests/dynamic_volume_e2e_test.go` covering the Basic Provisioning suite for the `type=dynamic` disk selection feature. The suite validates that `selectDiskTypeFromTopologies()` correctly resolves `type=dynamic` to a concrete disk type based on node topology labels (`disk-type.gke.io/*`), and that cleanup leaves no orphaned GCE resources.

The four test cases added are:

1. **HD Node Dynamic Provisioning** — Verifies that `type=dynamic` resolves to `hyperdisk-balanced` on a c3-standard-4 (HD-capable) node when the `Preferred` topology carries the `disk-type.gke.io/hyperdisk-balanced=true` label.

2. **PD Fallback Provisioning** — Verifies that `type=dynamic` falls back to `pd-balanced` on an n2d-standard-4 (PD-only) node when the `Preferred` topology has only `disk-type.gke.io/pd-*=true` labels (no hyperdisk label).

3. **Default Parameters Provisioning** — Verifies that omitting `pd-type` and `hyperdisk-type` from StorageClass parameters causes the driver to use its built-in defaults: `hyperdisk-balanced` on HD-capable nodes and `pd-balanced` on PD-only nodes.

4. **Volume Lifecycle (Create and Delete)** — Verifies the full volume lifecycle: `CreateVolume` resolves `dynamic` to the correct type, the GCE disk is created in `READY` state with a valid volume ID, and `DeleteVolume` removes the disk completely with no orphaned resources.

Each test explicitly passes `Preferred` topology segments so results are consistent across both GKE clusters (where the node labeler stamps real labels) and OSS clusters (where no node labels exist).

---

```
NONE
```
---

```
release-note
NONE

### DIFF
--- a/test/e2e/tests/dynamic_volume_e2e_test.go
+++ b/test/e2e/tests/dynamic_volume_e2e_test.go
@@ -26,20 +26,6 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// ============================================================================
-// Basic Provisioning Test Suite
-// ============================================================================
-
-// HD Node Dynamic Provisioning
-//
-// Verifies that when type=dynamic is used on a hyperdisk-capable node (c3-standard-4),
-// the driver resolves the disk type to hyperdisk-balanced instead of leaving it as "dynamic".
-//
-// Test flow:
-//  1. Pick a c3-standard-4 VM — supports hyperdisk-balanced
-//  2. Call CreateVolume with type=dynamic, pd-type=pd-balanced, hyperdisk-type=hyperdisk-balanced
-//  3. Verify via GCE API that the created disk is hyperdisk-balanced (not pd-balanced or "dynamic")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mattcary** [comment]
The presubmit is failing with the following. I think you need to update the deployment to enable dynamic volumes? I think we can do that by default for all tests as it should be backwards compatible:

```
  [FAILED] CreateVolume failed: rpc error: code = InvalidArgument desc = failed to extract parameters: parameters contain invalid disk type dynamic, must enable dynamic volumes
```

mattcary:

I think that was just a flake

su-sudhir:
Closing this PR and raising this with a different id for CLA verification.

PR #2329, opened 2026-06-05
Add an overlay for SELinuxMount

jsafrane:
**What type of PR is this?**
/kind feature

SELinuxMount feature [1] needs /etc/selinux and /sys/fs/selinux bind-mounted from the host to the driver container in the node DaemonSet.

/bin/mount in the container, together with libselinux, uses it to detect that SELinux is enabled on the host and pass SELinux mount options to kernel.

Without these HostPath volumes, /bin/mount stripes the SELinux mount option and the feature does not work.

There are no code changes necessary. The driver only needs to pass SELinux mount options from NodeStage / NodePublish calls to kernel. And the driver announces capability to do so in its CSIDriver instance.

I don't expect a high demand for SELinux, so I keep it as a new overlay. Tested with Fedora 34 + kubeadm + Kubernetes 1.36.

1: https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1710-selinux-relabeling and especially
[#csi-driver-considerations](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1710-selinux-relabeling#csi-driver-considerations)

**Why it is important**

1. kOps uses manifests from this repository as a base for Kubernetes installation on GCE. We use kOps as an installer in [kubernetes/kubernetes CI jobs](https://github.com/kubernetes/test-infra/blob/69815d7bc187455c01e317cf1a447c8cc5f77b71/config/jobs/kubernetes/sig-storage/sig-storage-gce-config.yaml#L226) that require SELinux, as it's the only one that supports SELinux in k/k CI out of the box. Right now, [we have code in kOps](https://github.com/kubernetes/kops/blob/551a385517e3b5d5d6808b521877577ef06a2df6/upup/models/cloudup/resources/addons/gcp-pd-csi-driver.addons.k8s.io/k8s-1.23.yaml.template#L703-L706) to add the support, however, next time someone syncs the yaml files from here, these changes may get lost.
2. SELinuxMount is graduating to GA soon and community CSI drivers should provide a way how to use it. I admin the demand may be very little.

```release-note
Added a new kustomize overlay to enable SELinuxMount support
```

### DIFF
--- a/deploy/kubernetes/README.md
+++ b/deploy/kubernetes/README.md
@@ -20,7 +20,8 @@ The current structure for kustomization is as follows. Note that Windows support
   * `noauth` Based on stable-master, patches the [base controller configuration](deploy/kubernetes/base/controller.yaml) to remove any dependencies on service account keys.
   * `noauth-debug`: Based on stable-master, used for debugging purposes only, see docs/kubernetes/development.md.
   * `prow-stable-sidecar-rc-master`: Used for prow tests on OSS testgrid to test the latest sidecars. Contains deployment specs of a driver with k8s master branch, driver latest release candidate, and stable sidecars.
-  * `prow-canary-sidecar`: Used for prow tests on OSS testgrid to test release candidates when a new release is being cut. Contains deployment specs of a driver with k8s master branch, Kubernetes driver latest build, and canary sidecars. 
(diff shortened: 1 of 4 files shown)


mattcary:
Sorry for the delay, Sunny has been out.

/lgtm
/approve

PR #2334, opened 2026-06-09
Refactor node_taint webhook to fix data drop issue.

hime:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

This PR fixes a critical data-loss issue in the node_taint admission webhook. Previously, the webhook utilized PatchResponseFromRaw to compute a diff against an unmarshaled corev1.Node struct. Because the Go struct does not track unknown/custom fields, this approach silently stripped out any unrecognized fields or annotations from the raw JSON payload cluster-wide.

The webhook now explicitly constructs targeted JSON patches ([]map[string]interface{}) and relies on the native corev1.Taint struct to safely append or initialize the startup taint without modifying or dropping any other fields on the Node.

Which issue(s) this PR fixes:
N/A

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/webhook/taint/node_taint.go
+++ b/pkg/webhook/taint/node_taint.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"net/http"
 
+	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/constants"
 )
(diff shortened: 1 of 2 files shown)


davis-haba (approved):

pkg/webhook/taint/node_taint.go
```diff
@@ -45,17 +47,60 @@ func (a *NodeTainter) Handle(ctx context.Context, req admission.Request) admissi
...
+	return createPatchResponse(patches, "Applying startup taint")
+}
+
+// buildTaintPatches generates the JSON patch operations for adding the taint
+func buildTaintPatches(existingTaints []corev1.Taint, taintObj corev1.Taint) []map[string]interface{} {
+	hasExistingTaints := len(existingTaints) > 0
+	var patches []map[string]interface{}
+
+	// Initialize array if null, otherwise append to it
+	if !hasExistingTaints {
```
hajiler: nit: Maybe reword this to be in the affirmative (https://engdoc.corp.google.com/eng/doc/tott/episodes/764.md?cl=head)

hime: Done.

PR #2335, opened 2026-06-09
Fix concurrent map access in csiErrorBackoff

hime:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Multiple pods failing due to attachment failures can cause the controller to reach the csiErrorBackoff map concurrently, causing the pdcsi controller to crash. We are adding a mutex to this map to safely access the data and prevent crash.

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud/meta"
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -190,6 +191,7 @@ func (c ListVolumesConfig) listDisksFields() []googleapi.Field {
 type csiErrorBackoffId string
 
 type csiErrorBackoff struct {
+	mu         sync.Mutex
```
hajiler, line 194: Did you consider using RWMutext instead? 

I'm not super familiar here with the backoff usage, but if there are more read than write operations we shouldn't block concurrent reads. Granted, these are map operations so it should be pretty fast, but something to consider.

hime: I considered, but I'm not sure how read vs writes compare and if its sufficient to justify a RWMutex. I went with the safe option of regular mutex. Do we know how many more read operations are there?

hajiler: Got it, since the map operations should be very fast, this looks good to me.

PR #2342, opened 2026-06-10
e2e: add Basic Provisioning test suite for type=dynamic disk selection

sudhirshanmugam:
**What type of PR is this?**

/kind feature

---

Adds a new E2E test file `test/e2e/tests/dynamic_volume_e2e_test.go` covering the Basic Provisioning suite for the `type=dynamic` disk selection feature. The suite validates that `selectDiskTypeFromTopologies()` correctly resolves `type=dynamic` to a concrete disk type based on node topology labels (`disk-type.gke.io/*`), and that cleanup leaves no orphaned GCE resources.

The four test cases added are:

1. **HD Node Dynamic Provisioning** — Verifies that `type=dynamic` resolves to `hyperdisk-balanced` on a c3-standard-4 (HD-capable) node when the `Preferred` topology carries the `disk-type.gke.io/hyperdisk-balanced=true` label.

2. **PD Fallback Provisioning** — Verifies that `type=dynamic` falls back to `pd-balanced` on an n2d-standard-4 (PD-only) node when the `Preferred` topology has only `disk-type.gke.io/pd-*=true` labels (no hyperdisk label).

3. **Default Parameters Provisioning** — Verifies that omitting `pd-type` and `hyperdisk-type` from StorageClass parameters causes the driver to use its built-in defaults: `hyperdisk-balanced` on HD-capable nodes and `pd-balanced` on PD-only nodes.

4. **Volume Lifecycle (Create and Delete)** — Verifies the full volume lifecycle: `CreateVolume` resolves `dynamic` to the correct type, the GCE disk is created in `READY` state with a valid volume ID, and `DeleteVolume` removes the disk completely with no orphaned resources.

Each test explicitly passes `Preferred` topology segments so results are consistent across both GKE clusters (where the node labeler stamps real labels) and OSS clusters (where no node labels exist).

---

```
None
```

---

```
NONE
```
/release-note-none

### DIFF
--- a/test/e2e/tests/dynamic_volume_e2e_test.go
+++ b/test/e2e/tests/dynamic_volume_e2e_test.go
@@ -26,20 +26,6 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-// ============================================================================
-// Basic Provisioning Test Suite
-// ============================================================================
-
-// HD Node Dynamic Provisioning
-//
-// Verifies that when type=dynamic is used on a hyperdisk-capable node (c3-standard-4),
-// the driver resolves the disk type to hyperdisk-balanced instead of leaving it as "dynamic".
-//
-// Test flow:
-//  1. Pick a c3-standard-4 VM — supports hyperdisk-balanced
-//  2. Call CreateVolume with type=dynamic, pd-type=pd-balanced, hyperdisk-type=hyperdisk-balanced
-//  3. Verify via GCE API that the created disk is hyperdisk-balanced (not pd-balanced or "dynamic")
-//  4. Cleanup: DeleteVolume and confirm disk is gone
 var _ = Describe("GCE PD CSI Driver Dynamic Volumes HD Node Provisioning", func() {
 
 	It("Should provision hyperdisk-balanced on HD-capable node when type=dynamic", func() {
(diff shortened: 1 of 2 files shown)


sudhirshanmugam:
@mattcary 

Raised a new PR with CLA verification.

mattcary:
The arm and windows failures can be ignored.

test/e2e/tests/dynamic_volume_e2e_test.go
```diff
@@ -0,0 +1,297 @@
...
+						Segments: map[string]string{
+							"topology.gke.io/zone": z,
+						},
+					},
+				},
+				Preferred: []*csi.Topology{
+					{
+						Segments: map[string]string{
+							"topology.gke.io/zone":                        z,
+							common.DiskTypeLabelKey("hyperdisk-balanced"): "true",
```
mattcary, line 60: Won't these show up in requisite topologies in a real allocation?

test/e2e/tests/dynamic_volume_e2e_test.go
```diff
@@ -0,0 +1,297 @@
...
+						Segments: map[string]string{
+							"topology.gke.io/zone": z,
+						},
+					},
+				},
+				Preferred: []*csi.Topology{
+					{
+						Segments: map[string]string{
+							"topology.gke.io/zone":                 z,
+							common.DiskTypeLabelKey("pd-balanced"): "true",
```
mattcary, line 122: ditto about requisite

test/e2e/tests/dynamic_volume_e2e_test.go
sudhirshanmugam: Updated the Requisite topology to include the same disk-type labels as Preferred, which matches what k8s would actually send - when a node has disk-type.gke.io/* labels, those appear in both Requisite and Preferred in a real WaitForFirstConsumer allocation. Applied the same fix across all test cases in the file.

PR #2352, opened 2026-06-22
fix duplicate VG detection bypass and unconditional uncaching on Spot VM preemption.

sandeep-gh-acc:
**What type of PR is this?**
kind bug

When running on Spot VMs, sudden preemptions can prevent volume unmounting from terminating cleanly, leaving the disk in a broken LVM state.

Fixes two issues. 

Bug 1: Duplicate VG Name Collision (Preemption Deadlock)

Abrupt node preemption leaves stale LVM Volume Group metadata (csi-vg-<node-hash>) on the GCE PD, causing a name collision when the volume is rescheduled on a replacement node with the same name. The baseline driver lacked LVM command isolation, and its duplicate detection logic failed to trigger due to a CLI output-parsing bug. We resolved this by implementing a robust metadata parser (getVgInfoForPv) that queries the GCE PD under a strict LVM device filter to retrieve the stale VG's name and UUID. The driver now isolates LVM commands to the GCE PD, prunes the missing SSD references, and renames the conflicting VG to a unique temporary name, cleanly breaking the deadlock.

Bug 2: Unconditional Cache Purge (Linear Volume Corruption Risk)

During preemption recovery, the driver unconditionally executed lvconvert --uncache to detach the missing local SSD cache pool from the volume group. If the volume was a standard linear (uncached) volume, running this command was invalid and risked LVM metadata corruption, filesystem errors, and data loss. We resolved this by introducing a dynamic volume layout check (isLvCachedOnDevice) that queries the volume's cache state before executing recovery. The driver now runs the uncache operation only on cached volumes and safely bypasses it for linear volumes, protecting standard Persistent Volumes from corruption.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/tests/single_zone_e2e_test.go
+++ b/test/e2e/tests/single_zone_e2e_test.go
@@ -1500,6 +1500,162 @@ var _ = Describe("GCE PD CSI Driver", func() {
 		Expect(err).To(BeNil(), "Failed to go through volume lifecycle")
 
 	})
+	It("Should successfully recover from hard preemption and duplicate VG name conflict on same node", func() {
+		Expect(testContexts).ToNot(BeEmpty())
+		zoneToContext := map[string][]*remote.TestContext{}
+		testZoneContexts := []*remote.TestContext{}
+		for _, tc := range testContexts {
+			_, z, _ := tc.Instance.GetIdentity()
+			if _, ok := zoneToContext[z]; !ok {
+				zoneToContext[z] = []*remote.TestContext{tc}
+			} else {
+				zoneToContext[z] = append(zoneToContext[z], tc)
+			}
+			if len(zoneToContext[z]) >= 2 {
+				testZoneContexts = zoneToContext[z]
+				break
+			}
+		}
+		if len(testZoneContexts) < 2 {
+			Skip("Skipping preemption test as less than 2 VM instances are available in the same zone")
+		}
+
+		testContextForVm1 := testZoneContexts[0]
+		if testContextForVm1.Instance.GetLocalSSD() == 0 {
+			Skip("Skipping Data Cache preemption test as no local SSD is available in context")
+		}
+
+		p, z, _ := testContextForVm1.Instance.GetIdentity()
+		client1 := testContextForVm1.Client
+		firstInstance := testContextForVm1.Instance
+
+		testContextForVm2 := testZoneContexts[1]
+		client2 := testContextForVm2.Client
+		secondInstance := testContextForVm2.Instance
+
+		volName, volID := createAndValidateUniqueZonalDisk(client1, p, z, standardDiskType)
+		defer deleteVolumeOrError(client1, volID)
+
+		// State tracking for robust cleanup to avoid host VM contamination
+		var stageDir1, publishDir1 string
+		var stageDir2, publishDir2 string
+		var staged1, published1, attached1 bool
+		var staged2, published2, attached2 bool
+
+		defer func() {
+			klog.Infof("Cleaning up preemption test resources on VM1 %s...", firstInstance.GetName())
+			if published1 && publishDir1 != "" {
+				err := client1.NodeUnpublishVolume(volID, publishDir1)
+				if err != nil {
+					klog.Errorf("Cleanup NodeUnpublishVolume on VM1 failed: %v", err)
+				}
+			}
+			if staged1 && stageDir1 != "" {
+				err := client1.NodeUnstageVolume(volID, stageDir1)
+				if err != nil {
+					klog.Errorf("Cleanup NodeUnstageVolume on VM1 failed: %v", err)
+				}
+			}
+			if attached1 {
+				err := client1.ControllerUnpublishVolume(volID, firstInstance.GetNodeID())
+				if err != nil {
+					klog.Errorf("Cleanup ControllerUnpublishVolume (detach) on VM1 failed: %v", err)
+				}
+			}
+
+			klog.Infof("Cleaning up preemption test resources on VM2 %s...", secondInstance.GetName())
+			if published2 && publishDir2 != "" {
+				err := client2.NodeUnpublishVolume(volID, publishDir2)
+				if err != nil {
+					klog.Errorf("Cleanup NodeUnpublishVolume on VM2 failed: %v", err)
+				}
+			}
+			if staged2 && stageDir2 != "" {
+				err := client2.NodeUnstageVolume(volID, stageDir2)
+				if err != nil {
+					klog.Errorf("Cleanup NodeUnstageVolume on VM2 failed: %v", err)
+				}
+			}
+			if attached2 {
+				err := client2.ControllerUnpublishVolume(volID, secondInstance.GetNodeID())
+				if err != nil {
+					klog.Errorf("Cleanup ControllerUnpublishVolume (detach) on VM2 failed: %v", err)
+				}
+			}
+			klog.Infof("Cleanup completed for both VMs.")
+		}()
+
+		attachMountArgs := attachAndMountArgs{
+			readOnly:       false,
+			useBlock:       false,
+			forceAttach:    false,
+			setupDataCache: true,
+		}
+
+		// 1. Stage and Mount the volume with Data Cache enabled on VM1
+		klog.Infof("Step 1: Staging and mounting Data Cache volume %s on VM1 %s", volName, firstInstance.GetName())
+		err, _, args := testAttachAndMount(volID, volName, firstInstance, client1, attachMountArgs)
+		Expect(err).To(BeNil(), "Failed to stage and mount volume on VM1")
+		stageDir1 = args.stageDir
+		publishDir1 = args.publishDir
+		attached1 = true
+		staged1 = true
+		published1 = true
+
+		// 2. Write a verification file to prove data persistence on VM1
+		klog.Infof("Step 2: Writing verification file to volume on VM1")
+		writeVerifyFunc, readVerifyFunc := testWriteAndReadFile(firstInstance, false /* readOnly */)
+		err = writeVerifyFunc(args)
+		Expect(err).To(BeNil(), "Failed to write verification file to mounted volume on VM1")
+
+		// Verify we can read the file back immediately on VM1 before preemption
+		klog.Infof("Verifying we can read the file back on VM1 before preemption...")
+		err = readVerifyFunc(args)
+		Expect(err).To(BeNil(), "Failed to read verification file back on VM1 before preemption")
+
+		// Flush filesystem page cache to physical GCE PD before force-detaching!
+		klog.Infof("Flushing filesystem buffers to disk using sync on VM1...")
+		_, err = firstInstance.SSHNoSudo("sync")
+		Expect(err).To(BeNil(), "Failed to run sync command on VM1")
+
+		// Wait a bit to ensure GCE PD storage backend commits are fully settled
+		klog.Infof("Waiting 10 seconds to ensure storage backend commits are fully settled...")
+		time.Sleep(10 * time.Second)
+
+		// 3. Simulate hard preemption: Force-detach the GCE PD from VM1 WITHOUT unstaging!
+		// This leaves the LVM volume group metadata active on both VM1's local SSDs and the GCE PD.
+		klog.Infof("Step 3: Simulating preemption by force-detaching the disk from VM1 without unstaging...")
+		err = firstInstance.DetachDisk(volName)
+		Expect(err).To(BeNil(), "Failed to force-detach disk from VM1 to simulate preemption")
+		attached1 = true
+
+		// Wait a few seconds for the GCE control plane to register the detach
+		time.Sleep(10 * time.Second)
+
+		// 4. Attach the GCE PD to VM2 (re-triggering the duplicate conflict on stage because VM2 has its own csi-vg-node)
+		klog.Infof("Step 4: Attaching the disk to VM2 %s to trigger duplicate VG conflict...", secondInstance.GetName())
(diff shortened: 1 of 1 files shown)


hime:
Could we clean up the PR metadata so it's easier to track in our release notes and commit history?

1. Update the title to be more descriptive about what is being done here e.g fix duplicate VG detection bypass and unconditional uncaching on Spot VM preemption.
2. Update PR description to address how we are addressing the two bugs that are being addressed.

pkg/gce-pd-csi-driver/cache_test.go
```diff
@@ -121,55 +122,137 @@ func TestFetchNumberGiB(t *testing.T) {
...
+			pvsOutput: "  csi-vg-456  xyz-uvw-456  /dev/sdc\n",
+			expVgName: "",
+			expVgUuid: "",
 		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Mock runCommand
+			oldRunCommand := runCommand
```
hime: We tend to add t.Parallel to the tests, can you add a comment saying why we can't parallelize this test.

sandeep-gh-acc: Updated

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -51,6 +53,20 @@ func fetchRAIDedLocalSsdPath() (string, error) {
...
 }
 
+func getVgUuid(vgName string) (string, error) {
+	args := []string{
+		"--noheadings",
+		"-o",
+		"vg_uuid",
+		vgName,
+	}
+	output, err := runCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "vgs", args...)
```
hime: Could we add a brief comment explaining why we are passing "" /* pipedCmd */ and nil /* pipedCmdArg */ here? This would help with context

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
 	// When a node is preempted and replaced, the GCE PD retains the old VG metadata/name.
 	// To prevent name conflicts, we use an LVM device filter to isolate commands to ONLY scan this GCE PD.
-	staleVgUuid, err := getVgUuidForPv(devicePath)
-	if err != nil {
-		klog.Errorf("Failed to check VG UUID for PV %v: %v", devicePath, err)
-	} else if staleVgUuid != "" {
-		hasDuplicate, dupErr := hasDuplicateVgName(volumeGroupName)
-		if dupErr != nil {
-			klog.Errorf("Failed to check duplicate VG name for %s: %v", volumeGroupName, dupErr)
+	staleVgName, staleVgUuid, err := getVgInfoForPv(devicePath)
```
hime: How do we know the volume group is stale at this point? Does this function only get called after reboot?

hime: I see fetchRAIDedLocalSsdPath, does this always stay persistent across reboots? Some machine types have different guarantees here.

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+	if err != nil {
+		klog.Errorf("Failed to check VG info for PV %v: %v", devicePath, err)
```
hime: can you comment on the driver behavior here? does it mean workloads using PDCSI fail to start?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
 	// To prevent name conflicts, we use an LVM device filter to isolate commands to ONLY scan this GCE PD.
-	staleVgUuid, err := getVgUuidForPv(devicePath)
-	if err != nil {
-		klog.Errorf("Failed to check VG UUID for PV %v: %v", devicePath, err)
-	} else if staleVgUuid != "" {
-		hasDuplicate, dupErr := hasDuplicateVgName(volumeGroupName)
-		if dupErr != nil {
-			klog.Errorf("Failed to check duplicate VG name for %s: %v", volumeGroupName, dupErr)
+	staleVgName, staleVgUuid, err := getVgInfoForPv(devicePath)
+	if err != nil {
```
hime: Some of the items in these if else blocks are good candidates for encapsulating in their own fuction. One example is the duplicate cleanup step in the else block.

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+			} else {
+				klog.Infof("Stale volume %s/%s is not cached, skipping uncache step during preemption recovery", volumeGroupName, mainLvName)
+			}
+
+			// 3. Clean up missing PVs on the stale VG using the device filter.
+			// Since the cache has been dissociated, this safely removes the missing local SSD references from the VG metadata.
+			klog.Infof("Step 3: Cleaning up missing PVs on duplicate VG %s using device filter", volumeGroupName)
 			reduceArgs := []string{"--removemissing", volumeGroupName, "--force", "--config", configFilter}
-			reduceInfo, reduceErr := common.RunCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "vgreduce", reduceArgs...)
+			_, reduceErr := runCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "vgreduce", reduceArgs...)
```
hime: can you explain why we got rid of reduceInfo? i would think its useful for debugging, but im not sure.

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
-			// 3. Now rename the stale VG to a unique temporary name using the device filter.
+			// 4. Compute unique temporary name for the stale VG.
 			shortUuid := staleVgUuid
 			if len(shortUuid) > 8 {
 				shortUuid = shortUuid[:8]
 			}
 			tempVgName := fmt.Sprintf("csi-vg-stale-%s", shortUuid)
-			klog.Infof("Renaming stale VG %s to %s using device filter", volumeGroupName, tempVgName)
+
+			// 5. Rename the stale VG using the device filter to resolve the duplicate name conflict.
```
hime: should we instead combine step 4 and 5 so that we get a unique temporary name the first time around?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+			klog.Infof("Step 5: Regenerating UUID for stale VG %s to resolve duplicate VGID conflict", tempVgName)
+			vgchangeArgs := []string{"-u", tempVgName, "--config", configFilter}
+			_, uuidErr := runCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "vgchange", vgchangeArgs...)
+			if uuidErr != nil {
+				klog.Warningf("Failed to regenerate UUID for stale VG %s: %v. Continuing recovery.", tempVgName, uuidErr)
 			}
-		} else {
-			klog.V(4).Infof("No duplicate VG name conflict detected for %s, skipping stale VG recovery", volumeGroupName)
 		}
+	} else {
```
hime: there's many if statements, so I forgot which if statement belongs to this else. if we could refresh with a small comment that would be great!

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -67,49 +83,84 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+		klog.V(4).Infof("No duplicate VG name conflict detected for %s, skipping stale VG recovery", volumeGroupName)
```
hime: Can we have multiple volumegroups in the node? Can we comment a bit of how that works here and explain that its fine to not clean up that non-duplicate volumegroup?

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -125,84 +176,61 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
...
+		return "", fmt.Errorf("failed to get VG name for PV %s: %w", devicePath, err)
 	}
 	klog.V(4).Infof("Physical volume is part of Volume group: %v", vgNameForPv)
 	if vgNameForPv == volumeGroupName {
 		klog.V(4).Infof("Physical Volume(PV) already exists in the Volume Group %v", volumeGroupName)
-	} else if vgNameForPv != "VG" && vgNameForPv != "" {
-
-		info, err = common.RunCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "vgchange", []string{"-an", vgNameForPv}...)
+	} else if vgNameForPv != "" {
+		klog.Infof("Stale VG %s found on GCE PD. Deactivating and uncaching first to preserve data.", vgNameForPv)
```
hime: We also take care of stale VG when deduplicating right? Wondering if we can recycle some logic here

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -273,7 +310,7 @@ func setupCaching(devicePath string, req *csi.NodeStageVolumeRequest, nodeId str
 			"--cachevol",
 			cacheLvName,
 			"--zero",
-			"y",
+			"n",
```
hime: can we comment on what this is supposed to do

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -485,20 +522,23 @@ func getVolumeGroupName(nodePath string) string {
 	return fmt.Sprintf("csi-vg-%s", nodeHash)
 }
 
-func getVgUuidForPv(pvPath string) (string, error) {
+func getVgInfoForPv(pvPath string) (string, string, error) {
```
hime: can you add a comment explaining the 3 variables the fucntion returns

// getVgInfoForPv gets info by ...
// returns vg name, vg uid, err

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -485,20 +522,23 @@ func getVolumeGroupName(nodePath string) string {
...
 	resolvedPvPath, err := filepath.EvalSymlinks(pvPath)
 	if err != nil {
 		resolvedPvPath = pvPath // Fallback
 	}
 
+	configFilter := fmt.Sprintf("devices { filter = [ \"a|%s|\", \"r|.*|\" ] }", resolvedPvPath)
```
hime: can you comment what the regex is doing

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -507,40 +547,48 @@ func getVgUuidForPv(pvPath string) (string, error) {
 			continue
 		}
 		fields := strings.Fields(line)
-		if len(fields) >= 2 {
-			resolvedFieldName, err := filepath.EvalSymlinks(fields[1])
+		if len(fields) >= 3 {
```
hime: why 3? can you explain what the output looks like for readability

pkg/gce-pd-csi-driver/cache.go
```diff
@@ -507,40 +547,48 @@ func getVgUuidForPv(pvPath string) (string, error) {
...
+		"-o", "pool_lv",
+		"--config", configFilter,
 	}
-	return count > 1, nil
+	output, err := runCommand("" /* pipedCmd */, nil /* pipedCmdArg */, "lvs", args...)
+	if err != nil {
+		// If lvs exits with a warning/error (which it always does when a PV is missing),
+		// we still want to parse the output. If the VG/LV genuinely doesn't exist,
+		// the output will be empty and we will correctly return false.
+		klog.V(4).Infof("lvs returned warning/error while checking cache on %s: %v. Parsing output anyway.", pvPath, err)
```
hime: wondering how parsing output helps here as im not very familar with lvs

pkg/gce-pd-csi-driver/cache_test.go
hime: Thanks!

test/e2e/utils/utils.go
```diff
@@ -287,6 +286,10 @@ func CopyFile(instance *remote.InstanceInfo, src, dest string) error {
 }
 
 func InstallDependencies(instance *remote.InstanceInfo, pkgs []string) error {
+	if instance.IsCloudtopHost() {
```
hime: Is this referring to internal cloudtop?

sandeep-gh-acc: this was local change, have removed it, the latest commit does not have these changes

test/e2e/utils/utils.go
```diff
@@ -73,7 +73,6 @@ func GCEClientAndDriverSetup(instance *remote.InstanceInfo, driverConfig DriverC
 		"--device-in-use-timeout=10s", // Set lower than the usual value to expedite tests
 		fmt.Sprintf("--fallback-requisite-zones=%s", strings.Join(driverConfig.Zones, ",")),
 		"--disk-topology=true",
-		"--dynamic-volumes=true",
```
hime, line 76: im not sure how this deletion is related to the rest of the code.

sandeep-gh-acc:
This PR has been broken to multiple smaller PRs

PR #2353, opened 2026-06-24
e2e: add dynamic volume Pod Workload and Snapshot/Restore tests

sudhirshanmugam:
**What type of PR is this?**

/kind feature

Adds new E2E test cases to `test/e2e/tests/dynamic_volume_e2e_test.go` covering Pod Workload and Snapshot/Restore behavior for the `type=dynamic` disk selection feature, plus two edge cases for `selectDiskTypeFromTopologies()`.

The seven test cases added are:

1. **Pod Migration** — Verifies that data written on node A is still readable after the volume is unpublished/unstaged from node A and reattached/mounted on node B, and that the disk type resolved at creation (`pd-balanced`) remains unchanged after migration.
2. **Pod Restart** — Verifies that a dynamic volume (resolved to `hyperdisk-balanced`) is correctly detached and reattached across a simulated pod restart, with data written before the restart still intact afterward.
3. **Snapshot of Dynamic HD Volume** — Verifies that a snapshot taken from a `hyperdisk-balanced` dynamic volume reaches `READY` state and correctly references the source disk.
4. **Restore from Snapshot to Correct Disk Type** — Verifies that restoring a snapshot into a new `type=dynamic` volume resolves to a concrete disk type (`hyperdisk-balanced`), not `dynamic`.
5. **Restore Snapshot to Larger Size** — Verifies that restoring a snapshot into a larger dynamic volume (2x source size) preserves the correct disk type and applies the expanded size.
6. **No Matching Node Labels** — Verifies that when the `Preferred` topology carries only a zone segment (no `disk-type.gke.io/*` labels), `selectDiskTypeFromTopologies()` falls back to the built-in default disk type (`hyperdisk-balanced`).
7. **Mixed Node Pool Disk Selection** — Verifies that in a cluster with both HD-capable and PD-only nodes, the driver resolves `hyperdisk-balanced` on the HD node and `pd-balanced` on the PD node based on each pod's node topology.

```
None
```

```
NONE
```

/release-note-none

### DIFF
--- a/test/e2e/tests/dynamic_volume_e2e_test.go
+++ b/test/e2e/tests/dynamic_volume_e2e_test.go
@@ -15,15 +15,19 @@ limitations under the License.
 package tests
 
 import (
+	"fmt"
 	"path/filepath"
+	"time"
 
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/common"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/constants"
 	gce "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/gce-cloud-provider/compute"
 	"sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/pkg/parameters"
 	testutils "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/test/e2e/utils"
+	remote "sigs.k8s.io/gcp-compute-persistent-disk-csi-driver/test/remote"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


test/e2e/tests/dynamic_volume_e2e_test.go
```diff
@@ -316,3 +322,724 @@ var _ = Describe("GCE PD CSI Driver Dynamic Volumes Lifecycle", func() {
...
+// Verifies that data written on node A persists after the pod is migrated to node B,
+// and that the disk type remains unchanged throughout.
+var _ = Describe("GCE PD CSI Driver Dynamic Volumes Pod Migration", func() {
+
+	It("Should persist data and retain disk type after pod migration from node A to node B", func() {
+		Expect(testContexts).To(HaveLen(2),
+			"Need at least 2 PD nodes for pod migration test")
+
+		tcA := testContexts[0]
+		tcB := testContexts[1]
```
mattcary: Will these two contexts be in the same zone? If not, then reattaching will fail as disks can't move across zones (basically).

It looks like we currently create these contexts in zonal pairs, which is why your test might be passing, but  that's fragile, we should explicitly ensure the contexts are in the same zone, eg AI suggested

```
var tcA, tcB *remote.TestContext
for i := 0; i < len(testContexts); i++ {
    for j := i + 1; j < len(testContexts); j++ {
        _, zoneA, _ := testContexts[i].Instance.GetIdentity()
        _, zoneB, _ := testContexts[j].Instance.GetIdentity()
        if zoneA == zoneB {
            tcA = testContexts[i]
            tcB = testContexts[j]
            break
        }
    }
}
Expect(tcA).NotTo(BeNil(), "Need at least 2 nodes in the same zone for zonal pod migration")
``

sudhirshanmugam: made the changes

mattcary:
Please investigate the e2e failure. 

The arm and windows failures are expected.

mattcary:
Please investigate the e2e test failure. The arm & windows failures are expected.

su-sudhir:
> Please investigate the e2e test failure. The arm & windows failures are expected.

Removed the Restore Snapshot to Larger Size test case from this PR.
The e2e suite was hitting panic: test timed out after 50m0s since all specs run serially in one process, and this test's runtime was pushing the total past Go's 50-minute timeout. With it removed, the suite passes.

mattcary:
> > Please investigate the e2e test failure. The arm & windows failures are expected.
> 
> Removed the Restore Snapshot to Larger Size test case from this PR. The e2e suite was hitting panic: test timed out after 50m0s since all specs run serially in one process, and this test's runtime was pushing the total past Go's 50-minute timeout. With it removed, the suite passes.

Is it still skipped or did you do that manually? You should add a skip in the code so that this doesn't cause the CI to time out --- it's not clear to me if you did that or not.

sudhirshanmugam:
> > > Please investigate the e2e test failure. The arm & windows failures are expected.
> > 
> > 
> > Removed the Restore Snapshot to Larger Size test case from this PR. The e2e suite was hitting panic: test timed out after 50m0s since all specs run serially in one process, and this test's runtime was pushing the total past Go's 50-minute timeout. With it removed, the suite passes.
> 
> Is it still skipped or did you do that manually? You should add a skip in the code so that this doesn't cause the CI to time out --- it's not clear to me if you did that or not.

The skip hasn't been tested, Let me add the skip flag and do a push.

PR #2360, opened 2026-06-30
Handle access mode correctly for Hyperdisk Balanced and Throughput during retried CreateVolume()

msau42:
**What type of PR is this?**
/kind bug

Retain the existing behavior for using "" as the default RWO access mode for HdT and HdE disk types to match with PD.

The fix modifies the ValidateExistingDisk() function to handle the "" case for Hyperdisk types and enforces that the disk's access mode is RWO.

Fixes #

```release-note
Fix issue where creating a Hyperdisk Throughput or Extreme disk from a large snapshot fails with access mode mismatch 
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce-compute.go
+++ b/pkg/gce-cloud-provider/compute/gce-compute.go
@@ -476,6 +476,9 @@ func validAccessMode(want, got string) bool {
 		return true
 	}
 	switch want {
+	case "":
+		// Empty string defaults to RWO
+		return got == constants.GCEReadWriteOnceAccessMode
 	case constants.GCEReadOnlyManyAccessMode, constants.GCEReadWriteOnceAccessMode:
 		return got == constants.GCEReadWriteManyAccessMode
 	// For RWX, no other access mode is valid.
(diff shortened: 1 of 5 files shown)


pkg/gce-cloud-provider/compute/gce-compute.go
```diff
@@ -476,6 +476,9 @@ func validAccessMode(want, got string) bool {
 		return true
 	}
 	switch want {
+	case "":
+		// Empty string defaults to RWO
+		return got == constants.GCEReadWriteOnceAccessMode
```
mattcary, line 481: Maybe for future-proofing, return got == RWO || got == ""?

mattcary: ie it seems like we should be as flexible as possible

msau42: matching string is already handled by L475 above. I will add a test case for it.

msau42:
Adding a test here to repro the issue and verify this fix: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/2363

PR #2374, opened 2026-07-30
Add support for exapool-hyperdisk-balanced

hime:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/pkg/parameters/constants.go
+++ b/pkg/parameters/constants.go
@@ -59,6 +59,9 @@ const (
 	DiskTypeHdE  = "hyperdisk-extreme"
 	DiskTypeHdML = "hyperdisk-ml"
 
+	// Exapool hyperdisk types
+	ExapoolHyperdiskBalanced = "exapool-hyperdisk-balanced"
+
 	// Parameters for VolumeSnapshotClass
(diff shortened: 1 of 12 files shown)


pkg/parameters/constants.go
```diff
@@ -59,6 +59,9 @@ const (
 	DiskTypeHdE  = "hyperdisk-extreme"
 	DiskTypeHdML = "hyperdisk-ml"
 
+	// Exapool hyperdisk types
+	ExapoolHyperdiskBalanced = "exapool-hyperdisk-balanced"
```
sunnylovestiramisu: Why does this need to be a separate disk type? Can it stay hyperdisk-balanced?

Because it means in the storage class the user needs to specify type: exapool-hyperdisk-balanced ?

hime: Yes, seems like the easiest way to do it and scale it. For example, we also have "hyperdisk-balanced-high-availability" which is also hyperdisk balanced.

PR #2379, opened 2026-08-13
Fix: improve default SC rollback handler resiliency and support standard container mode

savirg:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

The `default-sc-rollback-handler` was originally introduced to handle storage class defaults, but running it early in the pod lifecycle exposed a race condition: the controller could boot before the Addon Manager finished applying the necessary RBAC permissions for `StorageClass` access. This led to a `Forbidden` (403) error. 

Furthermore, if we use an `initContainer` and simply wait/retry for RBAC, it blocks the main PD CSI driver from starting and responding to RPCs. 

To resolve this, this PR adapts the rollback handler to run safely as a standard sidecar container:
1. **Adds RBAC Retries:** Introduces a retry loop when fetching the storage class, allowing the container to gracefully wait for RBAC application instead of failing fatally.
2. **Idles on Success:** Adds a blocking OS signal channel at the end of the handler's execution. This ensures that when it runs as a standard container, it enters a zero-CPU idle state upon success rather than exiting and triggering an endless `CrashLoopBackOff`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/scdefaultrestorer/restorer.go
+++ b/pkg/scdefaultrestorer/restorer.go
@@ -42,6 +42,7 @@ const (
 type Config struct {
 	StorageClassName string
 	APITimeout       time.Duration
+	AccessTimeout    time.Duration
 }
 
 type Restorer struct {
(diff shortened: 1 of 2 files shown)


pkg/scdefaultrestorer/restorer.go
```diff
@@ -90,16 +91,54 @@ func (r *Restorer) waitForAPIServer(ctx context.Context, client kubernetes.Inter
...
+		return false, fmt.Errorf("failed while waiting for StorageClass %s access: %w", targetSCName, err)
+	}
+
+	return exists, nil
+}
+
+func (r *Restorer) restoreDefaultAnnotation(ctx context.Context, client kubernetes.Interface) error {
 	targetSCName := r.config.StorageClassName
 
+	// Wait for permissions and check existence
```
mattcary, line 131: Can you add a comment to the effect that the restorer is only used to update existing storage classes. So there might be race conditions with PD CSI component rbac during cluster update, but there's no race condition with waiting for a storage class to be created, because the restorer only cares about pre-existing storage classes.

Otherwise the behavior in waitForStorageClassAccess is confusing, as usually you would be concerned about race conditions with resource creation and so not exit immediately on NotFound errors.

savirg: Done. I added an explanatory comment directly to the waitForStorageClassAccess method.

PR #2380, opened 2026-08-13
Actually update device path when manually set

mattcary:
/kind failing-test

I noticed the e2e tests would flake around waiting for a device symlink. Turns out we weren't setting the path in the return value. This would have a minor effect on real workloads, as the node stage would get retried.

```release-note
Make device symlink detection more reliable.
```

### DIFF
--- a/pkg/deviceutils/device-utils.go
+++ b/pkg/deviceutils/device-utils.go
@@ -239,7 +239,12 @@ func (m *deviceUtils) VerifyDevicePath(devicePaths []string, deviceName string) 
 			// Couldn't find a /dev/disk/by-id path for this deviceName, so we need to
 			// find a /dev/* with a serial that matches deviceName. Then we attempt
 			// to repair the symlink.
-			klog.Warningf("For disk %s couldn't find a device path, calling udevadmTriggerForDiskIfExists", deviceName)
+			dirListing, err := os.ReadDir("/dev/disk/by-id")
+			if err != nil {
+				klog.Warningf("(will ignore) When checking for device paths, /dev/disk/by-id unreadable: %v", err)
+			} else {
+				klog.Warningf("For disk %s couldn't find a device path, calling udevadmTriggerForDiskIfExists. /dev/disk/by-id: %v", deviceName, dirListing)
+			}
 			innerErr := udevadmTriggerForDiskIfExists(deviceName)
 			if innerErr != nil {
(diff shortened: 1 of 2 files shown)


pkg/gce-pd-csi-driver/utils_linux.go
```diff
@@ -39,12 +40,10 @@ func getDevicePath(ns *GCENodeServer, volumeID, partition string) (string, error
 	}
 	devicePaths := ns.DeviceUtils.GetDiskByIdPaths(deviceName, partition)
 	devicePath, err := ns.DeviceUtils.VerifyDevicePath(devicePaths, deviceName)
-	if err != nil {
+	if devicePath == "" || err != nil {
+		klog.Errorf("VerifyDevicePath failed for %q: %v", deviceName, err)
 		return "", status.Error(codes.Internal, fmt.Sprintf("error verifying GCE PD (%q) is attached: %v", deviceName, err.Error()))
```
sunnylovestiramisu, line 48: For devicePath == "" and err == nil this error logging will panic because of err.Error()?

pkg/deviceutils/device-utils.go
```diff
@@ -239,7 +239,11 @@ func (m *deviceUtils) VerifyDevicePath(devicePaths []string, deviceName string)
 			// Couldn't find a /dev/disk/by-id path for this deviceName, so we need to
 			// find a /dev/* with a serial that matches deviceName. Then we attempt
 			// to repair the symlink.
-			klog.Warningf("For disk %s couldn't find a device path, calling udevadmTriggerForDiskIfExists", deviceName)
+			dirListing, err := os.ReadDir("/dev/disk/by-id")
+			if err != nil {
+				klog.Warningf("(will ignore) When checking for device paths, /dev/disk/by-id unreadable: %v", err)
+			}
+			klog.Warningf("For disk %s couldn't find a device path, calling udevadmTriggerForDiskIfExists. /dev/disk/by-id: %v", deviceName, dirListing)
```
sunnylovestiramisu: Calling out - if err !=nil dirListing will be logged as nil here, there will not be a crash.

pkg/gce-pd-csi-driver/utils_linux.go
mattcary: Thanks!  good catch

This was a double-check because we shouldn't return a nonempty device path with a non-nil error.

pkg/deviceutils/device-utils.go
mattcary: Oops, should probably have an else.

PR #2343, opened 2026-06-11
Bump gcp-compute-persistent-disk-csi-driver image to v1.26.0 in stable-master overlay

darkweaver87:
# Motivation

The `stable-master` controller template ships the `taint-controller` container running `--run-taint-controller`, but this overlay's image transformer pins `gcp-compute-persistent-disk-csi-driver` to `v1.17.4`, which lacks that flag — so `kustomize build deploy/kubernetes/overlays/stable-master` produces a controller that crashloops with `flag provided but not defined: -run-taint-controller`. Every other image in this `image.yaml` is current; this bumps the driver to `v1.26.0` (latest release, promoted to registry.k8s.io, supports the flag).

```release-note
NONE
```

PR #2388, opened 2026-08-26
Bump gcp-compute-persistent-disk-csi-driver image to v1.26.5

DTLP:
**What type of PR is this?**
/kind bug

The `stable-master` image.yaml on the `release-1.26` branch pins `gcp-compute-persistent-disk-csi-driver` to `v1.17.4`, well behind the `v1.26.x` line — `v1.26.5` is the latest release on this branch. This bumps the driver image tag to `v1.26.5` so the overlay matches the branch it ships on.

Mirrors #2343, which made the same fix on `master`; this applies it to `release-1.26`.

```release-note
NONE
```


DTLP:
The CI failures look unrelated — #2343 (which this mirrors) hit the same errors and got merged anyway. Everything else is green.

PR #2389, opened 2026-08-27
Add ResponseHeaderTimeout to GCE oauth http client transport

Rishita-Golla:
**What type of PR is this?**
/kind bug

Configure custom http.Transport with ResponseHeaderTimeout: 30s in newOauthClient to ensure that stalled GCE API socket reads fail fast. This allows the driver's CreateVolume and other CSI operations to return errors promptly upon endpoint hangs, triggering defer volumeLocks.Release and preventing in-memory volume lock deadlocks on subsequent retries.

Added testing logs and verification plan in b/552618197

NONE

```release-note
Add 60s ResponseHeaderTimeout to GCE oauth http client transport
```

### DIFF
--- a/pkg/gce-cloud-provider/compute/gce.go
+++ b/pkg/gce-cloud-provider/compute/gce.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


sunnylovestiramisu:
Can you look into the e2e test failure and see it is caused by the change?

Rishita-Golla:
I've updated the ResponseHeaderTimeout to 60s to handle the failing tests and prevent e2e timeouts. The other test failures don't seem related to this PR.

sunnylovestiramisu:
Calling out - if GCE API legitimately needs >60s to send response headers (very rare), those requests will still fail—but this is a tradeoff for reliability.

Issue #40, opened 2018-07-02
Node DaemonSet in Kubernetes Deployment Should Not Run On Master


davidz627:
Currently the Daemonset constantly attempts to create the pod on master and fails because of resource limits. We should stop scheduling on master so that it does not continually loop attempted creation.

This is high priority as we have seen the daemonset pod kick off the api server from the master putting the whole cluster in a bad state.

msau42:
Looking at my master node,
```
Taints:             <none>
Unschedulable:      true
```

My master daemonset pod:
```
Tolerations:     node.kubernetes.io/disk-pressure:NoSchedule
                 node.kubernetes.io/memory-pressure:NoSchedule
                 node.kubernetes.io/not-ready:NoExecute
                 node.kubernetes.io/unreachable:NoExecute
```

@janetkuo, do you know how to prevent daemonset pods from being scheduled on the master?

msau42:
(this is gce, not gke)

davidz627:
From https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/:
```
The unschedulable field of a node is not respected by the DaemonSet controller.
```

Looks like taints are respected though, but my thought is tainting the master node could have unwanted side-effects

PR #2392, opened 2026-08-31
Add readiness probe to node driver registrar

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

In order to integrate with the [Node Readiness Controller (NRC)](https://github.com/kubernetes-sigs/node-readiness-controller), which taints a node until the PD CSI driver is ready, this PR adds a `readinessProbe` and `--http-endpoint=:9931` to `csi-node-driver-registrar` in Linux and Windows node DaemonSets.

Without this probe, there is a race condition where the node pod is marked `Ready` before the driver finishes registration with Kubelet. Consequently, NRC may remove the node taint prematurely, causing scheduled workloads to fail volume mounting. Adding this `readinessProbe` ensures the pod is only marked `Ready` after driver registration completes, preventing the race condition.

Additionally, I noticed that the node DaemonSet was previously scheduled onto master/control-plane nodes due to the wildcard toleration (`operator: Exists`), which was tracked by the longstanding `#TODO(#40): Force DaemonSet to not run on master` in [issue#40](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/40). Without a readiness probe, the pod on the master node was considered `Ready` on startup despite failing registration shortly after. However, with the new `readinessProbe`, the master pod fails Kubelet registration, enters `CrashLoopBackOff`, and remains `NotReady`—which indefinitely blocks CI driver startup checks (e.g., `wait-for-driver.sh` in upstream GCE integration tests).

To resolve this, this PR adds `affinity.nodeAffinity` to both Linux and Windows node DaemonSets to ensure pods are not scheduled on nodes labeled with `node-role.kubernetes.io/control-plane`. Additionally, it updates `test/k8s-integration/cluster.go` to explicitly label master nodes with `node-role.kubernetes.io/control-plane` post-creation in GCE E2E clusters, where [kube-up.sh](https://github.com/kubernetes/kubernetes/blob/b2ec8b6fefac451a2dedafc4dd71f2f16c7a6abe/cluster/gce/config-default.sh#L196) provisions master nodes without standard role labels.

```release-note
Add readiness probe to node driver registrar
```

### DIFF
--- a/deploy/kubernetes/base/node_linux/node.yaml
+++ b/deploy/kubernetes/base/node_linux/node.yaml
@@ -1,4 +1,3 @@
-#TODO(#40): Force DaemonSet to not run on master.
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
@@ -19,6 +18,17 @@ spec:
       hostNetwork: true
       priorityClassName: csi-gce-pd-node
       serviceAccountName: csi-gce-pd-node-sa
+      # Set node affinity to prevent scheduling on control plane (master) nodes.
+      # Because this DaemonSet specifies a wildcard toleration (operator: Exists),
+      # it would otherwise be scheduled on control plane nodes.
+      # See: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/40
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
+                operator: DoesNotExist
       nodeSelector:
         kubernetes.io/os: linux
       containers:
--- a/deploy/kubernetes/base/node_linux/node.yaml
+++ b/deploy/kubernetes/base/node_linux/node.yaml
@@ -28,6 +28,18 @@ spec:
             - "--v=5"
             - "--csi-address=/csi/csi.sock"
             - "--kubelet-registration-path=/var/lib/kubelet/plugins/pd.csi.storage.gke.io/csi.sock"
+            - "--http-endpoint=:9931"
+          ports:
+            - containerPort: 9931
+              name: healthz
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 1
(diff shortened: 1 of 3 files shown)


deploy/kubernetes/base/node_linux/node.yaml
```diff
@@ -19,6 +19,13 @@ spec:
       hostNetwork: true
       priorityClassName: csi-gce-pd-node
       serviceAccountName: csi-gce-pd-node-sa
+      affinity:
```
hime, line 25: Can you add a comment why we are doing this? Perhaps also linking https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/issues/40 would help. Thanks!

hime: Oh and removing the TODO 40 from the files too.

yaozile123: Done, thanks for catching that!

test/k8s-integration/cluster.go
```diff
@@ -149,6 +149,26 @@ func clusterUpGCE(k8sDir, gceZone string, numNodes int, numWindowsNodes int, ima
...
+		return fmt.Errorf("failed to get nodes: %s, err: %w", out, err)
+	}
+	for _, node := range strings.Fields(string(out)) {
+		if strings.HasSuffix(node, "-master") {
+			labelCmd := exec.Command("kubectl", "label", "node", node, "node-role.kubernetes.io/control-plane=", "--overwrite")
+			out, err := labelCmd.CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("failed to label master node %s: %s, err: %w", node, out, err)
+			}
+			klog.Infof("Labeled master node %s with node-role.kubernetes.io/control-plane", node)
```
sunnylovestiramisu, line 170: Nit, do we want to log when no master node is found with "-master"?

yaozile123: Good point. Added a warning log if no master node with suffix `-master` is found.

PR #2393, opened 2026-09-03
Add Persistent Disk to Hyperdisk conversion via VolumeAttributesClass

sudhirshanmugam:
**What type of PR is this?**

/kind feature

Adds Persistent Disk to Hyperdisk conversion driven entirely through the
Kubernetes API, so users can migrate volumes with `kubectl` alone rather than
needing access to GCP REST APIs.

Applying a `VolumeAttributesClass` whose `type` differs from the volume's actual
disk type triggers a conversion via the GCE `disks.convert` API. A conversion is
a long-running operation — minutes to hours depending on disk size — that runs
outside any CSI call, so this also adds the state and tracking needed to follow
one to completion and survive a driver restart:

- **Operation tracking.** A background watcher polls the GCE operation with capped
  backoff. The operation selfLink is recorded on the PV, so state survives a
  controller restart or leader election, with a fallback that records completion
  at the next attach, snapshot or expand if no watcher is alive.
- **Annotation lifecycle.** `Pending` → operation selfLink → `converted-from` /
  `converted-to`, with reconciliation for volumes queued for a conversion that is
  no longer wanted or already the target type. An operation-ownership check stops
  a superseded watcher from clearing a newer conversion.
- **Failure classification.** Terminal and retriable failures are distinguished by
  the GCE error *reason* rather than HTTP status, since 400 and 403 cover both.
  Applied to failures returned when the conversion is requested and to failures
  discovered mid-flight by the poller. Retriable failures return the volume to
  `Pending` and pass the GCE error through to the PVC status; terminal failures
  return `InvalidArgument` and leave attach unblocked.
- **Operation guards.** Attach, expand, snapshot and delete are refused while a
  conversion is in progress, and fail closed if the conversion state cannot be
  read.
- **Request pacing.** Conversion API calls are rate limited to stay within the
  per-region limit on `disks.convert` calls.

A VAC whose `type` matches the disk exactly is treated as an IOPS/throughput
update and does not trigger a conversion.

Fixes #

```release-note
action required: The GCE PD CSI driver can now convert Persistent Disks to
Hyperdisk in place. Applying a VolumeAttributesClass whose `type` differs from a
volume's current disk type converts it, preserving the disk's name and identity.
The disk must be detached: conversion starts immediately if it is, otherwise at
the next detach. Attach, expand, snapshot and delete are blocked while a
conversion runs. Enable with `--enable-pd-conversion`.

Two deployment requirements: csi-resizer must match the VolumeAttributesClass API
version the cluster serves (v1.13.2 for `storage.k8s.io/v1beta1`, v2.0.0 for
`storage.k8s.io/v1`) and must run with `--feature-gates=VolumeAttributesClass=true`;
and nodes must be Hyperdisk-capable (C3, C4 or N4) to attach converted volumes.
```


mattcary:
We are working on getting the CLA signed.

hime:
Closing as it was replaced with #2397

PR #2397, opened 2026-09-08
Pd to hyperdisk conversion

sudhirshanmugam:
**What type of PR is this?**

/kind feature

Adds Persistent Disk to Hyperdisk conversion driven entirely through the
Kubernetes API, so users can migrate volumes with `kubectl` alone rather than
needing access to GCP REST APIs.

Applying a `VolumeAttributesClass` whose `type` differs from the volume's actual
disk type triggers a conversion via the GCE `disks.convert` API. A conversion is
a long-running operation — minutes to hours depending on disk size — that runs
outside any CSI call, so this also adds the state and tracking needed to follow
one to completion and survive a driver restart:

- **Operation tracking.** A background watcher polls the GCE operation with capped
  backoff. The operation selfLink is recorded on the PV, so state survives a
  controller restart or leader election, with a fallback that records completion
  at the next attach, snapshot or expand if no watcher is alive.
- **Annotation lifecycle.** `Pending` → operation selfLink → `converted-from` /
  `converted-to`, with reconciliation for volumes queued for a conversion that is
  no longer wanted or already the target type. An operation-ownership check stops
  a superseded watcher from clearing a newer conversion.
- **Failure classification.** Terminal and retriable failures are distinguished by
  the GCE error *reason* rather than HTTP status, since 400 and 403 cover both.
  Applied to failures returned when the conversion is requested and to failures
  discovered mid-flight by the poller. Retriable failures return the volume to
  `Pending` and pass the GCE error through to the PVC status; terminal failures
  return `InvalidArgument` and leave attach unblocked.
- **Operation guards.** Attach, expand, snapshot and delete are refused while a
  conversion is in progress, and fail closed if the conversion state cannot be
  read.
- **Request pacing.** Conversion API calls are rate limited to stay within the
  per-region limit on `disks.convert` calls.

A VAC whose `type` matches the disk exactly is treated as an IOPS/throughput
update and does not trigger a conversion.

Fixes #

```release-note
action required: The GCE PD CSI driver can now convert Persistent Disks to
Hyperdisk in place. Applying a VolumeAttributesClass whose `type` differs from a
volume's current disk type converts it, preserving the disk's name and identity.
The disk must be detached: conversion starts immediately if it is, otherwise at
the next detach. Attach, expand, snapshot and delete are blocked while a
conversion runs. Enable with `--enable-pd-conversion`.

Two deployment requirements: csi-resizer must match the VolumeAttributesClass API
version the cluster serves (v1.13.2 for `storage.k8s.io/v1beta1`, v2.0.0 for
`storage.k8s.io/v1`) and must run with `--feature-gates=VolumeAttributesClass=true`;
and nodes must be Hyperdisk-capable (C3, C4 or N4) to attach converted volumes.
```

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -88,6 +88,28 @@ const (
 	HyperdiskExtremeIopsPerGB          = 2
 	HyperdiskThroughputThroughputPerGB = 10
 	BytesInGB                          = 1024
+
+	// Maximum provisioned IOPS/throughput ceilings, per
+	// https://cloud.google.com/compute/docs/disks/hyperdisk-perf-limits
+	//
+	// hyperdisk-balanced: max IOPS is MIN(500 * sizeGb, 160000); max
+	// throughput depends on the provisioned IOPS value (P) rather than size:
+	// MIN(2400, P/4) MiB/s.
+	HyperdiskBalancedMaxIops               = 160000
+	HyperdiskBalancedMaxThroughput         = 2400
+	HyperdiskBalancedThroughputIopsDivisor = 4
+	// hyperdisk-extreme: max IOPS is MIN(1200 * sizeGb, 350000). Throughput is
+	// not user-provisionable for this type.
+	HyperdiskExtremeMaxIopsPerGB = 1200
+	HyperdiskExtremeMaxIops      = 350000
+	// hyperdisk-throughput: max throughput is MIN(90 * sizeTiB, 2400) MiB/s.
+	// IOPS is not user-provisionable for this type.
+	HyperdiskThroughputMaxThroughputPerTiB = 90
+	HyperdiskThroughputMaxThroughput       = 2400
+	// hyperdisk-ml: max throughput is MIN(1600 * sizeGb, 2097152) MiB/s. IOPS
+	// is not user-provisionable for this type.
+	HyperdiskMLMaxThroughputPerGB = 1600
+	HyperdiskMLMaxThroughput      = 2097152
 )
 
 var (
@@ -587,4 +609,98 @@ func minThroughputForThroughput(disk *computev1.Disk, requestGb int64) (needed b
 		return true, 0, minRequiredThroughput
 	}
 	return false, 0, 0
+}
+
+// ValidateMaxProvisioned checks the requested provisioned IOPS and/or
+// throughput against GCE's published ceiling for diskType and sizeGb, so an
+// out-of-range request is rejected client-side instead of round-tripping to
+// the API.
+// https://cloud.google.com/compute/docs/disks/hyperdisk-perf-limits
+//
+// currentIops is the disk's already-provisioned IOPS. It is used as the basis
+// for hyperdisk-balanced's throughput ceiling, which depends on the
+// provisioned IOPS value rather than disk size, when the request changes
+// throughput without also changing IOPS.
+func ValidateMaxProvisioned(diskType string, sizeGb, currentIops int64, iops, throughput *int64) error {
+	switch {
+	case strings.Contains(diskType, parameters.DiskTypeHyperdiskBalanced):
+		return maxIopsThroughputForBalanced(sizeGb, currentIops, iops, throughput)
+	case strings.Contains(diskType, parameters.DiskTypeHdE):
+		return maxIopsForExtreme(sizeGb, iops)
+	case strings.Contains(diskType, parameters.DiskTypeHdT):
+		return maxThroughputForThroughputType(sizeGb, throughput)
+	case strings.Contains(diskType, parameters.DiskTypeHdML):
+		return maxThroughputForMLType(sizeGb, throughput)
+	}
+	return nil
+}
+
+func maxIopsThroughputForBalanced(sizeGb, currentIops int64, iops, throughput *int64) error {
+	maxIops := sizeGb * HyperdiskBalancedIopsPerGB
+	if maxIops > HyperdiskBalancedMaxIops {
+		maxIops = HyperdiskBalancedMaxIops
+	}
+	if iops != nil && *iops > maxIops {
+		return fmt.Errorf("provisioned IOPS %d exceeds the maximum of %d for a %d GiB hyperdisk-balanced disk", *iops, maxIops, sizeGb)
+	}
+	if throughput == nil {
+		return nil
+	}
+	// The throughput ceiling depends on the provisioned IOPS value (P) rather
+	// than size. Use the requested IOPS when this request also changes it,
+	// otherwise fall back to the disk's already-provisioned IOPS.
+	p := currentIops
+	if iops != nil {
+		p = *iops
+	}
+	maxThroughput := p / HyperdiskBalancedThroughputIopsDivisor
+	if maxThroughput > HyperdiskBalancedMaxThroughput {
+		maxThroughput = HyperdiskBalancedMaxThroughput
+	}
+	if *throughput > maxThroughput {
+		return fmt.Errorf("provisioned throughput %d MiB/s exceeds the maximum of %d MiB/s for provisioned IOPS %d on hyperdisk-balanced", *throughput, maxThroughput, p)
+	}
+	return nil
+}
+
+func maxIopsForExtreme(sizeGb int64, iops *int64) error {
+	if iops == nil {
+		return nil
+	}
+	maxIops := sizeGb * HyperdiskExtremeMaxIopsPerGB
+	if maxIops > HyperdiskExtremeMaxIops {
+		maxIops = HyperdiskExtremeMaxIops
+	}
+	if *iops > maxIops {
+		return fmt.Errorf("provisioned IOPS %d exceeds the maximum of %d for a %d GiB hyperdisk-extreme disk", *iops, maxIops, sizeGb)
+	}
+	return nil
+}
+
+func maxThroughputForThroughputType(sizeGb int64, throughput *int64) error {
+	if throughput == nil {
+		return nil
+	}
+	maxThroughput := sizeGb * HyperdiskThroughputMaxThroughputPerTiB / BytesInGB
+	if maxThroughput > HyperdiskThroughputMaxThroughput {
+		maxThroughput = HyperdiskThroughputMaxThroughput
+	}
+	if *throughput > maxThroughput {
+		return fmt.Errorf("provisioned throughput %d MiB/s exceeds the maximum of %d MiB/s for a %d GiB hyperdisk-throughput disk", *throughput, maxThroughput, sizeGb)
+	}
+	return nil
+}
+
+func maxThroughputForMLType(sizeGb int64, throughput *int64) error {
+	if throughput == nil {
+		return nil
+	}
+	maxThroughput := sizeGb * HyperdiskMLMaxThroughputPerGB
+	if maxThroughput > HyperdiskMLMaxThroughput {
+		maxThroughput = HyperdiskMLMaxThroughput
+	}
+	if *throughput > maxThroughput {
+		return fmt.Errorf("provisioned throughput %d MiB/s exceeds the maximum of %d MiB/s for a %d GiB hyperdisk-ml disk", *throughput, maxThroughput, sizeGb)
+	}
+	return nil
 }
--- a/pkg/gce-pd-csi-driver/controller.go
+++ b/pkg/gce-pd-csi-driver/controller.go
@@ -18,8 +18,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	neturl "net/url"
+	"path"
 	"sort"
 	"strconv"
 	"strings"
@@ -33,8 +35,10 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/strings/slices"
@@ -56,6 +60,20 @@ type GCEControllerServer struct {
 	volumeEntries     []*csi.ListVolumesResponse_Entry
 	volumeEntriesSeen map[string]int
 	listVolumesLock   sync.Mutex
+
+	// The background watchers following disk type conversions, keyed by volume
+	// key. A watcher outlives the call that started it, so it has to be
+	// stoppable when the volume goes away.
+	conversionWatchers     map[string]*conversionWatcher
+	conversionWatchersLock sync.Mutex
+
+	// Tracks the background conversions started on detach, so that they can be
+	// waited for rather than assumed finished.
+	conversionWorkers sync.WaitGroup
+
+	// How often a running conversion is checked. Zero means the default,
+	// conversionPollBackoff. Tests set it so they need not wait minutes.
+	conversionPollBackoffOverride wait.Backoff
 
 	snapshots         []*csi.ListSnapshotsResponse_Entry
 	snapshotTokens    map[string]int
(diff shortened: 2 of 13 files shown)


pkg/gce-pd-csi-driver/controller.go
```diff
@@ -913,8 +931,48 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+	}
+
+	// If the VolumeAttributesClass requests a disk type that doesn't match the
+	// disk's actual type, this is a disk type conversion request rather than an
+	// IOPS/throughput update. This must be handled before the checks below,
+	// which are keyed on the disk's current type and would reject types that
+	// are valid conversion sources.
 	diskType := existingDisk.GetPDType()
+	if volumeModifyParams.DiskType != nil && *volumeModifyParams.DiskType != diskType {
+		return gceCS.convertDiskType(ctx, project, volKey, volumeID, existingDisk, volumeModifyParams)
```
hime, line 956: The current implementation relies on an in-memory watcher and continuously sends duplicate conversion requests during routine retries, causing driver restarts to lose track of in-flight operations and needlessly burning regional API quota.

To solve both issues, the driver should treat the PersistentVolume annotation as the source of truth for ongoing conversions. Before initiating a conversion or attaching a disk, the driver should explicitly check this annotation and poll the existing operation status rather than defaulting to new backend requests.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
 
+// convertDiskType starts an in-place disk type conversion for a volume whose
+// VolumeAttributesClass requests a type different from the disk's current type.
+//
+// The GCE conversion can run for minutes to hours, far longer than a single
+// ModifyVolume call may block, so this returns as soon as the operation has
+// been accepted. The modification stays pending and is retried; once the disk
+// reports the target type, the retry falls through to the regular
+// IOPS/throughput path and the modification completes.
+func (gceCS *GCEControllerServer) convertDiskType(ctx context.Context, project string, volKey *meta.Key, volumeID string, existingDisk *gce.CloudDisk, params parameters.ModifyVolumeParameters) (*csi.ControllerModifyVolumeResponse, error) {
```
hime, line 1013: We should acquire gceCS.volumeLocks.TryAcquire in ControllerModifyVolume and runQueuedConversion just like ControllerPublishVolume, ControllerUnpublishVolume, CreateSnapshot, and DeleteVolume do.

Without this per-volume lock, a concurrent attach, snapshot, or resize can pass checkNoConversionInProgress and mutate the disk in the window while ConvertDiskType is in flight before pdcsi.gke.io/disk-type-conversion-operation is written to the PV.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -930,6 +988,10 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
 		return nil, err
 	}
 
+	if err := common.ValidateMaxProvisioned(diskType, existingDisk.GetSizeGb(), existingDisk.GetProvisionedIops(), volumeModifyParams.IOPS, volumeModifyParams.Throughput); err != nil {
```
hime: When converting a Persistent Disk to hyperdisk-balanced with throughput set and iops omitted, ValidateMaxProvisioned falls back to existingDisk.GetProvisionedIops, which is 0 on a Persistent Disk, causing maxIopsThroughputForBalanced to calculate a max throughput ceiling of 0 MiB/s and falsely reject the conversion. Instead, when iops is omitted and currentIops is 0, maxIopsThroughputForBalanced should calculate the throughput ceiling using hyperdisk-balanced's default provisioned IOPS for that disk size or skip the IOPS-derived check so GCE can validate against the target disk's default IOPS.


Image

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -913,8 +931,48 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+	if volumeModifyParams.DiskType != nil {
+		if err := gceCS.completeConversionIfInProgress(ctx, volKey, volumeID, diskType); err != nil {
+			return nil, err
+		}
+	}
+
+	// The disk is already the requested type. If no other attributes were
+	// requested there is nothing left to do, which is also how a completed
+	// conversion is reported as successful on a subsequent retry.
+	if volumeModifyParams.IOPS == nil && volumeModifyParams.Throughput == nil {
```
hime: we should check if the volume already has these parameters set

pkg/common/utils.go
```diff
@@ -88,6 +88,28 @@ const (
...
 	HyperdiskThroughputThroughputPerGB = 10
 	BytesInGB                          = 1024
+
+	// Maximum provisioned IOPS/throughput ceilings, per
+	// https://cloud.google.com/compute/docs/disks/hyperdisk-perf-limits
+	//
+	// hyperdisk-balanced: max IOPS is MIN(500 * sizeGb, 160000); max
+	// throughput depends on the provisioned IOPS value (P) rather than size:
+	// MIN(2400, P/4) MiB/s.
+	HyperdiskBalancedMaxIops               = 160000
```
hime, line 98: We are currently doing the IOPS and Throughput validation checks in the PDCSI driver. We should instead have the GCE API validate whether the IOPS/Throughput given by the customer is valid. This is because these values can change on GCE side either through an update or allowlist of customer higher limits. In those cases, the driver would be out of date.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -913,8 +931,48 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
 		return nil, err
 	}
 
-	// Check if the disk supports dynamic IOPS/Throughput provisioning
+	// A class that asks for nothing is how a volume's class is withdrawn, so a
+	// conversion still queued for it is dropped rather than left waiting for a
+	// class that no longer asks for anything.
+	if volumeModifyParams.DiskType == nil && volumeModifyParams.IOPS == nil && volumeModifyParams.Throughput == nil {
```
hime, line 937: Only removing the DiskType from the VAC should be enough to immediately abort a queued conversion.

Because this if block requires all three parameters to be nil, it introduces latency in the time it takes to abort the operation. if a user removes the type parameter but simultaneously provides an iops update, this block is bypassed entirely.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -913,8 +931,48 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+
+	// If the VolumeAttributesClass requests a disk type that doesn't match the
+	// disk's actual type, this is a disk type conversion request rather than an
+	// IOPS/throughput update. This must be handled before the checks below,
+	// which are keyed on the disk's current type and would reject types that
+	// are valid conversion sources.
 	diskType := existingDisk.GetPDType()
+	if volumeModifyParams.DiskType != nil && *volumeModifyParams.DiskType != diskType {
+		return gceCS.convertDiskType(ctx, project, volKey, volumeID, existingDisk, volumeModifyParams)
+	}
```
hime, line 957: An improvement based on the edge case of a restart: To restore prompt completion tracking after a driver restart, the driver should safely spawn a new background watcher whenever it checks the PV annotation and discovers an active conversion that isn't currently being tracked.

pkg/gce-pd-csi-driver/controller.go
hime, line 970: ConvertDiskType already sets IOPS and Throughput during conversion, but this check only tests for nil and falls through to call UpdateDisk a second time when those parameters are present in the VAC. We should solve this, perhaps by also returning early when volumeModifyParams.IOPS and volumeModifyParams.Throughput already match existingDisk.GetProvisionedIops and existingDisk.GetProvisionedThroughput.

GCE enforces a 4-hour cooldown on Hyperdisk Disks.Update calls, so this redundant update can fail with rateLimitExceeded and leave the PVC stuck in InProgress. Skipping it also avoids burning the disk's modification window and API quota on a no-op write.

hime: Because convertDiskType returns codes.Unavailable while the conversion runs, external-resizer retries ControllerModifyVolume with the same VAC parameters (type, iops, throughput) until it gets OK. Once the conversion finishes in GCE, existingDisk.GetPDType matches volumeModifyParams.DiskType, so the next retry skips line 956, completes the conversion at line 962, and lands here on line 970 with volumeModifyParams.IOPS still set.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+// The GCE conversion can run for minutes to hours, far longer than a single
+// ModifyVolume call may block, so this returns as soon as the operation has
+// been accepted. The modification stays pending and is retried; once the disk
+// reports the target type, the retry falls through to the regular
+// IOPS/throughput path and the modification completes.
+func (gceCS *GCEControllerServer) convertDiskType(ctx context.Context, project string, volKey *meta.Key, volumeID string, existingDisk *gce.CloudDisk, params parameters.ModifyVolumeParameters) (*csi.ControllerModifyVolumeResponse, error) {
+	targetDiskType := *params.DiskType
+	currentDiskType := existingDisk.GetPDType()
+
+	if !gceCS.EnablePdConversion {
```
hime, line 1017: The EnablePdConversion is already used for a different feature.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+
+	operation, err := gceCS.startDiskTypeConversion(ctx, project, volKey, currentDiskType, targetDiskType, params)
+	if err != nil {
+		// A conversion already running means the disk is being rebuilt, so the
+		// volume stays blocked whatever the failure is classified as.
+		if isConversionInProgressError(err) {
+			gceCS.markConversionPending(ctx, volKey, volumeID)
+			return nil, status.Errorf(codes.Unavailable, "conversion of volume %s from %s to %s is in progress", volumeID, currentDiskType, targetDiskType)
+		}
+		if isUnsupportedConversionIntentError(err) {
```
hime, line 1066: When a queued conversion fails with a permanent error, withdrawInfeasibleClass records the failure and tries to clear the VolumeAttributesClass, but it never removes the Pending conversion annotation from the PV. If Kubernetes refuses to clear the class from the PVC (such as on clusters before 1.34 or when a previous class was already applied), that leftover Pending annotation will permanently block the volume from ever attaching again, so withdrawInfeasibleClass should also remove the conversion operation annotation.

pkg/gce-pd-csi-driver/controller.go
hime: If the feature is disabled, we should make sure the driver does not block volume reattachment to a node or similar.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+		return nil, status.Errorf(codes.InvalidArgument, "cannot convert volume %s from %s to %s: %s", volumeID, currentDiskType, targetDiskType, reason)
+	}
+
+	// Conversion requires the disk to be detached. Report this as retryable so
+	// the conversion starts once the workload using the volume is scaled down.
+	if users := existingDisk.GetUsers(); len(users) > 0 {
+		// The annotation is what makes the detach hook start this conversion, so
+		// losing it means the conversion never resumes. The error returned below
+		// is retryable, and the next attempt writes the annotation again.
+		gceCS.markConversionPending(ctx, volKey, volumeID)
```
hime, line 1054: In markConversionPending, if reading the existing PV annotation fails (err != nil), the function still proceeds to line 1146 and writes "Pending" to the PV.

This is dangerous because it will overwrite and permanently erase an active GCE conversion operation URL if a conversion is already running (such as when called from isConversionInProgressError on line 1063). The operation URL is the only record the driver has to track the running conversion.

Instead of blindly overwriting the annotation when the read fails, markConversionPending should return an error so the caller can fail the current CSI operation and let Kubernetes safely retry without clobbering the in-flight operation ID.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+// lets the conversion be tracked and completed later, by whichever call sees the
+// disk next. It returns the conversion operation.
+//
+// The caller decides what a failure means for the operation it is serving, and
+// is responsible for the checks that depend on that context, such as whether the
+// disk is still attached.
+func (gceCS *GCEControllerServer) startDiskTypeConversion(ctx context.Context, project string, volKey *meta.Key, currentDiskType, targetDiskType string, params parameters.ModifyVolumeParameters) (string, error) {
+	klog.V(4).Infof("Converting disk %s from %s to %s", volKey.Name, currentDiskType, targetDiskType)
+
+	operation, err := gceCS.CloudProvider.ConvertDiskType(ctx, project, volKey, targetDiskType, params.IOPS, params.Throughput)
```
hime, line 1161: If the disk is already detached when the conversion starts, **the PV does not have a Pending annotation yet**, so if saving the PV annotations fails after ConvertDiskType is called, the conversion will run in GCE with no annotations on the PV to block reattachment or track completion.

We should save the Pending and converted-from annotations before calling ConvertDiskType (and return an error if that fails), and then update the annotation with the operation URL once ConvertDiskType succeeds.

hime: we previously set PV annotation to pending when `if len(users) > 0`, but perhaps we should just set it if its not there, period.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+				// state is cleared and the volume released. Any other failure,
+				// including one that carries no recognised reason, is retriable:
+				// the volume goes back to waiting for a conversion, which keeps
+				// it blocked and starts a new conversion on the next attempt.
+				klog.Errorf("Disk type conversion %s of volume %s failed: %v", opName, volKey.Name, err)
+				if isUnsupportedConversionIntentError(err) {
+					gceCS.recordFailedConversion(ctx, volKey, operation, err)
+					return
+				}
+				gceCS.requeueFailedConversion(ctx, volKey, err)
```
hime, line 1341: requeueFailedConversion should verify that operation still matches the PV's operation annotation before resetting the PV to Pending, so a stale watcher cannot overwrite a newer conversion (similar to recordFailedConversion and recordCompletedConversion).

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+			continue
+		}
+		if !done {
+			continue
+		}
+
+		disk, err := gceCS.CloudProvider.GetDisk(ctx, project, volKey)
+		if err != nil {
+			klog.Warningf("Disk type conversion %s of volume %s finished but the disk could not be read, leaving it to be recorded at the next attach: %v", opName, volKey.Name, err)
+			return
```
hime, line 1356: If GetDisk or recordCompletedConversion fails, we should continue so the watcher retries on the next poll tick (similar to line 1347). Exiting early stops the watcher from clearing the annotation, and falling through runs reconcileLatestVACAfterConversion before the completed conversion is even recorded.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+		}
+
+		// The class may have been changed while this conversion was running, so
+		// what it asks for now decides whether another conversion follows.
+		//
+		// Recording the completion stops this watcher, which cancels the context
+		// the poll was using, so the work that follows gets one that outlives it.
+		reconcileCtx := context.Background()
+		if err := gceCS.reconcileLatestVACAfterConversion(reconcileCtx, project, volKey, disk); err != nil {
+			klog.Warningf("Failed to act on the VolumeAttributesClass of volume %s after its conversion: %v", volKey.Name, err)
```
hime, line 1370: Because `recordCompletedConversion` removes the conversion operation annotation on line 1359, the volume is unblocked before reconcileLatestVACAfterConversion runs on line 1369. That leaves a window where a waiting pod can attach the disk right as reconcileLatestVACAfterConversion tries to start a follow-up conversion, so we should check whether another conversion is needed before clearing the annotation.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+	if !exists {
+		return
+	}
+
+	// Check if this is the current operation
+	// We need to check if the error we're recording matches the current operation
+	// For simplicity, if there's an operation annotation, we'll clear it on failure.
+	// This is safe because a failed conversion should always unblock the volume.
+
+	if err := k8sclient.RemovePVAnnotation(ctx, volKey.Name, constants.DiskTypeConversionOperationKey); err != nil {
```
hime, line 1434: When a conversion fails terminally in GCE, clearing the PV annotations without recording disk-type-conversion-infeasible causes the next ControllerModifyVolume retry to start the same doomed conversion all over again. 

We should set the disk-type-conversion-infeasible annotation here (like withdrawInfeasibleClass does) and check it in convertDiskType so the next retry returns InvalidArgument instead of calling ConvertDiskType again.

pkg/gce-pd-csi-driver/controller.go
```diff
@@ -940,6 +1002,947 @@ func (gceCS *GCEControllerServer) ControllerModifyVolume(ctx context.Context, re
...
+	// A. Read the current VAC using the existing queuedConversionOutcome
+	params, cancelReason, err := gceCS.queuedConversionOutcome(ctx, volKey, disk)
+	if err != nil {
+		return status.Errorf(codes.Unavailable,
+			"failed to read VolumeAttributesClass for disk %s after conversion: %v",
+			volKey.Name, err)
+	}
+
+	// B. Handle VAC removal or cancellation
+	if cancelReason != "" {
```
hime, line 1752: Because queuedConversionOutcome on line 1744 already checks params.DiskType == nil, disk.GetPDType() == *extracted.DiskType, and unconvertibleToTargetReason (returning all of them as cancelReason), returning nil on line 1755 makes lines 1761–1796 completely unreachable.

If the user updated the VAC to an infeasible target during the conversion, line 1755 silently returns nil instead of calling withdrawInfeasibleClass to record the disk-type-conversion-infeasible annotation and emit the failure event.

pkg/k8sclient/client.go
```diff
@@ -141,3 +162,306 @@ func listPodsInNamespace(ctx context.Context, kubeClient kubernetes.Interface, n
...
+// A volume with no PersistentVolume, no claim, or no class reports no class
+// rather than failing, since none of those are errors: they all mean there is
+// nothing the user is currently asking for.
+func GetVolumeAttributesClassForPV(ctx context.Context, pvName string) (map[string]string, bool, error) {
+	kubeClient, err := GetClient()
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to get kubernetes client: %w", err)
+	}
+
+	pv, err := GetPersistentVolumeWithRetry(ctx, pvName)
```
hime, line 184: GetVolumeAttributesClassForPV retries fetching the PV on line 184, but fetches the PVC (line 195) and the VolumeAttributesClass (lines 219 and 227) without retries. We should wrap those Get calls in ExponentialBackoffWithContext as well (and pass kubeClient to getPersistentVolumeWithRetry on line 184 so we don't call GetClient() twice).

PR #2398, opened 2026-09-09
Use a manually created device symlink in VerifyDevicePath.

mattcary:
/kind bug

Previously, if symlink was manually created, it would not be passed back to VerifyDevicePath, and be ignored.

```release-note
Use manually created device path during device verification.
```


pkg/deviceutils/device-utils.go
hime, line 215: // As a last resort, if udevadm fails, it will fall back to manually scanning 
// for the device by its hardware serial and manually creating the symlink.

PR #2401, opened 2026-09-14
Add node readiness reporter for PDCSIReady condition

yaozile123:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

This PR lets the PD CSI node driver participate in GKE's Node Readiness Controller (NRC) flow. On GKE, the control plane taints new nodes with `readiness.k8s.io/gke-pd-csi-ready:NoSchedule`, and the NRC removes that taint once the `node.gke.io/PDCSIReady` condition is `True` on the Node — keeping volume-dependent workloads off the node until PD CSI is usable there. The driver is the missing node-side piece to reports the condition, so the NRC has no signal to act on and the taint never clears.

This PR closes that gap behind a new opt-in flag `--enable-node-readiness-condition`. When set, the node service:
1. Polls the co-located registrar's `/healthz` (`http://127.0.0.1:9931`, from #2392) every 500ms until it returns HTTP 200 — i.e. kubelet registration has completed.
2. Sets `node.gke.io/PDCSIReady=True` on its own `nodes/status` via a StrategicMergePatch, then exits. Skipped if the condition is already `True`.
3. Retries transient API errors every 5s until the context is cancelled.

Additionally:
- Updates Linux and Windows node `ClusterRole`s to allow `patch` on `nodes/status` (and `get` on `nodes` for Windows).
- Passes `--node-name=$(KUBE_NODE_NAME)` in the Windows daemonset manifest to match Linux.

Fixes #

- We use `127.0.0.1` rather than  `localhost`to avoid failed IPv6 `[::1]` attempt.

```release-note
Added an opt-in `--enable-node-readiness-condition` flag to the node driver. When enabled, the driver waits for `csi-driver-registrar` to become healthy and sets `node.gke.io/PDCSIReady=True` on the Node status to signal readiness to node bootstrap taint controllers.

```


deploy/kubernetes/base/controller/cluster_setup.yaml
```diff
@@ -220,6 +223,12 @@ rules:
     verbs:     ['use']
     resourceNames:
     - csi-gce-pd-node-psp-win
+  - apiGroups: [""]
+    resources: ["nodes"]
```
hime: interesting that windows didnt have nodes.get

pkg/gce-pd-csi-driver/node_readiness.go
```diff
@@ -0,0 +1,187 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+)
+
+const (
+	PDCSIReadyConditionType = v1.NodeConditionType("node.gke.io/PDCSIReady")
+	PDCSIReadyReason        = "CSIDriverReady"
+	PDCSIReadyMessage       = "GCE PD CSI driver node service is registered and ready"
+	registrarHealthzURL     = "http://127.0.0.1:9931/healthz"
```
hime: Please make the port configurable via flag. Also, please see this AI suggestion and let me know your thoughts:

In addition to port collisions, hardcoding the IPv4 loopback (127.0.0.1) will break this on IPv6 single-stack clusters (where the loopback is [::1]). Defaulting to localhost and allowing the entire endpoint to be configurable via a flag (e.g., --registrar-healthz-endpoint=localhost:9931) would resolve both issues.

deploy/kubernetes/base/controller/cluster_setup.yaml
```diff
@@ -220,6 +223,12 @@ rules:
...
+    verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["nodes/status"]
```
hime, line 227: do we also have to update `deploy/kubernetes/base/node/node.yaml`?

pkg/gce-pd-csi-driver/node_readiness.go
```diff
@@ -0,0 +1,187 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+)
+
+const (
+	PDCSIReadyConditionType = v1.NodeConditionType("node.gke.io/PDCSIReady")
```
hime, line 37: regarding using the `node.gke` prefix, this driver is also used on non-gke environments, can you check if this is fine to do based on other codebase examples? thanks!

pkg/gce-pd-csi-driver/node_readiness.go
```diff
@@ -0,0 +1,187 @@
...
+		kubeClient:    kubeClient,
+		httpClient:    &http.Client{Timeout: registrarHTTPTimeout},
+		pollInterval:  registrarPollInterval,
+		patchInterval: patchRetryInterval,
+	}, nil
+}
+
+// Run polls the registrar healthz endpoint until HTTP 200 is returned, then issues a one-shot
+// StrategicMergePatch setting node.gke.io/PDCSIReady=True on the Node status, unless it is already set.
+func (r *NodeReadinessReporter) Run(ctx context.Context) {
```
hime, line 85: Because this is a one-shot function that exits after setting the condition to True, it has no way to revert the condition if the driver terminates or the node restarts, leaving a stale state. Can we do a few things here to mitigate this?

1. Handle graceful shutdowns: Modify this to catch context cancellation and patch the condition back to False.
2. Clear stale state on startup: Proactively label the node as False at the very beginning of Run() before waiting for the registrar. This ensures that if the pod previously crashed, the newly restarted container instantly clears the stale condition.
2. Document the staleness period: Add a comment in the code acknowledging that hard crashes (e.g., OOMKilled) will still leave a temporary stale condition until the external NRD takes over.

hime (changes requested):

deploy/kubernetes/base/controller/cluster_setup.yaml
yaozile123: We do have nodes.get for windows internally, but not in our oss repo

deploy/kubernetes/base/controller/cluster_setup.yaml
yaozile123: Did you mean to add `--node-name=$(KUBE_NODE_NAME)` in /node_linux/node.yaml? We actually have that https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/40bb262cc8f28bf22999b8f8247454ca93288df3/deploy/kubernetes/base/node_linux/node.yaml#L72

pkg/gce-pd-csi-driver/node_readiness.go
yaozile123: Good question, I used this prefix since NRC already used this prefix for node conditions on the control-plane side. For non-GKE deployments, they could run their own node-readiness-controller and define a `NodeReadinessRule` whose `conditions[].type` equals the condition the driver reports. NRC keeps the taint on the node until that condition is True:
```
apiVersion: readiness.node.x-k8s.io/v1alpha1
kind: NodeReadinessRule
metadata:
  name: pd-csi-ready
spec:
  enforcementMode: bootstrap-only
  conditions:
  - type: node.gke.io/PDCSIReady
    requiredStatus: "True"
  taint:
    key: readiness.k8s.io/pd-csi-ready
    value: pending
    effect: NoSchedule
```

pkg/gce-pd-csi-driver/node_readiness.go
yaozile123: Done, I have made the port configurable via flag and default to `127.0.0.1:9931`. I asked AI to confirm that Linux always assigns 127.0.0.1/8 on lo even in IPv6 single-stack clusters, and the registrar listens on :9931 (dual-stack) in the same netns. So there's shouldn't be a problem by using 127.0.0.1.

pkg/gce-pd-csi-driver/node_readiness.go
yaozile123: Thanks for the comment, I have addressed (2) and (3). We will patches `PDCSIReady=False` for single attempt on startup before polling registrar `/healthz`, and also added the comments about stale condition.  also deleted `conditionAlreadyTrue` (and the nodes get RBAC rule for Windows) since `Run()` now resets the condition to False on startup, so it will never already be True when `waitForRegistrarHealthz `finishes.

For (1) I'd prefer to keep that out of scope for this PR since [main.go](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/3ff53121e3758b8de329faf6da4f0913d7335a5f/cmd/gce-pd-csi-driver/main.go#L237-L238) does not currently watch SIGTERM and [gceDriver.Run()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/3ff53121e3758b8de329faf6da4f0913d7335a5f/cmd/gce-pd-csi-driver/main.go#L393) blocks indefinitely on [s.Wait()](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/3ff53121e3758b8de329faf6da4f0913d7335a5f/pkg/gce-pd-csi-driver/server.go#L56-L67) which mean it will process exits immediately on termination without ever cancelling ctx.

deploy/kubernetes/base/controller/cluster_setup.yaml
yaozile123: We actually don't need nodes.get any more since the reporter now will patch node condition as false during start up, so checking node status would be redundant.

